PR #2, opened 2018-12-11
Set up basic Circle CI configuration

janpaul123:
With linting and testing for now. Tightened the linter a bit to not even
allow warnings, and fixed the one warning that we had.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,37 @@
+# See https://circleci.com/docs/2.0/language-javascript/
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/node:11-browsers
+
+    steps:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,37 @@
...
+      - checkout
+
+      - restore_cache:
+          keys:
+          - global-deps-{{ checksum "package-lock.json" }}
+          - global-deps-
+
+      - restore_cache:
+          keys:
+          - regl-worldview-deps-{{ checksum "package-lock.json" }}
```
jtbandes: Looks like this doesn't match the `checksum "packages/regl-worldview/package-lock.json"` used for this key in the `save_cache` below

PR #3, opened 2018-12-12
Move `// @flow` to before license header

janpaul123:
Babel seems to not strip out types properly otherwise!


janpaul123:
Yeah, it for some reason didn't strip out generics otherwise. 🤷‍♂️

PR #7, opened 2018-12-12
Overhaul linting/testing/etc

janpaul123:
Our code has to have the same linting and testing configuration as our
internal/proprietary codebase so we can easily move things around
between them. This commit brings them roughly in sync, though there is
probably a bunch of cleanup that I missed. There will be more PRs
following this, but I figured it would be good to merge this soon to
prevent conflicts.

Test plan: CI passes; made sure Storybook still works.


docs/src/routes.js
```diff
@@ -50,35 +48,37 @@ export const componentList = {
...
   },
 ];
 
 // convert route names to component names, e.g. 'Quick Start' => 'QuickStart'
 function getComponentName(str) {
   return str.replace(/(?:^\w|[A-Z]|\b\w|\s+)/g, (match, index) => {
-    if (+match === 0) return ''; // or if (/\s+/.test(match)) for white spaces
+    if (+match === 0) {
+      return "";
+    } // or if (/\s+/.test(match)) for white spaces
```
vidaaudrey: Issue with auto formatting, need to move the comment to line 79

jest/cssTransform.js
```diff
@@ -0,0 +1,12 @@
+// This is a custom Jest transformer turning style imports into empty objects.
+// http://facebook.github.io/jest/docs/tutorial-webpack.html
```
vidaaudrey: this link is no longer valid

jest/fileTransform.js
```diff
@@ -0,0 +1,10 @@
+const path = require("path");
+
+// This is a custom Jest transformer turning file imports into filenames.
+// http://facebook.github.io/jest/docs/tutorial-webpack.html
```
vidaaudrey: same as above

jest/jest.config.js
```diff
@@ -0,0 +1,23 @@
...
+
+//  Copyright (c) 2018-present, GM Cruise LLC
+//
+//  This source code is licensed under the Apache License, Version 2.0,
+//  found in the LICENSE file in the root directory of this source tree.
+//  You may not use this file except in compliance with the License.
+
+module.exports = {
+  rootDir: "..",
+  collectCoverageFrom: ["client/**/*.{js,jsx}"],
```
vidaaudrey: we no longer have the `client/` dir

package.json
```diff
@@ -56,34 +77,32 @@
...
     "check-packages": "yarn clean && yarn compile && yarn test && yarn lint",
     "publish-packages": "./scripts/publish",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
     "flow": "flow",
     "flow-mono": "flow-mono create-symlinks ./dist/.flowconfig",
-    "flow-mono-install": "flow-mono install-types",
-    "precommit": "lint-staged"
-  },
-  "lint-staged": {
```
vidaaudrey: If we don't use lint-staged any more, we also need to delete flow-typed file

PR #8, opened 2018-12-12
Export commands and build types

vidaaudrey:
- Export regl commands for reusability 
- Build a separate type file so we could use `import type { Vec3, Arrow } from 'regl-worldview/dist/types'`

### DIFF
--- a/packages/regl-worldview/src/commands/index.js
+++ b/packages/regl-worldview/src/commands/index.js
@@ -10,14 +10,14 @@ export { default as Command, SimpleCommand, makeCommand } from "./Command";
 
 // Primitives
 export { default as Arrows } from "./Arrows";
-export { default as Cones } from "./Cones";
-export { default as Cubes } from "./Cubes";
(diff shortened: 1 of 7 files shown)


packages/regl-worldview/src/commands/index.js
```diff
@@ -10,14 +10,14 @@ export { default as Command, SimpleCommand, makeCommand } from "./Command";
 
 // Primitives
 export { default as Arrows } from "./Arrows";
-export { default as Cones } from "./Cones";
-export { default as Cubes } from "./Cubes";
-export { default as Cylinders } from "./Cylinders";
+export { default as Cones, cones } from "./Cones";
+export { default as Cubes, cubes } from "./Cubes";
```
jtbandes: Where/how are we hoping to reuse these exactly? What about just letting someone use `Cubes.command` instead of separately exporting `cubes`?

vidaaudrey: Makes sense. Going to revert the changes. Thanks!

PR #12, opened 2018-12-13
Add stories for instanced cubes

brianc:
- Move cube commands into their own file
- Move common util functions into their own file.  I think this was the main motivation for not breaking up other stories and should make it easier in the future.
- Add stories for instanced cubes with and without colors.

Test plan: added stories.


packages/regl-worldview/src/stories/Cubes.stories.js
```diff
@@ -0,0 +1,99 @@
...
+import { storiesOf } from "@storybook/react";
+import React from "react";
+
+import Container from "./Container";
+import { cube, p, UNIT_QUATERNION, buildMatrix, rng } from "./util";
+import withRange from "./withRange";
+
+import { Cubes, DEFAULT_CAMERA_STATE } from "..";
+
+class Wrapper extends React.Component<any> {
```
vidaaudrey: Would be nice to use functional components to follow where react is going

janpaul123:
I'll soon also add automatic story inclusion for all docs examples so that sounds like a good idea! :)

PR #18, opened 2018-12-13
Load live editor examples in QuickStart

vidaaudrey:
^


brianc:
FYI: There are a couple of lint errors in circle

PR #20, opened 2018-12-13
Remove some unused packages

janpaul123:
And add some that were missing. And also made it so docs open in a
browser just like storybook.

Test plan: covered by CI, manually checked storybook and docs.


package.json
```diff
@@ -5,15 +5,11 @@
     "@babel/cli": "^7.1.5",
     "@babel/core": "^7.2.0",
     "@babel/plugin-proposal-class-properties": "^7.2.1",
-    "@babel/plugin-proposal-export-default-from": "^7.0.0",
     "@babel/plugin-syntax-dynamic-import": "^7.2.0",
     "@babel/plugin-transform-modules-commonjs": "^7.2.0",
-    "@babel/plugin-transform-runtime": "^7.2.0",
```
vidaaudrey: we use `transform-runtime` in rollup.config.js

package.json
```diff
@@ -5,15 +5,11 @@
...
-    "@babel/preset-env": "^7.1.6",
     "@babel/preset-flow": "^7.0.0",
     "@babel/preset-react": "^7.0.0",
     "@mdx-js/loader": "^0.16.6",
-    "@mdx-js/mdx": "^0.16.6",
```
vidaaudrey: we are still using `mdx-js` loader in webpack.config.js right?

package.json
```diff
@@ -23,12 +19,8 @@
     "babel-eslint": "^10.0.1",
     "babel-jest": "^23.4.2",
     "babel-loader": "^8.0.4",
-    "babel-plugin-transform-es2015-modules-commonjs": "^6.26.2",
-    "babel-preset-flow": "^6.23.0",
```
vidaaudrey: we use it in babel.config.js

package.json
janpaul123: Ah shoot I totally missed that. For some reason the build still works without failing. 🤔

package.json
janpaul123: Yep!

package.json
janpaul123: I think that one is using `@babel/preset-flow`.

janpaul123:
> eslint can't resolve the regl-worldview module now, we'll need to add the package path to import resolver kriasoft/react-starter-kit#1180

Hm the webviz resolver should take care of that automatically. I'll dig into that.

package.json
vidaaudrey: oh, you are right. forgot about it

PR #29, opened 2018-12-18
Generate sourcemaps for all builds

vidaaudrey:
Generate sourcemaps for all build files and uglify umd build. 

image

sourcemaps
Ref: https://developers.google.com/web/tools/chrome-devtools/javascript/source-maps

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -5,6 +5,6 @@
   "devDependencies": {
     "lerna": "^3.4.3"
   },
-  "version": "0.0.4",
+  "version": "0.0.4-alpha",
   "npmClient": "npm"
 }
--- a/packages/regl-worldview/rollup.config.js
+++ b/packages/regl-worldview/rollup.config.js
@@ -10,7 +10,7 @@ import copy from "rollup-plugin-copy";
 import resolve from "rollup-plugin-node-resolve";
 import peerDepsExternal from "rollup-plugin-peer-deps-external";
 import replace from "rollup-plugin-replace";
-import { uglify } from "rollup-plugin-uglify";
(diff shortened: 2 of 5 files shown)


packages/regl-worldview/rollup.config.js
```diff
@@ -45,6 +47,7 @@ export default [
       babel(getBabelOptions({ useESModules: true })),
       commonjs({ include: "node_modules/**" }),
       replace({ "process.env.NODE_ENV": JSON.stringify("development") }),
+      uglify(),
```
janpaul123: Might want to consider using `rollup-plugin-terser` instead of uglify if we want `for await` support (see https://github.com/cruise-automation/webviz/blob/530a08ef98ea43d64b85982c26e6d1b81e23bdb5/webpack.config.js#L84). Not sure if that still applies though (and not sure if we even have that in this codebase, but we might have soon if we add more webviz stuff here).

lerna.json
```diff
@@ -5,6 +5,6 @@
   "devDependencies": {
     "lerna": "^3.4.3"
   },
-  "version": "0.0.4",
+  "version": "0.0.4-alpha",
```
janpaul123: Why this version change in this PR?

vidaaudrey: I published a testing version just to make sure it works. Once we merged the changes I'll publish `0.0.5`. I committed the version change here to make it explicit. Shouldn't matter much?

janpaul123: Yeah I don't think it matters much, just think we should generally try to avoid that in PRs to avoid merge conflicts.

vidaaudrey: fair point, will update

PR #32, opened 2018-12-19
Fix deploy

janpaul123:
I broke this in #30 because the files created by the docker images are
owned by the `root` user. This is easily fixed by a `chown` command.

Test plan: removed the check for being on the master branch temporarily
to test this, and seems to work:
https://circleci.com/gh/cruise-automation/webviz/170


janpaul123:
> Is this the canonical way to avoid this issue with Docker or should we be configuring the Docker container with a build arg that sets the user/group or something like that?

Yeah I tried doing that but it seems not possible or at least not easy?

PR #34, opened 2018-12-20
Scroll to top when navigating in the docs

janpaul123:
Classic `HashRouter` issue, with a simple workaround.

Test plan: tested manually.


docs/src/App.js
```diff
@@ -5,16 +5,20 @@
 //  You may not use this file except in compliance with the License.
 
 import { MDXProvider } from "@mdx-js/tag";
+import createBrowserHistory from "history/createBrowserHistory";
```
vidaaudrey: Do we need to add `history` to package.json as well?

PR #35, opened 2018-12-20
Simplify Worldview by updating a few defaults

vidaaudrey:
Update a few props to make Worldview simple to use:

- default `cameraState` and `defaultCameraState` 
- change `hideDebug` to `showDebug`, and don't show debugging view by default 
- modify the camera stories to make the controlled/uncontrolled more explicit 
- update docs, stories, types based on the prop changes 

Test plan: 
Screenshots shouldn't change.


packages/regl-worldview/src/Worldview.js
```diff
@@ -11,49 +11,40 @@ import pickBy from "lodash/pickBy";
...
+export type BaseProps = {|
   backgroundColor?: Vec4,
   hitmapOnMouseMove?: boolean,
-  hideDebug?: boolean,
+  showDebug?: boolean,
   children?: React.Node,
 
+  cameraState: CameraState,
+  defaultCameraState: CameraState,
+  // if onCameraStateChange is present, Worldview is controlled, otherwise it's uncontrolled
```
janpaul123: We should probably use the presence of `cameraState` to determine whether it should be a controller component, to stay consistent with built-in React components. For example, when rendering an `<input value="test"/>` then it is a controlled component, even if we don't specify an `onChange` handler.

docs/src/3.3.Command.mdx
```diff
@@ -62,7 +62,7 @@ class Triangle extends Command {
 // usage
 function CommandDemo() {
   return (
-    <Worldview defaultCameraState={{ ...DEFAULT_CAMERA_STATE, perspective: true }}>
+    <Worldview >
```
jtbandes: nit: stray space

docs/src/Demo.js
```diff
@@ -137,9 +137,8 @@ const HelloWorldview = () => {
         targetOffset: [0, 0, 0],
         targetOrientation: [0, 0, 0, 1],
         thetaOffset: 0,
-        perspective: true,
       }}
-      hideDebug={true}>
+     >
```
jtbandes: nit: our prettier formatting usually puts this immediately after the `}}`

docs/src/jsx/CameraStateControlled.js
```diff
@@ -4,38 +4,41 @@
...
-      cameraState: {
-        ...DEFAULT_CAMERA_STATE,
-        perspective: true,
-      },
-    };
-    this.onCameraStateChange = this.onCameraStateChange.bind(this);
-  }
+function CameraStateControlled() {
+  const [distance, setDistance] = useState(100);
+  const [cameraState, setCamerState] = useState({ ...DEFAULT_CAMERA_STATE, distance });
```
jtbandes: typo: setCameraState

packages/regl-worldview/src/Worldview.js
```diff
@@ -11,49 +11,40 @@ import pickBy from "lodash/pickBy";
...
 
-type SharedProps = {|
+export type BaseProps = {|
   backgroundColor?: Vec4,
   hitmapOnMouseMove?: boolean,
-  hideDebug?: boolean,
+  showDebug?: boolean,
   children?: React.Node,
 
+  defaultCameraState: CameraState,
```
jtbandes: I thought we intentionally kept them separate in the past to prevent people from passing _both_ defaultCameraState and cameraState?

packages/regl-worldview/src/Worldview.js
```diff
@@ -284,6 +269,10 @@ const Worldview = (props: Props) => (
   </ContainerDimensions>
 );
 
+Worldview.defaultProps = {
+  defaultCameraState: DEFAULT_CAMERA_STATE,
```
jtbandes: Why not put this in WorldviewBase?

docs/src/3.3.Command.mdx
vidaaudrey: Good eyes!

packages/regl-worldview/src/Worldview.js
vidaaudrey: with the api change, thought it's no longer necessary to keep it so strict. There are a few scenarios:
- nothing is passed in => uncontrolled, DEFAULT_CAMERA_STATE will apply
- defaultState => uncontrolled 
- state => uncontrolled, state is respected 
- state, defaultState => uncontrolled, defaultState will get ignored 
- onChange => uncontrolled, DEFAULT_CAMERA_STATE will apply
- defaultState, onChange => uncontrolled

- state, onChange => controlled 
- state, defaultState, onChange => controlled, defaultState will get ignored 

The behavior is similar to text input.

packages/regl-worldview/src/Worldview.js
vidaaudrey: Was easy to get around flow warning this way. But I guess this approach is not ideal. Will update in another PR

packages/regl-worldview/src/Worldview.js
janpaul123: Hm I think these are not quite right:

> state => uncontrolled, state is respected
> state, defaultState => uncontrolled, defaultState will get ignored

Should both be controlled; read-only.

jtbandes: > defaultState will get ignored

Isn't this kind of a bug? On the other hand, maybe it makes things easier for some use cases if we allow this. I don't feel strongly.

vidaaudrey: I've updated in https://github.com/cruise-automation/webviz/pull/40/commits/6c453ccf90fd8e77f56d7f21ef93b3ddd73b6d9a. I think it's more accurate to not default the defaultState in static defaultProps, but in constructor, so that we could provide more explicit warnings.

PR #36, opened 2018-12-20
Fix code highlighting when navigating to different pages

vidaaudrey:
The code highlighting for markdown code blocks is lost when navigating to a different page. Fixed by adding rehypePrism plugin (Ref: https://github.com/mdx-js/mdx/issues/253). 

Before:
then

After:
now

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2140,6 +2140,17 @@
         "write-file-atomic": "^2.3.0"
       }
     },
+    "@mapbox/rehype-prism": {
+      "version": "0.3.0",
+      "resolved": "https://registry.npmjs.org/@mapbox/rehype-prism/-/rehype-prism-0.3.0.tgz",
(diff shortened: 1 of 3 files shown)


jtbandes:
Just curious, what does this have to do with navigating between pages? Your gif doesn't seem to show anything changing between different pages; none of them had syntax highlighting.

vidaaudrey:
> Just curious, what does this have to do with navigating between pages? Your gif doesn't seem to show anything changing between different pages; none of them had syntax highlighting.

When we use react router to navigate to another route, the markdown code blocks are not automatically tokenized and stays as template strings, as a result, the css classes can not be applied. If you look carefully, the `before` gif's code blocks doesn't have any color. The babel loader will help generate tokenized code blocks.

janpaul123:
Yeah it kind of accidentally worked on the first page load originally and we're not super sure why. Added this webpack loader is the official way of making syntax highlighting work but for some reason it kind of did that some of the time before. Who knows. 😅

PR #37, opened 2018-12-20
Better positioning of error message

janpaul123:
Overlay it on top of the output so you can actually see it when on a
small screen.

Test plan: manual.

screen shot 2018-12-20 at 3 20 25 pm

### DIFF
--- a/docs/src/jsx/CodeEditor.js
+++ b/docs/src/jsx/CodeEditor.js
@@ -11,6 +11,7 @@ import styled, { css } from "styled-components";
 const StyledProvider = styled(LiveProvider)`
   overflow: hidden;
   margin-bottom: 60px;
+  position: relative;
 `;
 
 const LiveWrapper = styled.div`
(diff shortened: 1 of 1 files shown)


docs/src/jsx/CodeEditor.js
```diff
@@ -59,6 +60,9 @@ const StyledError = styled(LiveError)`
   padding: 1rem;
```
jtbandes: Is prettier not formatting these? Seems the line above has inconsistent indentation. Maybe it only does so for me because I have a VS Code plugin...

janpaul123: Hm the file looks fine to me? And I also don't see tabs mixed with spaces or so?

jtbandes: It's actually line 58 that I was commenting about, it doesn't show up in the PR unless you click the button to expand.

image

janpaul123: Hah weird looks like a github rendering bug? There are only two spaces in front of both lines!

PR #39, opened 2018-12-21
Nicer and faster demo on landing page

janpaul123:
screen shot 2018-12-20 at 5 27 45 pm

### DIFF
--- a/docs/src/Demo.js
+++ b/docs/src/Demo.js
@@ -80,8 +80,8 @@ const Container = styled.div`
 `;
 
 const HelloWorldview = () => {
-  const [numSpheres, setNumSpheres] = useState(1000);
-  const [distance, setDistance] = useState(150);
+  const [spheresPerAxis, setSpheresPerAxis] = useState(6);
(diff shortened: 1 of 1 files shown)


docs/src/Demo.js
```diff
@@ -136,22 +144,19 @@ const HelloWorldview = () => {
...
+        thetaOffset: Math.PI / 4,
+        perspective: true,
+      }}>
       <Spheres>
-        {new Array(${numSpheres}).fill().map((_, i) => ({
+        {{
+          points,
+          colors,
           pose: {
+            position: { x:0, y:0, z:0 },
```
jtbandes: nit: missing spaces.

Wondering if we should allow `pose` to be left blank and default to ((0,0,0), (0,0,0,1)); thoughts?

janpaul123: > Wondering if we should allow pose to be left blank and default to ((0,0,0), (0,0,0,1)); thoughts?

 Yeah that would be nice!

PR #40, opened 2018-12-21
Address review comments for prop changes(#35)

vidaaudrey:
Address comments from: https://github.com/cruise-automation/webviz/pull/35


packages/regl-worldview/src/Worldview.js
```diff
@@ -36,10 +35,15 @@ export type BaseProps = {|
...
+|};
+
+export type BaseProps = {|
   ...Dimensions,
+  ...SharedProps,
+  defaultCameraState: CameraState,
 |};
 
-export type Props = $Diff<BaseProps, Dimensions>;
+export type Props = {| ...SharedProps, defaultCameraState?: CameraState |};
```
jtbandes: So defaultCameraState is optional for Props, but not for BaseProps? Not sure why these are different.

Perhaps it would help to use the `ElementConfig` type: https://flow.org/en/docs/react/types/#toc-react-elementconfig

packages/regl-worldview/src/Worldview.js
```diff
@@ -70,12 +68,32 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
...
+    const { width, height, top, left, backgroundColor, onCameraStateChange, cameraState, defaultCameraState } = props;
+    if (onCameraStateChange) {
+      if (!cameraState) {
+        console.warn(
+          "Worldview camera state is read-only as you provided `onCameraStateChange` prop without `cameraState` prop. Turn worldview to controlled with `cameraState` and `onCameraStateChange` props, or uncontrolled with `defaultCameraState` prop."
+        );
+      }
+      if (cameraState && defaultCameraState) {
+        console.warn(
+          "You provided `cameraState`, `onCameraStateChange` and `defaultCameraState` props. `defaultCameraState` will be ignored for Worldview is controlled when both `cameraState` and `onCameraStateChange` are present."
```
jtbandes:
```suggestion
          "You provided both `cameraState` and `defaultCameraState`. `defaultCameraState` will be ignored."
```

packages/regl-worldview/src/Worldview.js
```diff
@@ -70,12 +68,32 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
 
   constructor(props: BaseProps) {
     super(props);
-    const { width, height, top, left, backgroundColor } = props;
+    const { width, height, top, left, backgroundColor, onCameraStateChange, cameraState, defaultCameraState } = props;
+    if (onCameraStateChange) {
+      if (!cameraState) {
+        console.warn(
+          "Worldview camera state is read-only as you provided `onCameraStateChange` prop without `cameraState` prop. Turn worldview to controlled with `cameraState` and `onCameraStateChange` props, or uncontrolled with `defaultCameraState` prop."
```
jtbandes:
```suggestion
           "You provided `onCameraStateChange` without `cameraState`. Use Worldview as a controlled component with `cameraState` and `onCameraStateChange`, or uncontrolled with `defaultCameraState`."
```

packages/regl-worldview/src/Worldview.js
```diff
@@ -70,12 +68,32 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
...
+      }
+      if (cameraState && defaultCameraState) {
+        console.warn(
+          "You provided `cameraState`, `onCameraStateChange` and `defaultCameraState` props. `defaultCameraState` will be ignored for Worldview is controlled when both `cameraState` and `onCameraStateChange` are present."
+        );
+      }
+    } else {
+      if (cameraState) {
+        console.warn(
+          "Failed prop type: you provided `cameraState` prop without an `onCameraStateChange` handler. This will render a read-only field. If the field should be mutable use `defaultCameraState`. Otherwise set `onCameraStateChange`."
```
jtbandes: I'm not sure this should be a warning, seems like a valid use case to have a fixed camera. But if you want to keep it here are some wording suggestions:
```suggestion
           "You provided `cameraState` without an `onCameraStateChange` handler. This will prevent moving the camera. If the camera should be movable, use `defaultCameraState`, otherwise set `onCameraStateChange`."
```

packages/regl-worldview/src/Worldview.js
```diff
@@ -25,10 +25,10 @@ export type BaseProps = {|
   showDebug?: boolean,
   children?: React.Node,
 
-  defaultCameraState: CameraState,
-  // Worldview is controlled if cameraState and onCameraStateChange are both present
+  // Worldview is only controlled if both cameraState and onCameraStateChange are present
```
jtbandes: I think it should be considered controlled if `cameraState` is present, and uncontrolled if it absent. The presence of `onCameraStateChange` is not relevant other than to indicate whether the developer cares about callbacks when the user tries to move the camera. Theoretically you could have an uncontrolled component with an onChange handler, or a controlled component without one.

I guess for normal React form fields, you usually wouldn't set `value` without `onChange` or `onChange` without `value`, or `onChange` with `defaultValue`, but for Worldview I think those are ok. What do others think?

vidaaudrey: My interpretation of `controlled` was that Worldview will always take in the newest `cameraState` from props. But it's not the case now since we only update internal `cameraState` when both `cameraState` prop and `onCameraStateChange` props are present. I'm considering:

1. Remove the comment here to avoid confusion since the console msgs and docs are very explicit about what are needed for controlled/uncontrolled props
2. Always update internal `cameraState` if the prop exists since it's closer to what native input box would do. Currently it only updates if both `cameraState` and `onCameraStateChange` are present. 

What do you think?

jtbandes: 2nd option sounds good, the current behavior you mentioned sounds confusing.

janpaul123: Agreed with 2nd option. To compare with, say, `<input>` field: passing in `value` but not `onChange` makes the text box read-only. We should probably behave in the same way.

PR #41, opened 2018-12-21
Add stories based on code demos

janpaul123:
And remove redundant stories.


stories/config.js
```diff
@@ -25,8 +25,12 @@ prepareForScreenshots();
 // automatically import all files ending in *.stories.js
 // $FlowFixMe - require.context seems not correctly typed.
 const req = require.context("../packages", true, /.stories.js$/);
+// $FlowFixMe - require.context seems not correctly typed.
+const reqDocs = require.context("../docs", true, /.stories.js$/);
```
jtbandes: both of these should be `\.stories\.js$`

docs/src/jsx/allDemos.stories.js
```diff
@@ -0,0 +1,67 @@
...
+  SpheresInstancedColor,
+  SpheresSingle,
+  Text,
+  Triangles,
+};
+
+const stories = storiesOf("Worldview docs", module);
+stories.addDecorator(withScreenshot());
+
+Object.keys(allDemos).map((demoName) => {
```
jtbandes: Does Storybook automatically sort these alphabetically or should we do it manually (or organize them some other way)?

stories/config.js
janpaul123: Ahh good catch :)

docs/src/jsx/allDemos.stories.js
janpaul123: It sorts automatically!

vidaaudrey: The stories follows the order of `allDemos` object

janpaul123: Oh yeah I think you're right, they might not do automatic sorting after all

PR #42, opened 2018-12-21
Make Camera State example fast

janpaul123:
By rounding some values and then memoizing the function.

Plus animate it to demonstrate following the target. I wonder if we
should show something like this on the landing page, too? This is a
core feature of Worldview that I think is worth highlighting.

camera-state-demo

### DIFF
--- a/docs/src/jsx/CameraState.js
+++ b/docs/src/jsx/CameraState.js
@@ -8,10 +8,12 @@ import { quat, vec3 } from "gl-matrix";
 import React, { useState } from "react";
 
 import CameraStateControls from "./CameraStateControls";
+import useRange from "./useRange";
 import { p, q } from "./utils";
 import Worldview, { Arrows, Spheres, Axes, Grid, cameraStateSelectors } from "regl-worldview";
 
 function CameraState() {
+  const range = useRange();
   const [perspective, setPerspective] = useState(true);
   const [distance, setDistance] = useState(50);
   const [thetaOffset, setThetaOffset] = useState(0.3);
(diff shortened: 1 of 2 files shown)


jtbandes:
What is contributing to the slowness here? Adding a memoize wrapper seems like an easy solution but I wonder if there's a more direct fix we could make which would help improve the perf of all the docs pages. The "distance from target" slider on the main page is slow too, I'm guessing for the same reasons.

Also, I would like to see if we can combine these x,y,z values into a single field (formatted as an array) since it seems a bit unnecessarily verbose to keep them all separate. The fields here should correspond to the fields of the CameraState type.

janpaul123:
> What is contributing to the slowness here? Adding a memoize wrapper seems like an easy solution but I wonder if there's a more direct fix we could make which would help improve the perf of all the docs pages. The "distance from target" slider on the main page is slow too, I'm guessing for the same reasons.

The reason it was slow was mostly just many rerenders. Reducing the number of decimals and then preventing rerenders (`memo` is new React syntax for making it a pure component) solves that pretty easily.

> Also, I would like to see if we can combine these x,y,z values into a single field (formatted as an array) since it seems a bit unnecessarily verbose to keep them all separate. The fields here should correspond to the fields of the CameraState type.

Ah interesting idea. I'd prefer not merging them because now I made the posX value read-only so I can animate it, but it's nice to still be able to change posY and posZ. Plus I like the look of it now where it's arranged neatly in rows and columns. 😛 But happy to take another PR if you find a nicer look!

jtbandes:
See #43  😃

PR #43, opened 2018-12-21
docs: improved CameraStateControls

jtbandes:
A different approach to #42. The slowness was coming from the fact that React [doesn't automatically batch state updates](https://github.com/facebook/react/issues/14259) unless they are inside a lifecycle hook / React event handler.

Anyway, it seems a lot cleaner to put the entire cameraState in one useState hook, so I did that. Also factored out the Slider from the landing page and added labels to it, and made a new Scrubber:


.flowconfig
```diff
@@ -9,6 +9,9 @@ dist
 [lints]
 
 [options]
+# Cheap workaround to make flow typechecking work across packages in the monorepo
```
jtbandes: I think I had to do this because we no longer use flow-mono-cli. @janpaul123 Do you remember why it was removed?

docs/src/utils/Scrubber.js
```diff
@@ -0,0 +1,58 @@
...
+|};
+
+export default function Scrubber({ value, onChange, speed = 1, vertical = false }: Props) {
+  const [dragging, setDragging] = useState(false);
+  const [start, setStart] = useState({
+    value,
+    position: { clientX: NaN, clientY: NaN },
+  });
+
+  useEventListener(window, "mousemove", dragging, (event: MouseEvent) => {
```
vidaaudrey: the `useEventListener` is nice!

.flowconfig
janpaul123: Yeah it didn't seem to do anything — Flow seemed to keep passing just fine without it. Any idea why we needed it now particularly in this PR? Anyway, this seems like a simpler solution than their whole symlink business. 😄

jtbandes: > Any idea why we needed it now particularly in this PR? 

I think this is the first time `// @flow` was added to stuff in the docs folder. 😬 

> Anyway, this seems like a simpler solution than their whole symlink business.

Yeah, and my solution is probably incomplete / wrong in some edge cases too 😅

janpaul123: > Yeah, and my solution is probably incomplete / wrong in some edge cases too 😅

Haha I dunno! It's basically the same as we have in our private webviz repo and that seems to be working just fine!

vidaaudrey: it's working for me too

docs/src/utils/Slider.js
```diff
@@ -0,0 +1,69 @@
...
+
+type Props = {
+  onChange: (number) => void,
+  min?: number,
+  max?: number,
+  minLabel: boolean | string,
+  maxLabel: boolean | string,
+};
+
+export default function Slider({ onChange, minLabel, maxLabel, ...props }: Props) {
```
vidaaudrey: Will be nice to show the slider value on hover.

jtbandes: +1, I agree but I want to let @edenh figure out how it should look :)  (Another option is just to use `Scrubber` everywhere instead?)

vidaaudrey: I like that the slider shows min and max. If you add a value indicator at the bottom of the scrubber (similar to progress bar) we could totally use it everywhere.

jtbandes:
> I do wonder if we should keep the animation that I added there. I think it's quite nice to demonstrate the intent of our camera model (following a moving target but still allowing to drag around to keep an offset to that target). It might be a bit unclear why we have targetOffset otherwise?

I like the animation but I played around with useRange locally and the performance is pretty bad (tons of CPU consumed since everything re-renders). Could work around this by using memo for all the controls and avoiding creating new arrow functions on each render, but I don't want to do it in this PR.

PR #44, opened 2018-12-21
docs: use retext-smartypants

jtbandes:
Per example at https://github.com/mdx-js/mdx/blob/ad58be384c07672dc415b3d9d9f45dcebbfd2eb8/docs/advanced/retext-plugins.md

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3248,6 +3248,12 @@
         "es-abstract": "^1.7.0"
       }
     },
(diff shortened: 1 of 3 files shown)


jtbandes:
We already use quotes in many places. This includes single apostrophes too. 

Producing build warnings for spelling errors seems nice, though I wonder if it’s easy to disable for false positives?

PR #45, opened 2018-12-22
Make demos for some primitives consistent

janpaul123:
It’s nice if the demos can just be copy-pasted and work. Also nice if
they don’t have magic numbers and if they’re consistent with each other.
I took a pass at some of the simpler components and made them basically
share the exact same simple code.

Also added a simple .import.js file for people who use
[import-js](https://github.com/Galooshi/import-js).


docs/src/jsx/Cones.js
```diff
@@ -4,47 +4,26 @@
...
+      color: { r: 1, g: 0, b: 1, a: 0.5 },
     },
-    scale: p(scaleX, scaleY, scaleZ),
-    color: { r: 1 - range * 0.5, g: range, b: 1, a: 1 - range * 0.3 },
-  };
+  ];
 
   return (
     <Worldview>
-      <ConeControls
```
vidaaudrey: The ConeControls was added because the scale x/y/z represent different meanings and it's nice to learn about it with the controls.

janpaul123:
> Not quite sure how .importjs.js is going to be used, could you explain a little in the PR description?

Ah yeah it's just a configuration file for people who use [import-js](https://github.com/Galooshi/import-js) (like me 😛). I'll add a note.

PR #46, opened 2018-12-24
Move Flow under API section

vidaaudrey:
- Move Flow docs to API section
- Move a few small util components under `/utils` directory 

image


docs/src/3.16FlowTypes.mdx
```diff
@@ -0,0 +1,38 @@
+import GithubLink from './utils/GithubLink'
+
+# Flow types
+
+Flow types are supported. To use the type, import directly from `regl-worldview`. More type definitions: <GithubLink href="https://github.com/cruise-automation/webviz/blob/master/packages/regl-worldview/src/types/index.js" />
```
janpaul123: "To use the type, import directly from" => "To use types, import them directly from"?

janpaul123: Also, instead of `More type definitions: <github icon link>` what about making the whole sentence "More type definitions." a link (maybe prefixed with the github icon)?

vidaaudrey: Makes sense. Probably need more design refinement later
image

PR #48, opened 2018-12-25
Clean up numbers and demo utils

vidaaudrey:
- Remove complicated numbers from demos 
- Remove the react-live scope dependency for utils so we could build a new feature to copy-paste the demo code directly 
- Allow `Points` to accept Vec3 `points` as prop


docs/src/jsx/LinesInstability.js
```diff
@@ -6,33 +6,32 @@
...
+    { x: 812, y: 2959, z: 0 },
+    { x: 812, y: 2959, z: 0 },
+    { x: 812, y: 2959, z: 0 },
+    { x: 812, y: 2959, z: 0 },
+    { x: 812, y: 2959, z: 0 },
+    { x: 812, y: 2959, z: 0 },
+    { x: 812, y: 2959, z: 0 },
+    { x: 812, y: 2959, z: 0 },
+    { x: 812, y: 2959, z: 0 },
+    { x: 812, y: 2958, z: 0 },
```
janpaul123: This whole line instability business shouldn't be a demo right? I thought it was intended as a regression test for an edge case and should probably only exist as a screenshot test. I wonder if we should touch it at all — maybe messing with these numbers makes it not test the edge case properly.

vidaaudrey: Good point. I'll leave the demo unchanged.

PR #50, opened 2018-12-25
Docs: add optional column and use code style for names and defaults

vidaaudrey:
- Add `optional` column to some props tables
- Use code style for names and defaults in props tables
- Minor left-over work from prev PR #49. Somehow I forgot to push up yesterday

### DIFF
--- a/docs/src/3.1.Worldview.mdx
+++ b/docs/src/3.1.Worldview.mdx
@@ -2,20 +2,20 @@
 
 ## Props
 
-| Name                | Optional | Type                     | Default              | Description                                                                                             |
-| ------------------- | -------- | ------------------------ | -------------------- | ------------------------------------------------------------------------------------------------------- |
-| cameraState         | yes      | `$Shape<CameraState>`    |                      | control the [cameraState](/api/camera), must use with `onCameraStateChange`                             |
-| onCameraStateChange | yes      | `?(CameraState) => void` |                      | callback whenever the camera state changes                                                              |
-| defaultCameraState  |          | `$Shape<CameraState>`    | DEFAULT_CAMERA_STATE | pass in this prop to turn Worldview to an uncontrolled component. `onCameraStateChange` will be ignored |
-| backgroundColor     |          | `Vec4`                   | [0,0,0,1]            | change the scene background color                                                                       |
-| hitmapOnMouseMove   |          | `boolean`                | false                | if enabled, the clicked objectId will be returned from the mouse event handler                          |
-| children            | yes      | `React.Node`             |                      | regl components to render inside the Worldview                                                          |
-| showDebug           | yes      | `?boolean`               |                      | show debug information                                                                                  |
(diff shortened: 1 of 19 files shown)


docs/src/3.1.Worldview.mdx
```diff
@@ -2,20 +2,20 @@
...
+| --------------------- | -------- | ------------------------ | ---------------------- | ------------------------------------------------------------------------------------------------------- |
+| `cameraState`         | yes      | `$Shape<CameraState>`    |                        | control the [cameraState](/api/camera), must use with `onCameraStateChange`                             |
+| `onCameraStateChange` | yes      | `?(CameraState) => void` |                        | callback whenever the camera state changes                                                              |
+| `defaultCameraState`  |          | `$Shape<CameraState>`    | `DEFAULT_CAMERA_STATE` | pass in this prop to turn Worldview to an uncontrolled component. `onCameraStateChange` will be ignored |
+| `backgroundColor`     |          | `Vec4`                   | `[0,0,0,1]`            | change the scene background color                                                                       |
+| `hitmapOnMouseMove`   |          | `boolean`                | `false`                | if enabled, the clicked objectId will be returned from the mouse event handler                          |
```
janpaul123: All these three are also optional right? (After all, they have defaults.)

janpaul123: Maybe we should just note that all fields are optional?

docs/src/3.3.Command.mdx
```diff
@@ -5,9 +5,9 @@ import CommandDemo from './jsx/Command.js';
 Command is the base class for all regl-based drawing commands. Subclasses should override getDrawProps and (and optionally getHitmapProps). Note: getHitmapProps is not implemented in this base class.
 Checkout [Writing A Custom Command Component](/guides/writing-a-custom-command-component) for more.
 
-| Name       | type    | Default | Description                                                                       |
-| ---------- | ------- | ------- | --------------------------------------------------------------------------------- |
-| layerIndex | ?number |         | defines the draw order of each command, e.g. -1 will put the shape in lower layer |
+| Name         | optional | type      | Default | Description                                                                       |
+| ------------ | -------- | --------- | ------- |
+| `layerIndex` | yes      | `?number` |         | defines the draw order of each command, e.g. -1 will put the shape in lower layer |
```
janpaul123: Don't need the question mark any more if we explicitly note it's optional right?

docs/src/3.1.Worldview.mdx
vidaaudrey: Good idea. Will remove the column and add the note instead.

docs/src/3.3.Command.mdx
vidaaudrey: Our type definition wasn't quite strict before. Technically we allow `layIndex={null}` to be passed in since the flow definition is `layerIndex?: ?number`. But I think we should just have strict definition `layerIndex?: number` instead. Will update the Command component and docs both.

PR #51, opened 2018-12-26
Add theme config to globally control colors, fonts and spacing

vidaaudrey:
- Add theme config for colors, fonts and spacing. The settings are mostly based on our internal styleguide, with minor adjustments. 
- Update existing components to use the theme config 

Note: more design updates from Eden/Jeff will come later.

### DIFF
--- a/docs/src/theme.js
+++ b/docs/src/theme.js
@@ -46,7 +46,7 @@ const opacities = [
   { value: 30, hex: "4D" },
   { value: 20, hex: "33" },
   { value: 10, hex: "1A" },
-  { value: "05", hex: "0D" },
+  { value: 5, hex: "0D" },
 ];
 
 const palette = {};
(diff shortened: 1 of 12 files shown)


docs/src/theme.js
```diff
@@ -0,0 +1,134 @@
...
+  { value: 90, hex: "E6" },
+  { value: 80, hex: "CC" },
+  { value: 70, hex: "B3" },
+  { value: 60, hex: "99" },
+  { value: 50, hex: "80" },
+  { value: 40, hex: "66" },
+  { value: 30, hex: "4D" },
+  { value: 20, hex: "33" },
+  { value: 10, hex: "1A" },
+  { value: "05", hex: "0D" },
```
janpaul123: Why is value a string here?

janpaul123: Ah I see the function below now. Maybe just make them all strings for consistency then? A bit confusing otherwise.

vidaaudrey: Good catch! Actually `color.white5' should be valid and straightforward,  we can just remove the string and keep all numbers.

PR #57, opened 2019-01-04
Make <Command /> usage not require inheritance.

brianc:
The react team recommends not inheriting from components but rather using them through composition.

I feel like this simplifies using `<Command />` in that you don't have to implement particular methods on a subclass, you just call `<Command />` with the props you want to render.

I wanted to open this PR for discussion.  I still need to update the docs to reflect the new API before this is merged.

Now when one implements a custom command it would look like so:

```js
const reglCommand = {
  vert: ...
  frag: ...
}

export class MyCommand extends React.Component {
  render() {
    return <Command reglCommand={reglCommand} drawProps={...} hitmapProps={...} />
  }
}
```

Instead of having `MyCommand` extend `Command` directly.


packages/regl-worldview/src/commands/Command.js
```diff
@@ -9,79 +9,75 @@
...
 
-type CommandProps = {
-  layerIndex?: number,
-};
-
-// The base class for all regl-based drawing commands.
-// Subclasses should override getDrawProps and (and optionally getHitmapProps).
-export default class Command<T: CommandProps> extends React.Component<T> {
+// Component to dispatch draw props and hitmap props and a reglCommand to the render loop to render with regl.
+export default class Command extends React.Component<Props<T>> {
```
jtbandes: Where did `<T>` go? Surprised Flow is happy with this as is

packages/regl-worldview/src/commands/Command.js
```diff
@@ -9,79 +9,75 @@
...
-    return null;
-  }
-
   render() {
     return (
       <WorldviewReactContext.Consumer>
         {(ctx: ?WorldviewContextType) => {
           if (ctx) {
             this.context = ctx;
+            this._updateContext(ctx);
```
jtbandes: Why move this here from `componentDidUpdate`?

packages/regl-worldview/src/commands/Command.js
```diff
@@ -90,45 +86,38 @@ export default class Command<T: CommandProps> extends React.Component<T> {
...
+// Sample usage: const Cubes = makeCommand('Cubes', rawCommand)
+// When you have children as the drawProps input, it's useful to simply call makeCommand
+// which creates a new regl component. It also handles basic hitmap interactions.
+export function makeCommand<T>(name: string, command: RawCommand<T>) {
+  const cmd = (props: Props<T>) => {
+    return (
+      <Command
+        reglCommand={command}
+        drawProps={props.children}
+        hitmapProps={getHitmapProps(props.getHitmapId, props.children)}
```
jtbandes: A little unfortunate to have to run this every time even if the hitmap is not going to be used. On the other hand, I would eventually like to combine the hitmap props with regular props anyway and avoid doing any `map`.

brianc: yeah - and `getHitmapProps` is a no-op if `props.getHitmapId` is not supplied.  I could do: `hitmapProps={props.getHitmapId ? getHitmapProps(props.getHitmapId, props.children) : undefined}` to skip the function call.

brianc:
> Can you remember why we didn't originally do this?

Didn't think of this approach at the time is all.

packages/regl-worldview/src/commands/Command.js
brianc: I moved it back

packages/regl-worldview/src/commands/Command.js
brianc: This file isn't currently flow checked. There's a TODO comment at the top to flow check it.

docs/src/3.3.Command.mdx
```diff
@@ -13,48 +13,46 @@ Sample usage:
...
     }
+  `
+  frag: `
+    precision mediump float;
+    uniform vec4 color;
+    void main () {
+      gl_FragColor = color;
+  }`,
+  attributes: {
+    position: (context, props) => props.points,
```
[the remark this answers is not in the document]
brianc: done

docs/src/3.3.Command.mdx
```diff
@@ -13,48 +13,46 @@ Sample usage:
...
-      }`,
-      attributes: {
-        position: (context, props) => props.points,
-      },
-      uniforms: {
-        color: (context, props) => props.color,
-      }
-      count: (context, props) => props.points.length,
+// define our regl command
+const reglTriangle = () => ({
```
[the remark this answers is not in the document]
brianc: Made it take `regl` as an argument to use `regl.prop` and yea it does require the command to be a function currently.

docs/src/jsx/Command.js
```diff
@@ -8,15 +8,36 @@ import React from "react";
...
+  attributes: {
+    position: (context, props) => props.points,
+  },
+  uniforms: {
+    color: (context, props) => props.color,
+  },
+  count: (context, props) => props.points.length,
+});
+
+class Triangle extends React.Component {
```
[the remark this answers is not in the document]
brianc: done

PR #58, opened 2019-01-11
Allow custom getHitmapProps() function to be passed to simple commands.

brianc:
Also export WorldviewReactContext so non-command components can grab access to the regl instance.


packages/regl-worldview/src/index.js
```diff
@@ -16,6 +16,7 @@ export { default as fromGeometry } from "./utils/fromGeometry";
 export * from "./utils/Raycast";
 export * from "./commands/index";
 export * from "./types/index";
+export { default as WorldviewReactContext } from "./WorldviewReactContext";
```
jtbandes: Cool, I was just thinking this might be handy in a command I'm working on (specifically the ability to call `onDirty()`)

PR #60, opened 2019-01-16
Fix inconsistent updates in camera rotation

vidaaudrey:
Problem: currently we  consolidate `cameraState` updates in `_setStateAndNotify ` and skip updating `cameraState` if the next state is deeply equal to the current state. This approach has two issues:
- potential race condition as `cameraState` can be set elsewhere during the movement (e.g. worldview props)
- not efficient because we only need to compare a few primitives before updating the `cameraState`: `x`, `y`, `distance`. 

Test plan: test the new esm build internally, we should be able to rotate the 3d view when the webviz bag play is paused. 

Note: upon further investigation, we found the root problem is related to object spread in Chrome. Changing `const newState = {...this.state, thetaOffset: newThetaOffset}` to `const newState = Object.assign({}, this.state); newState.thetaOffset = newThetaOffset` work.  Canary and Firefox don't have any issues.

### DIFF
--- a/packages/regl-worldview/src/camera/CameraStore.js
+++ b/packages/regl-worldview/src/camera/CameraStore.js
@@ -56,6 +56,7 @@ export default class CameraStore {
   };
 
   cameraRotate = ([x, y]: Vec2) => {
+    // This can happen in 2D mode, when both e.movementX and e.movementY are evaluated as negative and mouseX move is 0
     if (x === 0 && y === 0) {
       return;
     }
@@ -70,6 +71,7 @@ export default class CameraStore {
 
   // move the camera along x, y axis; do not move up/down
   cameraMove = ([x, y]: Vec2) => {
+    // moveX and moveY both be 0 sometimes
     if (x === 0 && y === 0) {
       return;
     }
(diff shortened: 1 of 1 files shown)


packages/regl-worldview/src/camera/CameraStore.js
```diff
@@ -56,47 +55,50 @@ export default class CameraStore {
...
-    // we only want to call onChange when cameraState values actually changes, as each
-    // onChange will trigger Worldview to rerender by forceUpdate or onCameraStateChange
-    if (!isEqual(this.state, cameraState)) {
-      this.state = cameraState;
-      this._onChange(this.state);
-    }
-  }
-
   cameraRotate = ([x, y]: Vec2) => {
+    if (x === 0 && y === 0) {
```
jtbandes: When will this happen?

vidaaudrey: e.g. in 2d mode, when `e.movementX` and `e.movementY` are both evaluated as negative , and mouse x move is 0: https://github.com/cruise-automation/webviz/blob/master/packages/regl-worldview/src/camera/CameraListener.js#L157

jtbandes:
For posterity, the actual bug here was a Chrome bug where the use of `{...this.state, thetaOffset, phi}` was actually mutating `this.state`. It only reproduced in our internal build of webviz, and Firefox didn't exhibit the same bug. The latest Chrome Canary seems to work fine too, so perhaps the issue has been fixed. This PR avoids the problem by removing the isEqual check.

PR #64, opened 2019-01-21
Add <GLTFScene>, loadGLB, and Duck example

jtbandes:
I had to do one strange thing here, which was to make `duckModel` (the URL) a global variable in the DuckScene.js example code, because apparently `LivePreview` doesn't support passing props into the rendered component. @vidaaudrey do you have a better idea for how to do this?

image


docs/src/jsx/allDemos.stories.js
```diff
@@ -50,6 +52,7 @@ const allDemos = {
   SpheresSingle,
   Text,
   Triangles,
+  DuckScene: () => <DuckScene modelURL={duckModel} />,  // eslint-disable-line
```
brianc: nit: might be nice to use `eslint-disable-line <eslint-rule-that-fails>` as a small form of documentation as to what lint rule you're opting out of.

packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -0,0 +1,210 @@
...
+  function prepareDrawCallsIfNeeded(model) {
+    if (drawCalls) {
+      return;
+    }
+
+    const textures = model.json.textures.map((texture) => {
+      const sampler = model.json.samplers[texture.sampler];
+      return regl.texture({
+        data: model.images[texture.source],
+        min: "linear", //glConstantToRegl(sampler.minFilter),
```
brianc: what's the commented out part here for?

packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -0,0 +1,210 @@
...
+export default class GLTFScene extends React.Component<Props, *> {
+  state = {
+    model: undefined,
+  };
+  _context = undefined;
+
+  componentDidMount() {
+    loadGLB(this.props.modelURL).then((model) => {
+      this.setState({ model });
+      if (this._context) {
```
brianc: This is an unfortunate oversight in the API that we need to call `this._context.onDirty` when the child `<Command >` props change.  This is totally fine for now, but just fyi I have some unfinished work to make this unneeded in the future.  IMO if `<Command />` receives new props it should trigger a draw call.  To prevent over-rendering the parent should implement `shouldComponentUpdate` if needed.  I think this is more in line w/ how one would expect react to work: render unless told specifically not to.

packages/regl-worldview/src/utils/loadGLB.js
```diff
@@ -0,0 +1,126 @@
...
+      bufferView.byteLength / arrayType.BYTES_PER_ELEMENT
+    );
+  });
+
+  // load embedded images
+  const images = await Promise.all(
+    json.images.map(
+      (imgInfo) =>
+        new Promise((resolve, reject) => {
+          const bufferView = json.bufferViews[imgInfo.bufferView];
```
brianc: as a future enhancement we might be able to use [createImageBitmap](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/createImageBitmap) here and let the browser do a touch more optimized image decoding.

docs/src/4.12.GLTFScene.mdx
```diff
@@ -0,0 +1,19 @@
+import {DuckScene} from './jsx/allLiveEditors';
+
+# GLTFScene
+
+`<GLTFScene />` provides support for loading and rendering [glTF models](https://www.khronos.org/gltf/). Only **Binary glTF** (`.glb`, without external resources) and basic color textures are currently supported.
+
+## Props
+
+| Name       | Type       | Default | Description                       |
```
vidaaudrey: we could format the table so the cells are aligned

docs/src/jsx/DuckScene.js
```diff
@@ -0,0 +1,39 @@
+//  Copyright (c) 2019-present, GM Cruise LLC
```
vidaaudrey: Wonder if we should update existing file headers to be `2019` as well?

packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -0,0 +1,210 @@
...
+  void main() {
+    vec3 baseColor = texture2D(baseColorTexture, vTexCoord).rgb;
+    float diffuse = light.diffuseIntensity * max(0.0, dot(vNormal, -light.direction));
+    gl_FragColor = vec4((light.ambientIntensity + diffuse) * baseColor, 1);
+  }
+  `,
+  });
+
+  let drawCalls;
+  function prepareDrawCallsIfNeeded(model) {
```
vidaaudrey: would be nice to add some comments

packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -0,0 +1,210 @@
...
+      command(drawCalls);
+    });
+  };
+};
+
+type Props = {|
+  modelURL: string,
+  children: {|
+    pose: Pose,
+    scale: Point,
```
vidaaudrey: We do have type Scale :)

packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -0,0 +1,210 @@
...
+  |},
+|};
+
+export default class GLTFScene extends React.Component<Props, *> {
```
vidaaudrey: could we have better flow type for `state` ?

packages/regl-worldview/src/commands/GLTFScene.js
jtbandes: For our car model this made some of the textures look bad when downsampling since they chose the “best” mip level. I don’t totally understand how mipmaps are supposed to work, so this is just a workaround that makes it look slightly better, although the downside is now it looks *too* sharp sometimes 🤷‍♂️

I guess I could add a prop to override this. Or maybe a prop to allow mutating/changing the model after it’s loaded...what do you think?

docs/src/jsx/DuckScene.js
janpaul123: Haha oh no how is this going to work with our linter that checks for one consistent header (at least in webviz-core)! 😮

packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -0,0 +1,210 @@
...
+  return (props) => {
+    prepareDrawCallsIfNeeded(props.model);
+    withPoseMatrix(props, () => {
+      command(drawCalls);
+    });
+  };
+};
+
+type Props = {|
+  modelURL: string,
```
janpaul123: Hm why not make this part of `children` too? That seems to be the way that we do things in other commands?

jtbandes: I kept this as a separate prop now that it can be either `string` or `() => Promise<Object>`.

jtbandes:
> possibly calling the component <GTLF /> or something w/o the world "scene" in it. To me "scene" kinda implies its the whole scene and not one of many things within the 3d scene

My reasoning for this is that it currently does render the whole `scene` (a glTF concept, the hierarchy is `scenes > nodes > meshes > primitives` and while the whole file can define multiple scenes, only one of them is the ["default scene"](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#gltfscene) that [should be rendered](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#scenes)). Maybe in the future we could expand the API to let you pick out certain nodes or non-default scenes to render, but for now rendering the singular default scene seemed like the simplest way to go.

PR #65, opened 2019-01-22
Add browser support to docs and update umd link

vidaaudrey:
- Add a simple browser support sections to docs 
- Update umd link which will automatically redirect to the latest minified regl-worldview umd build

### DIFF
--- a/docs/src/3.5.BrowserSupport.mdx
+++ b/docs/src/3.5.BrowserSupport.mdx
@@ -1,0 +1,10 @@
+# Browser Support
+
+Currently we support the following browsers/versions:
+
+- Chrome 71+
+- Firefox 64+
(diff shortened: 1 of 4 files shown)


docs/src/3.5.BrowserSupport.mdx
```diff
@@ -0,0 +1,10 @@
+# Browser Support
+
+Currently we support the following browsers/versions:
+
+- Chrome 71+
+- Firefox 64+
+- Desktop Safari 11+
+- Opera 57+
+
+Some unsupported browsers may require polyfill to work. We'll add more support in the future.
```
janpaul123: Small language thing: "polyfill" -> either "a polyfill" or "polyfills" (probably the latter makes most sense?)

PR #66, opened 2019-01-24
Prepare to move RPC into an npm module

brianc:
This copies the RPC code from [here](https://github.com/cruise-automation/webviz/blob/master/packages/webviz-core/src/util/Rpc.js) into a separate package for publishing.  I need to reuse the RPC code in another internal project.

I also found the world `transferrable` to be misspelled compared to the [Interface name](https://developer.mozilla.org/en-US/docs/Web/API/Transferable)...so I changed it.

Upgrade plan:
- publish this to npm
- install npm module in internal webviz repo, remove `Rpc.js` from internal utils folder
- update places where `Rpc.transferrable` is used in code to reflect new spelling
- push new version of internal webviz to this mirror

Test plan: unit test coverage is provided.


packages/@cruise-automation/worker-rpc/README.md
```diff
@@ -0,0 +1,74 @@
+# @cruise-automation/worker-rpc
+
+An RPC layer to make it easier to communicate between a WebWorker and the main thread.  Has support for sending and responding with [transferable](https://developer.mozilla.org/en-US/docs/Web/API/Transferable) objects to avoid structured cloning of large array buffers.  It also propagates errors thrown in receivers back to the calling thread.
```
vidaaudrey: `An RPC...` => `A RPC...`. 
I saw that you referred `WebWorker(s)` a few times in the doc text, I'm wondering if we can just call it `web workers` so it's easily to associate with native web worker concept.

vidaaudrey: `An RPC...` => `A RPC`

packages/@cruise-automation/worker-rpc/src/index.js
```diff
@@ -0,0 +1,138 @@
...
+// Errors in receivers are propigated back to the caller as a rejection.
+// It also supports returning transferables over the web-worker postMessage api,
+// which was the main shortcomming with the worker-rpc npm module.
+// To attach rpc to an instance of a worker in the main thread:
+//   const rpc = new Rpc(workerInstace);
+// To attach rpc within an a web worker:
+//   const rpc = new Rpc(global);
+// Check out the tests for more examples.
+export default class Rpc {
+  static transferables = "$$TRANSFERABLES";
```
jtbandes: is it possible to use `Symbol` for this, or does that break with MessagePort?

packages/@cruise-automation/worker-rpc/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "@cruise-automation/worker-rpc",
```
janpaul123: I wonder if we should just call it `@cruise-automation/rpc` to allow for using it over non-WebWorker channels in the future (e.g. websockets, like we discussed, or even things like [`BroadcastChannel`](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API) to communicate between webworkers, or different Node processes, etc)

packages/@cruise-automation/worker-rpc/src/index.js
```diff
@@ -0,0 +1,138 @@
...
+        throw new Error(`no receiver registered for ${topic}`);
+      }
+      // This works both when `handler` returns a value or a Promise.
+      resolve(handler(data));
+    })
+      .then((result) => {
+        if (!result) {
+          return this._channel.postMessage({ topic: RESPONSE, id });
+        }
+        const transferables = result[Rpc.transferables];
```
janpaul123: I wonder if instead of using this magic string we can just get rid of it and automatically detect transferables by walking the object?

packages/@cruise-automation/worker-rpc/src/index.js
brianc: Good call - should be fine since we don't actually _send_ this key across the message port. As far as it working across message ports I'd guess it _doesn't_ since it creates a unique value for every constructor invocation and there is a separate instance of `rpc` on both sides of the boundary.

packages/@cruise-automation/worker-rpc/package.json
brianc: renamed to `@cruise-automation/rpc`

packages/@cruise-automation/worker-rpc/src/index.js
brianc: Yeah thought about it but there are a few issues w/ that.  Recursively walking the object graph can be expensive.  Extra expensive since structured cloning supports circular references so we'd need a traversal algorithm that supports breaking on circular references.  Since the point of transferables is to be a speed optimization I'm okay w/ having it be an explicit if somewhat more awkward API for performance critical code.  This also more closely parrots the built-in APIs which make you declare what you want to transfer across.  There's also the concern that when you transfer something it becomes detached in the calling code so it could be a surprise if you _want_ to copy a small array buffer or image bitmap and it disappears after you return it from your function and throws next time you access it.

packages/@cruise-automation/worker-rpc/src/index.js
brianc: ehh...flow doesn't like using a symbol for a computed object property.  While it's technically valid JS I'd rather avoid using it since we'd have to attach `// $FlowFixMe` everywhere we used this.  I don't see it being a huge benefit versus the cost, so I'll just leave it as a string for now.  

https://github.com/facebook/flow/issues/3258

PR #71, opened 2019-01-28
fix keyMap behavior by using event.code; add shiftKeys

jtbandes:
cc @cdbarlow

#56 switched from [`e.keyCode`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) to [`e.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key), which caused a problem where keys could get "stuck" down. For example, after `press A, press Shift, release A, release Shift` the CameraListener would think the A key was still pressed.

`e.key` is explicitly supposed to take into account modifier keys such as shift & option, because it represents the character that would be typed when pressing the key, so using it to track key-up/down state doesn't work well. In this PR I'm switching to [`e.code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) instead, which is a more newly standardized property that is independent of keyboard layout (it always works as if the layout were QWERTY) and intended to be used when you care about the key's position rather than the character it represents.

This PR also ignores key repeats, removes an unused `onKeyDown` prop, and fixes a bug where the 1/10 magnitude adjustment for the shift key was applied twice, which made keyboard movement with the shift key even slower than intended.

Also adds a `shiftKeys={/*boolean*/}` to disable shift key modifiers.


packages/regl-worldview/src/camera/CameraListener.js
```diff
@@ -309,45 +312,45 @@ export default class CameraListener extends React.Component<Props> {
...
-    const { onKeyDown, keyMap } = this.props;
+  _onKeyDown = (e: SyntheticKeyboardEvent<HTMLDivElement>) => {
+    const { keyMap } = this.props;
     this._shiftKey = e.shiftKey;
     this._metaKey = e.metaKey;
     this._ctrlKey = e.ctrlKey;
+    const code = ((e.nativeEvent: any): KeyboardEvent).code;
+
+    // ignore repeated keydown events
+    if (e.repeat || this._keys.has(code)) {
```
janpaul123: `e.repeat` TIL!

docs/src/3.1.Worldview.mdx
```diff
@@ -8,6 +8,7 @@
 | `onCameraStateChange` | `(CameraState) => void` |                        | callback whenever the camera state changes                                                              |
 | `defaultCameraState`  | `$Shape<CameraState>`   | `DEFAULT_CAMERA_STATE` | pass in this prop to turn Worldview to an uncontrolled component. `onCameraStateChange` will be ignored |
 | `keyMap`              | `CameraKeyMap`          | {}                     | override default Worldview hotkeys                                                                      |
+| `shiftKeys`           | `boolean`               | true                   | enable/disable shift modifier for keyboard controls                                                     |
```
vidaaudrey: nit: would be nice to alway default boolean to `false` / `undefined`.  `shiftKeys` is very straightforward to understand, how about something like `disableSlowMotionWithShiftKey`?

packages/regl-worldview/src/camera/CameraListener.js
```diff
@@ -214,12 +213,16 @@ export default class CameraListener extends React.Component<Props> {
...
-  getKeyMotion = (key: string): ?KeyMotion => {
+  _getKeyMotion = (code: string): ?KeyMotion => {
     const moveSpeed = this._getMagnitude(KEYBOARD_MOVE_SPEED);
     const zoomSpeed = this._getMagnitude(KEYBOARD_ZOOM_SPEED);
     const spinSpeed = this._getMagnitude(KEYBOARD_SPIN_SPEED);
     const { keyMap } = this.props;
-    const action: CameraAction | false = (keyMap && keyMap[key]) || DEFAULT_KEYMAP[key] || false;
+    const action: CameraAction | false = (keyMap && keyMap[code]) || DEFAULT_KEYMAP[code] || false;
+
+    if (this._shiftKey && !this.props.shiftKeys) {
```
vidaaudrey: nit: can destruct `shiftKeys` props at  line 220

docs/src/3.1.Worldview.mdx
jtbandes: Chose this name to allow possible future expansion with an object, like `shiftKeys={{ KeyA: ..., KeyS... }}`

PR #73, opened 2019-02-04
 Add 'Open in CodeSandbox' link

vidaaudrey:
- add `Open in CodeSandbox` link to the demo code 
- prefill the sandbox with packages and js files for all examples (had to switch to POST form from GET request due to the file size https://codesandbox.io/docs/importing#define-api )
- update the subroute url so instead of `/basicexample`, we'll have `/basic-example`
- move non-demo files under `/jsx/utils` for simple sandbox structure 
- add an  `Open in CodePen` for UMD demo: https://codepen.io/vidaaudrey/pen/WPwmqg. 

bag


docs/src/jsx/utils/CodeSandboxLink.js
```diff
@@ -0,0 +1,70 @@
...
+import styled from "styled-components";
+
+import CodeSandboxIcon from "./icons/CodeSandbox";
+import { palette } from "./theme";
+
+function compress(string) {
+  return LZString.compressToBase64(string)
+    .replace(/\+/g, `-`) // Convert '+' to '-'
+    .replace(/\//g, `_`) // Convert '/' to '_'
+    .replace(/=+$/, ``); // Remove ending '='
```
brianc: Why are these replaces here? A comment would be helpful to explain the reasoning.

vidaaudrey: Good catch. it's not really necessary there. I was using the GET request in the beginning which has some restrictions on how urls are formatted. Later when I switched to the POST request, I forgot to remove these string manipulations. I'l just remove to avoid confusion.

PR #74, opened 2019-02-04
Add @babel/runtime as prod dependency

vidaaudrey:
We use @babel/transform-runtime in the [rollup config](https://github.com/cruise-automation/webviz/blob/master/packages/regl-worldview/rollup.config.js#L22) to reduce the codesize. As specified in the doc [here](https://babeljs.io/docs/en/babel-plugin-transform-runtime), we need to add @babel/runtime as prod dependency. 

Test: 
CodeSandBox only works if this dependency is present. https://codesandbox.io/s/j322yv070v 
Will publish a new package version to test once the PR is merged.

### DIFF
--- a/packages/regl-worldview/package-lock.json
+++ b/packages/regl-worldview/package-lock.json
@@ -4,6 +4,14 @@
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
+		"@babel/runtime": {
+			"version": "7.3.1",
+			"resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.3.1.tgz",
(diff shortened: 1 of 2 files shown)


vidaaudrey:
> How much code size did we save by using the runtime? Maybe it's easier to just remove that transform?

Right now it saves 1kb for cjs and esm build. Thought would be good to keep it as it not only reduces the code size, but also avoid polluting global scope (v.s. using polyfill). The general recommendation is:  
 > If you're building an app, you can use babel-polyfill. If you're building a library, make sure not to use babel-polyfill, and only the transform-runtime.

 https://codersmind.com/babel-polyfill-babel-runtime-explained/

PR #75, opened 2019-02-05
Component-based mouse interactions

robin-pham:
This PR adds component-based mouse interactions so that consumers can add `onClick` or `onMouseMove` etc. to a command component.

The story added showcases 3 interactions, clicking for details, double clicking for removal, and hovering for changing the mouse cursor.
out

This takes a similar approach from a previous PR, without a lot of the extra overhead.

### DIFF
--- a/packages/regl-worldview/src/Worldview.js
+++ b/packages/regl-worldview/src/Worldview.js
@@ -8,6 +8,7 @@
 
 import mapValues from "lodash/mapValues";
 import pickBy from "lodash/pickBy";
+import size from "lodash/size";
 import * as React from "react";
 import ContainerDimensions from "react-container-dimensions";
 
@@ -18,6 +19,8 @@ import { WorldviewContext } from "./WorldviewContext";
 import WorldviewReactContext from "./WorldviewReactContext";
 
 const DEFAULT_BACKGROUND_COLOR = [0, 0, 0, 1];
+
+export const MOUSE_HANDLERS = ["onClick", "onMouseUp", "onMouseMove", "onMouseDown", "onDoubleClick"];
 
 export type BaseProps = {|
   keyMap?: CameraKeyMap,
@@ -44,7 +47,7 @@ type State = {|
   worldviewContext: WorldviewContext,
 |};
 
-function handleMouseInteraction(objectId: number, ray: Ray, e: MouseEvent, handler: MouseHandler) {
+function handleWorldviewMouseInteraction(objectId: number, ray: Ray, e: MouseEvent, handler: MouseHandler) {
   try {
     handler(e, {
       ray,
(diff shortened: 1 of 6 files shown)


packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -69,6 +70,11 @@ export class WorldviewContext {
   _drawCalls: Map<React.Component<any>, any> = new Map();
   _hitmapCalls: Map<React.Component<any>, any> = new Map();
   _paintCalls: Map<PaintFn, PaintFn> = new Map();
+  mouseHandlers: {
```
vidaaudrey: would be nice to type this explicitly:
```
mouseHandlers: {
    [handlerName: string]: {
      [hitmapId: number]: [React.Component<any>, any],
    },
  };
```

packages/regl-worldview/src/commands/Command.js
```diff
@@ -121,7 +121,7 @@ export function makeCommand<T>(name: string, command: RawCommand<T>): React.Stat
     const hitmapProps = props.getHitmapProps
       ? props.getHitmapProps()
       : getHitmapProps(props.getHitmapId, props.children);
-    return <Command reglCommand={command} drawProps={props.children} hitmapProps={hitmapProps} />;
+    return <Command {...props} reglCommand={command} drawProps={props.children} hitmapProps={hitmapProps} />;
```
vidaaudrey: cc @brianc  this will fix the issue of `layerIndex`  not being passed down

packages/regl-worldview/src/Worldview.js
```diff
@@ -187,15 +186,24 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
...
-      .then((objectId) => handleMouseInteraction(objectId, ray, e, handler))
+      .then((objectId) => {
+        if (worldviewHandler) {
+          handleWorldviewMouseInteraction(objectId, ray, e, worldviewHandler);
+        }
+        const componentLevelHandler = worldviewContext.mouseHandlers[handlerName][objectId];
+        if (componentLevelHandler) {
+          const [component, drawProp] = componentLevelHandler;
+          component.props[handlerName](objectId, ray, e, drawProp);
+        }
```
brianc: so this means if the component has a mouse handler and worldview also has a handler for the same event they'll both fire.  Should we support `e.stopPropagation` in the component level handler?

robin-pham: hmm, doesn't seem straightforward to listen into an event to see if `stopPropagation` was called. We'd have to add our own state to each event if we want to do something similar. How often do ya think this use case will happen?

robin-pham:
@brianc @vidaaudrey I changed the paradigm so that `WorldviewContext` lets each component know that an event fired. This way we don't have to map draw prop ids -> components everytime the props change for each component (which is how it was before). Now we find draw prop ids on each mouse interaction.

@vidaaudrey is it okay if I add the documentation page in a separate PR? or would you prefer it to be added in this PR?

robin-pham:
> Now we find draw prop ids on each mouse interaction.

if we find this operation is too slow in the future I think it could easily be memoized. or I can just memoize it right now... what y'all think?

robin-pham:
o wait I implemented something a lil bit inefficient, going to change the paradigm again.

packages/regl-worldview/src/stories/PerComponentHitmapStory.js
```diff
@@ -0,0 +1,106 @@
...
+    mouseCursor: sample(randomMouseCursors),
+  };
+};
+
+function PerCubeInteractions() {
+  const [cubes, setCubes] = useState([createCube(0)]);
+  const [cubeDetails, setCubeDetails] = useState({});
+  const [cursor, setCursor] = useState("auto");
+
+  const onContainerClick = (e, args) => {
```
vidaaudrey: would be good to use plain function so it get hoisted :)

packages/regl-worldview/src/commands/Command.js
```diff
@@ -81,6 +82,37 @@ export default class Command<T> extends React.Component<Props<T>> {
...
+    return null;
+  }
+
+  fireRelevantMouseHandler(id: number, e: any, ray: any, mouseEventName: MouseEventEnum) {
+    const mouseHandler = this.props[mouseEventName];
+    if (!mouseHandler) {
+      return;
+    }
+
+    const drawProp = this.getDrawPropFromHitmapId(id);
```
vidaaudrey: The naming is a little confusing, technically we are returning the `hitmapProp`, not the `drawProp`? What's your consideration  on the type of prop to return?  If we are going to handle hitmapId internally, would it make sense to just return the `drawProp`?

packages/regl-worldview/src/Worldview.js
brianc: It looks like you can't read if `e.stopPropagation` has been called on an event.  I think we can add it later if we want to figure out a way at that point.

packages/regl-worldview/src/commands/Command.js
```diff
@@ -81,6 +82,37 @@ export default class Command<T> extends React.Component<Props<T>> {
...
+    if (!drawProp) {
+      return;
+    }
+
+    mouseHandler(e, {
+      ray,
+      interactedObject: drawProp,
+      component: this,
```
brianc: curious: why are we including the component?  I feel like there's not a lot one can do w/ a component instance in a callback.

packages/regl-worldview/src/commands/Command.js
```diff
@@ -20,6 +20,7 @@ export type Props<T> = {
   reglCommand: RawCommand<T>,
   getHitmapId?: (T) => ?number,
   layerIndex?: number,
+  [MouseEventEnum]: ComponentMouseHandler,
```
jtbandes: Shouldn't this be optional? Why doesn't Flow complain about it being missing?

packages/regl-worldview/src/commands/Command.js
robin-pham: good point, lemme rename it to `hitmapProp`.

packages/regl-worldview/src/commands/Command.js
robin-pham: true! very unnecessary

packages/regl-worldview/src/commands/Command.js
```diff
@@ -81,6 +82,37 @@ export default class Command<T> extends React.Component<Props<T>> {
...
+        const drawPropId = getIdFromColor(drawProp.color.map((color) => color * 255));
+        if (drawPropId === id) {
+          return drawProp;
+        }
+      }
+    }
+    return null;
+  }
+
+  fireRelevantMouseHandler(id: number, e: any, ray: any, mouseEventName: MouseEventEnum) {
```
jtbandes: nit: would be nice to stay consistent with variable names between functions, e.g. `objectId` instead of just `id`

naming: maybe `handleMouseEvent` or something?

packages/regl-worldview/src/commands/Command.js
```diff
@@ -81,6 +82,37 @@ export default class Command<T> extends React.Component<Props<T>> {
     }
   }
 
+  getDrawPropFromHitmapId(id: number) {
```
jtbandes: should this be a `_` private function since it's only used internally?

packages/regl-worldview/src/types/index.js
```diff
@@ -185,6 +193,8 @@ export type PolygonType = BaseShape & {
   id: number,
 };
 
+export type MouseEventEnum = "onClick" | "onMouseUp" | "onMouseMove" | "onMouseDown" | "onDoubleClick";
```
jtbandes: It would be great to update the documentation as well 😁

packages/regl-worldview/src/commands/Command.js
robin-pham: it's implicitly optional

PR #76, opened 2019-02-06
Fix process undefined in umd build

vidaaudrey:
The current umd build doesn't work without `const process = {}`  https://codepen.io/vidaaudrey/pen/WPwmqg  

Fix by always checking for `process` and `process.env` before getting NODE_ENV.

### DIFF
--- a/packages/regl-worldview/src/Worldview.js
+++ b/packages/regl-worldview/src/Worldview.js
@@ -205,7 +205,7 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
     const { worldviewContext } = this.state;
     const initializedData = worldviewContext.initializedData;
 
-    if (process.env.NODE_ENV === "production" || !initializedData) {
+    if ((process && process.env && process.env.NODE_ENV === "production") || !initializedData) {
       return null;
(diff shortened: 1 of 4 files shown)


packages/regl-worldview/src/Worldview.js
```diff
@@ -205,7 +205,7 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
     const { worldviewContext } = this.state;
     const initializedData = worldviewContext.initializedData;
 
-    if (process.env.NODE_ENV === "production" || !initializedData) {
+    if ((process && process.env && process.env.NODE_ENV === "production") || !initializedData) {
```
robin-pham: can probably just turn accessing the `NODE_ENV` into a utility function, i.e. `getNodeEnv()`.

lodash also has a utility called `get` where you can specify a path: `_.get(process, ['env', 'NODE_ENV'])`

jtbandes:
Shouldn't the build process cause `process.env.NODE_ENV` to get replaced with a constant? Maybe that's only when `webpack.DefinePlugin` is used?

PR #79, opened 2019-02-13
Allow text markers to specify background and foreground colors

vidaaudrey:
- Allow text markers to specify background and text colors with `colors` field so that the text can be more readable. 
- Add `autoBackgroundColor` prop to `Text` in order to automatically add high-contrast background color
 
bag

### DIFF
--- a/docs/src/4.10.Text.mdx
+++ b/docs/src/4.10.Text.mdx
@@ -6,13 +6,22 @@ import {Text} from './jsx/allLiveEditors'
 
 ## Props
 
-| Name       | Type     | Default | Description                     |
-| ---------- | -------- | ------- | ------------------------------- |
-| `children` | `Text[]` | `[]`    | array of Text markers to render |
+| Name            | optional | Type      | Default | Description                            |
+| --------------- | -------- | --------- | ------- |
+| `children`      |          | `Text[]`  | `[]`    | array of Text markers to render        |
+| `enableBgColor` | yes      | `boolean` |         | show solid background color if enabled |
+
 
 ### Text
 
 ```js
+type Color = {
+  r: number, // between 0 and 1
+  g: number, // between 0 and 1
+  b: number, // between 0 and 1
+  a: number, // between 0 and 1
+}
+
 type Text = {
   pose: {
     position: { x: number, y: number, z: number },
@@ -23,15 +32,18 @@ type Text = {
     y: number,
     z: number
   },
-  color: {
-    r: number, // between 0 and 1
-    g: number, // between 0 and 1
-    b: number, // between 0 and 1
-    a: number, // between 0 and 1
-  },
+  color: Color,
+  colors?: [
+  // text color, only applied when colors.length >= 2
+  Color,
+  // background color, only applied when colors.length >= 2
+  Color
+ ],
   name?: string,
   text: string,
 }
 ```
 
 <Text />
+
+
--- a/docs/src/4.10.Text.mdx
+++ b/docs/src/4.10.Text.mdx
@@ -51,11 +51,11 @@ type Text = {
   },
   color: Color,
   colors?: [
-  // text color, only applied when colors.length >= 2
-  Color,
-  // background color, only applied when colors.length >= 2
-  Color
- ],
+    // text color, only applied when colors.length >= 2
+    Color,
+    // background color, only applied when colors.length >= 2
+    Color
+  ],
   name?: string,
   text: string,
 }
--- a/docs/src/4.10.Text.mdx
+++ b/docs/src/4.10.Text.mdx
@@ -16,7 +16,7 @@ import {Text} from './jsx/allLiveEditors'
 There are three ways to control the text and background colors:
 
 - Text without background color: set `color` field
-- Text with automatically background color: set `color` field and set `autoBackgroundColor` to `true`
+- Text with automatically-chosen background color: set `color` field and set `autoBackgroundColor` to `true`
 - Custom text and background colors: set `colors` field where `colors[0]` is the text color and `colors[1]` is the background color
 
 Examples:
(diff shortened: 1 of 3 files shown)


docs/src/4.10.Text.mdx
```diff
@@ -6,13 +6,22 @@ import {Text} from './jsx/allLiveEditors'
 
 ## Props
 
-| Name       | Type     | Default | Description                     |
-| ---------- | -------- | ------- | ------------------------------- |
-| `children` | `Text[]` | `[]`    | array of Text markers to render |
+| Name            | optional | Type      | Default | Description                            |
+| --------------- | -------- | --------- | ------- |
+| `children`      |          | `Text[]`  | `[]`    | array of Text markers to render        |
+| `enableBgColor` | yes      | `boolean` |         | show solid background color if enabled |
```
brianc: I would probably call this the full words: `enableBackgroundColor`, though maybe we could change the API slightly....

If the marker has a colors array then there is no need to set `enableBackgroundColor` - we just always apply the first color in the `colors` array as the text color and the second as the background color.  Then we could change this prop to be called `autoColorBackground` and if it's set to true _and_ the marker has has a single `color` then we set the background color.  That way one could use a mix of markers with 2 colors (text & background) and others w/ only a single color and the single color markers would still have a transparent background as they do now.

Thoughts?

vidaaudrey: I love the idea! only I'd call it `autoBackgroundColor`:) Will update!

docs/src/4.10.Text.mdx
```diff
@@ -6,13 +6,39 @@ import {Text} from './jsx/allLiveEditors'
...
+| --------------------- | -------- | --------- | ------- |
+| `children`            |          | `Text[]`  | `[]`    | array of Text markers to render                                   |
+| `autoBackgroundColor` | yes      | `boolean` |         | automatically add solid background color with contrast if enabled |
+
+### Text and background colors
+
+There are three ways to control the text and background colors:
+
+- Text without background color: set `color` field
+- Text with automatically background color: set `color` field and set `autoBackgroundColor` to `true`
```
jtbandes:
```suggestion
- Text with automatically-chosen background color: set `color` field and set `autoBackgroundColor` to `true`
```

docs/src/4.10.Text.mdx
```diff
@@ -23,15 +49,18 @@ type Text = {
...
   },
-  color: {
-    r: number, // between 0 and 1
-    g: number, // between 0 and 1
-    b: number, // between 0 and 1
-    a: number, // between 0 and 1
-  },
+  color: Color,
+  colors?: [
+  // text color, only applied when colors.length >= 2
```
jtbandes: nit: indentation

vidaaudrey:
> This looks pretty good. I like the `colors[0]/[1]` support for background color. I'm not totally convinced that it's necessary to provide this auto background color feature, since the user could just choose a background color for themselves, and it's a pretty long/awkward name for a prop. Was also thinking of `<Text backgroundColor="auto"` but I don't really like it better.

Currently we are batching the text markers so we can render one single `<Text>` in worldview. Allowing the user to specify `backgroundColor` may cause some text labels to be illegible. The other advantage of `autoBackgroundColor` is to allow us to globally control the worldview label UI with default dark/light background, so it doesn't go out of control over time.

docs/src/4.10.Text.mdx
```diff
@@ -23,15 +49,18 @@ type Text = {
     y: number,
     z: number
   },
-  color: {
-    r: number, // between 0 and 1
-    g: number, // between 0 and 1
-    b: number, // between 0 and 1
-    a: number, // between 0 and 1
-  },
+  color: Color,
```
esthersweon: So this `color` field is always required, even when you just want to pass in a 2 element `colors` array?

packages/regl-worldview/src/commands/Text.js
```diff
@@ -50,11 +57,24 @@ function insertGlobalCss() {
...
 
+function isColorDark({ r, g, b }: Color): boolean {
+  // ITU-R BT.709 https://en.wikipedia.org/wiki/Rec._709
+  // 0.2126 * 255 * r + 0.7152 * 255 * g + 0.0722 * 255 * b
+  const luma = 54.213 * r + 182.376 * g + 18.411 * b;
+  return luma < BRIGHTNESS_THRESHOLD;
+}
+
+function isColorEqual(a: Color, b: Color): boolean {
+  return a.r === b.r && a.g === b.g && a.b === b.b && a.a === b.a;
```
esthersweon: use lodash's `isEqual` instead? Does deep comparison of objects.

docs/src/4.10.Text.mdx
esthersweon: My instinct is to just get rid of this first `color` prop, and just conditionally check if the `colors` array has 1 or 2 elements?

brianc: good instinct on getting rid of `color`, but we should support both single color as the `color` prop as well as multiple.  ROS markers support both [color and colors](http://docs.ros.org/api/visualization_msgs/html/msg/Marker.html).  Also for backwards compatibility reasons we should maintain support for `color` - though you're right it should be marked as an optional, not required, property in flow.

packages/regl-worldview/src/commands/Text.js
vidaaudrey: Since we are need the `Text` component to be as performant as possible (that's the reason we created `TextElement ` instead of using React's DOM update), it would be better to not introduce `isEqual` to do deep object comparison as we could just simply compare the 4 color fields

docs/src/4.10.Text.mdx
vidaaudrey: Good feedbacks. Marked the color as optional

PR #80, opened 2019-02-14
Fix Overlay disappearing in CodeSandbox

vidaaudrey:
Current: https://codesandbox.io/s/4z889qrzw7 
Fixed: https://codesandbox.io/s/p3p5v4567x


vidaaudrey:
> Ok! Why does this fix it though?

Explained offline and went for a different approach. The original issue is that the overlay is not absolutely positioned to `left: 0; top: 0`. Adding `display: flex` forces it to position it correctly but it's not very optimal since we can't expect the consumer to have the container set up correctly.  Adding `left: 0; top: 0` is nicer approach. Thanks!

PR #82, opened 2019-02-15
Use cruise/webviz-ci:0.0.3 in CI

janpaul123:
Which uses Chrome 72 instead of 71.


janpaul123:
It's the main Dockerfile in this repo. I don't think we were ever loading maps in screenshot tests (external dependencies should typically be disabled).

PR #83, opened 2019-02-17
allow for deeper than 1 level topic namespace

patrickpei:
By default, [ZED's ROS wrapper](https://github.com/stereolabs/zed-ros-wrapper) uses hierarchical namespaces for the camera topics (e.g. `zed/rgb/image_rect_color`) which currently doesn't work because the rectified topics assume the topics to be available on `<group>/image_rect_color` or `<group>/image_rect_color_compressed `.

### DIFF
--- a/packages/webviz-core/src/panels/ImageView/index.js
+++ b/packages/webviz-core/src/panels/ImageView/index.js
@@ -95,8 +95,8 @@ class ImageView extends Component<Props> {
       topics.sort(naturalSort("name"));
 
       const rectifiedTopics = topics
-        .filter(topic => RECTIFIED_TOPIC_REGEX.test(topic.name))
-        .map(topic => topic.name);
+        .filter((topic) => RECTIFIED_TOPIC_REGEX.test(topic.name))
+        .map((topic) => topic.name);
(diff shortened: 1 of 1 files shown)


jtbandes:
Thanks for your contribution! 😄 

We've recently made some internal changes to the ImageView panel that we need to sync with this public repo, which we should be able to do sometime this week. (We changed how this rectified/unrectified detection works specifically for [ImageMarker](http://docs.ros.org/api/visualization_msgs/html/msg/ImageMarker.html) support.) Just a heads up that those changes might conflict with this PR, and we'll also want to add some tests for this change. We can work with you a bit more on it once our latest changes are out 🔜.

janpaul123:
This should be fixed now. Thanks for your contribution! Let us know if you have any more issues. 😄

janpaul123:
@patrickpei I'd love to learn more about what you use Webviz for. We're looking to add a bunch more features in the near term, and would love to work with people outside of the company to help test them (and maybe even get featured in a blog post or so once we're ready for that!). Would you have time to hop on a quick call next week? Can you shoot me an email at jp.posma@getcruise.com? 🙂

PR #84, opened 2019-02-17
Bump react to ^16.8.2

vidaaudrey:


docs/src/jsx/utils/WorldviewCodeEditor.js
```diff
@@ -43,9 +43,8 @@ import Worldview, {
 // Add required packages and files for all examples to run
 const CODE_SANDBOX_CONFIG = {
   dependencies: {
-    // TODO(Audrey): update to use "latest" once hooks api is finalized
-    react: "16.7.0-alpha.2",
-    "react-dom": "16.7.0-alpha.2",
+    react: "latest",
+    "react-dom": "latest",
```
brianc: I feel like using `latest` here for all these dependencies might be a bit dangerous as opposed to pinning to some semver range...since it's just the code editor it's not a huge deal, but we might wanna pin these to allow floating patch version only or something in the future.

vidaaudrey: Good point. I've thought about that as well. The main reason for me to use the latest dependencies is so that we could identify and fix problems early. It's a trade-off I guess.

PR #85, opened 2019-02-17
Add mouse event support for instanced objects

vidaaudrey:
## Problem
Currently, mouse event for instanced objects are not supported: if we have a cube/triangle/sphere/point marker that contains thousands of objects, there is no way to interact with those objects.

## Solution
- For automatic instanced hitmap id management, simply add a supported mouse handler to the command. 

- For custom management use both:
  - `getHitmapProps` which generates the hitmap props based on children (markers)
  - `getObjectFromHitmapId` to map `objectId` returned by hitmap event to the specific `hitmapProp`/ marker object

## Test
Add a new example `MouseEventInstanced.js`, all can return the correct `objectId` when clicked.
The` Interactivity` and `MouseEvent` examples should continue to work. 

bag


packages/regl-worldview/src/utils/commandUtils.js
```diff
@@ -122,11 +122,23 @@ export function withPose(command: ReglCommand): ReglCommand {
...
 
-export function getVertexColors(props: any) {
-  if (!props.colors || !props.colors.length) {
-    return constantRGBAArray(props.points.length, props.color);
+export function getVertexColors({
+  colors,
+  color,
+  points,
+}: {
+  colors?: (Color | Vec4)[],
```
jtbandes: we should start using [`$ReadOnlyArray`](https://flow.org/en/docs/types/arrays/#toc-readonlyarray) for these because it will make typechecking easier for users.

packages/regl-worldview/src/commands/Command.js
```diff
@@ -133,13 +137,20 @@ function getHitmapProps(getHitmapId, children) {
...
 
   return children.reduce((memo, marker) => {
-    const hitmapId = getHitmapId(marker);
-    // filter out components that don't have hitmapIds
-    if (hitmapId != null) {
+    if (marker.colors) {
       memo.push({
         ...marker,
-        color: intToRGB(getHitmapId(marker) || 0),
+        colors: marker.colors.map((_, colorIndex) => intToRGB(getHitmapId(marker, colorIndex) || 0)),
```
jtbandes: Unless I'm reading this wrong, it seems like the hitmap ids will be different depending on whether the input marker had `color` or `colors`. That doesn't really make sense to me; the hitmap id is a very separate feature from the visible color.

docs/src/3.4.MouseEvents.mdx
```diff
@@ -48,3 +48,8 @@ Sample usage:
 ## Full Example
 
 <MouseEvents />
+
+## HitmapId Mapping for Instanced Objects
+Usually using instanced objects can help improve rendering performance, e.g. an instanced sphere can contain thousands of spheres. When the user clicks a single sphere, the mapped `objectId` which was generated through `getHitmapId` function will be returned from the event. A common way to generate the hitmapId is:  `getHitmapId={(marker, colorIndex) => marker.id + colorIndex}`.
```
jtbandes: might want to call this `instanceId` to match [GL terminology](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/gl_InstanceID.xhtml).

packages/regl-worldview/src/commands/Command.js
vidaaudrey: Good catch. Updated to use `points` field instead

docs/src/jsx/MouseEventsInstanced.js
```diff
@@ -0,0 +1,87 @@
...
+    points,
+  };
+
+  function onObjectClick(ev, { objectId, object }) {
+    let msg = "";
+    // use the objectId to find the particular object that's been clicked
+    if (object && object.points && objectId) {
+      const idx = objectId - object.id;
+      if (idx >= 0 && idx <= points.length) {
+        msg = `Clicked objectId is ${objectId} and position is ${JSON.stringify(markerMatrix.points[idx])}`;
```
robin-pham: nit: `Clicked object's id is ${objectId} and its position is...`

docs/src/jsx/MouseEventsInstanced.js
```diff
@@ -0,0 +1,87 @@
...
+  return (
+    <Worldview onClick={onWorldviewClick}>
+      <div
+        style={{
+          position: "absolute",
+          left: 0,
+          top: 0,
+          width: 320,
+          maxWidth: "100%",
+          background: "rgba(255, 255, 255, 0.66)",
```
robin-pham: can we keep the background color + text color consistent on this page?

i.e. use the same colors from the example above it

```
{
            color: "white",
            backgroundColor: "rgba(165, 94, 255, 0.1)", // can probably assign this to a constant
}


```
and then the span's style should also be `color: "white"`

packages/regl-worldview/src/commands/Command.js
```diff
@@ -84,9 +84,12 @@ export default class Command<T> extends React.Component<Props<T>> {
 
   _getHitmapPropFromHitmapId(objectId: number) {
     const { hitmapProps = [] } = this.props;
+
     return hitmapProps.find((hitmapProp) => {
-      // TODO handle objects with 'colors' property
-      if (hitmapProp.color) {
+      if (hitmapProp.points) {
+        // just return the original hitmapProp for instanced rendering
+        return true;
```
robin-pham: if there are multiple components with instanced rendering, doesn't this just always fire off or am I misunderstanding something? It seems like we'll have to calculate the range/set of ids that each `hitmapProp` can produce, and if the `objectId` is within that range/set, we can return `true`.

docs/src/jsx/MouseEventsInstanced.js
vidaaudrey: Will use `rgba(0, 0, 0, 0.5)` which was used as background color by the Full example at the same bg. Probably not worth to add as a constant, it'll introduce more files/variables to the global editor scope.

packages/regl-worldview/src/commands/Command.js
vidaaudrey: Good point! totally missed it.

packages/regl-worldview/src/commands/Command.js
```diff
@@ -84,9 +84,14 @@ export default class Command<T> extends React.Component<Props<T>> {
 
   _getHitmapPropFromHitmapId(objectId: number) {
     const { hitmapProps = [] } = this.props;
+
     return hitmapProps.find((hitmapProp) => {
-      // TODO handle objects with 'colors' property
-      if (hitmapProp.color) {
+      if (hitmapProp.points) {
+        // for instanced rendering, find the hitmapProp that produces the same id range and return it
+        if (objectId >= hitmapProp.id && objectId < hitmapProp.id + hitmapProp.points.length) {
```
robin-pham: hmm this relies on the implementation of `getHitmapId` to be `(marker, idx) => marker.id + idx`. I wonder if it makes more sense to default that implementation if this is how we go about the bool check?

vidaaudrey: Good question! Addressed this by introducing `testObjectIdInRange` prop :)

packages/regl-worldview/src/commands/Command.js
```diff
@@ -127,35 +138,71 @@ export default class Command<T> extends React.Component<Props<T>> {
...
+function defaultGetHitmapProps<T>(getHitmapId: GetHitmapId<T>, children: T[]): ?(HitmapProp[]) {
   if (!getHitmapId || !children || children.length === 0) {
     return undefined;
   }
 
   return children.reduce((memo, marker) => {
-    const hitmapId = getHitmapId(marker);
-    // filter out components that don't have hitmapIds
-    if (hitmapId != null) {
+    if (marker.points) {
```
brianc: If I'm reading this correctly, this changes the logic in a non-backwards-compatible way: previous if a marker had points (such as a polygon, line strip, or triangle strip) it would receive a single hitmap id...but now we assign each point in those markers it's own hitmap id?  ⚠️

packages/regl-worldview/src/commands/Command.js
```diff
@@ -83,10 +88,15 @@ export default class Command<T> extends React.Component<Props<T>> {
   }
 
   _getHitmapPropFromHitmapId(objectId: number) {
-    const { hitmapProps = [] } = this.props;
+    const { hitmapProps = [], testObjectIdInRange } = this.props;
+
     return hitmapProps.find((hitmapProp) => {
-      // TODO handle objects with 'colors' property
-      if (hitmapProp.color) {
+      if (hitmapProp.points) {
```
brianc: I don't think we can rely on an object having `.points` as it being an instanced object - some things may have `.points` and just be a single marker.  We might need to rethink how we handle custom hitmap props a bit for custom `<Command />`s.

packages/regl-worldview/src/commands/Command.js
```diff
@@ -127,35 +136,102 @@ export default class Command<T> extends React.Component<Props<T>> {
...
+      );
+      hitmapProps = defaultGetHitmapProps(getHitmapId, children);
+      getObjectFromHitmapId = defaultGetObjectFromHitmapId;
+    } else if (getHitmapProps && getObjectFromHitmapId) {
+      if (enableHitmap) {
+        hitmapProps = getHitmapProps(children, false);
+      } else if (enableInstanceHitmap) {
+        hitmapProps = getHitmapProps(children, true);
+      }
+    }
```
robin-pham: this method feels realllly complicated. is it because we want to preserve compatibility with pre-existing APIs?

I wonder if this can be handled elsewhere, since it's really difficult to read (it also seems brittle and would require testing).

packages/regl-worldview/src/commands/Command.js
```diff
@@ -127,35 +136,102 @@ export default class Command<T> extends React.Component<Props<T>> {
...
 
   return children.reduce((memo, marker) => {
-    const hitmapId = getHitmapId(marker);
-    // filter out components that don't have hitmapIds
-    if (hitmapId != null) {
-      memo.push({
-        ...marker,
-        color: intToRGB(getHitmapId(marker) || 0),
-      });
+    if (marker.color) {
```
brianc: I don't think the marker having a `.color` property should have bearing on whether or not it's passed to the provided `getHitmapId` function - markers without colors could be a thing in custom `<Command />` components which have colors hard coded in the shaders or use textures or other ways to provide colors to webgl.  I suggest this if check be removed.

packages/regl-worldview/src/commands/Command.js
```diff
@@ -9,18 +9,39 @@
...
-import type { ComponentMouseHandler, MouseEventEnum, RawCommand } from "../types";
+import type { ComponentMouseHandler, MouseEventEnum, RawCommand, Vec4, Color } from "../types";
 import { getIdFromColor, intToRGB } from "../utils/commandUtils";
 import { getNodeEnv } from "../utils/common";
 import { type WorldviewContextType } from "../WorldviewContext";
 import WorldviewReactContext from "../WorldviewReactContext";
 
+export type HitmapProp<T> = T & ({ colors: Vec4[] } | { color: Vec4 });
+export type GetObjectFromHitmapId<T> = (objectId: number, enableInstanceHitmap: boolean) => ?T;
+export type GetHitmapProps<T> = (children: T[], enaleInstanceHitmap: boolean) => ?HitmapProp<T>;
```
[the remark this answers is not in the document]
vidaaudrey: Removing both `enableHitmap` and `enableInstanceHitmap` 😄

packages/regl-worldview/src/commands/Command.js
```diff
@@ -82,33 +103,21 @@ export default class Command<T> extends React.Component<Props<T>> {
...
-        const hitmapPropId = getIdFromColor(hitmapProp.color.map((color) => color * 255));
-        if (hitmapPropId === objectId) {
-          return true;
-        }
-      }
-      return false;
-    });
-  }
-
   handleMouseEvent(objectId: number, e: any, ray: any, mouseEventName: MouseEventEnum) {
```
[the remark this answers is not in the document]
vidaaudrey: Updated as well since it's just one line change

packages/regl-worldview/src/commands/Command.js
```diff
@@ -82,33 +103,21 @@ export default class Command<T> extends React.Component<Props<T>> {
...
     const mouseHandler = this.props[mouseEventName];
-    if (!mouseHandler) {
+    const { hitmapProps = [], getObjectFromHitmapId, enableInstanceHitmap } = this.props;
+    if (!mouseHandler || hitmapProps.length === 0 || !getObjectFromHitmapId) {
       return;
     }
 
-    const hitmapProp = this._getHitmapPropFromHitmapId(objectId);
+    const hitmapProp = getObjectFromHitmapId(objectId, hitmapProps, enableInstanceHitmap);
```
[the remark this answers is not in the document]
vidaaudrey: Updated, incorporated your suggestion

vidaaudrey:
> my only concern is refactoring the code that preserves pre-existing APIs in the future. When/how will we decide to deprecate that code?

I'm not sure at this moment, probably not any time soon. It also depends on: 
- do we have features that require major release, e.g. significant perf improvement, breaking api changes
- can we provide a migration path or do we have time to help other teams migrate 

In general, I like Brian's suggestion, which is to follow React and always try to make it backward-compatible if possible. The consumer can see the console warnings and update at their own pace.

PR #86, opened 2019-02-18
[BugFix] Add id to filled polygon

troygibb:
### Bug

The `FilledPolygons` command didn't sustain the `id` (which is required by the `Polygon` type) into the `Triangles` command.

Side note: Do we have a way of testing these commands?


packages/regl-worldview/src/commands/FilledPolygons.js
```diff
@@ -50,18 +50,17 @@ type Props = {
...
   const triangles = [];
   for (const poly of polygons) {
-    const { points } = poly;
+    const { points, id, color } = poly;
     const pose = poly.pose ? poly.pose : NO_POSE;
     const earcutPoints: Vec3[] = getEarcutPoints(points);
     const polyPoints: Point[] = earcutPoints.map(([x, y, z]) => ({ x, y, z }));
-    const color = poly.color;
     triangles.push({
+      id,
```
robin-pham: oops good catch. we should just spread the rest of the props from the `poly` into this object.

```
const { points, color, ...restProps } = poly;
.
.
.
triangles.push({
  ...restProps,
  points: polyPoints,
})

```

packages/regl-worldview/src/commands/FilledPolygons.js
```diff
@@ -50,18 +50,17 @@ type Props = {
...
       points: polyPoints,
       pose,
       color: { r: color[0], g: color[1], b: color[2], a: color[3] },
       scale: DEFAULT_SCALE,
     });
   }
-
   return <Triangles getHitmapId={getHitmapId}>{triangles}</Triangles>;
```
robin-pham: here we should spread the rest of the react props in order to pass through the mouse handlers.

packages/regl-worldview/src/commands/FilledPolygons.js
```diff
@@ -47,21 +47,24 @@ type Props = {
 };
 
 // command to draw a filled polygon
-export default function FilledPolygons({ children: polygons = [], getHitmapId }: Props) {
+export default function FilledPolygons({ children: polygons = [], getHitmapId, ...props }: Props) {
   const triangles = [];
   for (const poly of polygons) {
-    const { points } = poly;
+    const { points, color, ...restOfProps } = poly;
```
vidaaudrey: nit: `restOfMarker`?

troygibb: done!

PR #88, opened 2019-02-19
Bind raycast

robin-pham:
encountered a bug where calling `raycast` from [react-dnd](https://github.com/react-dnd/react-dnd) had `this` as undefined.

wonder if it's a good idea to bind this to all of these class functions or if we're comfortable with just doing it for `raycast` for now.


vidaaudrey:
> Seems reasonable to just do it for everything.

this only happened with one function in one app, maybe we just change one for now and if we see more issues we'll update all later?

PR #91, opened 2019-02-22
Amend FilledPolygon color type

troygibb:
I should have added this update to my [previous PR](https://github.com/cruise-automation/webviz/pull/86) after realizing that `FilledPolygons` should probably use the shared `Color` type that the reset of the markers use. 

@robin-pham we don't have any dependencies on `FilledPolygons` elsewhere do we?

### DIFF
--- a/packages/regl-worldview/src/types/index.js
+++ b/packages/regl-worldview/src/types/index.js
@@ -188,6 +188,7 @@ export type TriangleList = BaseShape & {
 
 export type PolygonType = BaseShape & {
   points: Point[],
+  color: Color,
   id: number,
 };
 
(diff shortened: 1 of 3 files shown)


packages/regl-worldview/src/types/index.js
```diff
@@ -188,6 +188,7 @@ export type TriangleList = BaseShape & {
 
 export type PolygonType = BaseShape & {
   points: Point[],
+  color: Color,
```
vidaaudrey: Just a note that we'll need to add `colors?` as well in the future, to enable individual vertex coloring

robin-pham:
forking branch + PR -> https://github.com/cruise-automation/webviz/pull/93

PR #92, opened 2019-02-25
Fix Overlay so it doesn't flow outside of Worldview by default 

vidaaudrey:
## The Problem
The Overlay sometimes go over the Worldview main container which is not the expected behavior: [Codesandbox](https://codesandbox.io/s/nwzr8o3wll?fontsize=14)
image

The main cause is that we are using `React.Fragment` as the root container for Worldview, and it doesn't have any styling. When the Overlay gets rendered, it'll be at the same level as Worldview. 

## The Fix 
Use `div` instead of `React.Fragment` for Worldview root container and add default `relative` positioning so the Overlay will be positioned correctly relative it's parent. I also add `style` prop to allow consumer to add more custom styling to the root container. 

screen shot 2019-02-24 at 4 17 04 pm

## Test 
Manually test the Hitmap and Overlay examples to make sure the content doesn't overflow the container.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,5 +5,8 @@
   "javascript.validate.enable": false,
   "files.associations": {
     "*.mdx": "markdown"
+  },
+  "editor.codeActionsOnSave": {
+    "source.organizeImports": false
   }
 }
--- a/packages/regl-worldview/src/Worldview.js
+++ b/packages/regl-worldview/src/Worldview.js
@@ -27,6 +27,7 @@ export type BaseProps = {|
   hitmapOnMouseMove?: boolean,
   showDebug?: boolean,
   children?: React.Node,
+  style: { [styleAttribute: string]: number | string },
 
   cameraState?: CameraState,
   onCameraStateChange?: (CameraState) => void,
@@ -71,6 +72,7 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
 
   static defaultProps = {
     backgroundColor: DEFAULT_BACKGROUND_COLOR,
+    style: {},
   };
 
   constructor(props: BaseProps) {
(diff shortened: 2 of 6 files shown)


brianc:
If someone wanted to prevent the overlay from going outside the box they could wrap worldview in a `<div position='relative'>`. By not adding that ourselves to the component we allow the consumer to add it & remove the overflow, or in some cases not add it and keep the overflow...which could be desired behavior.  If we don't "fix" this then it can work either way.

vidaaudrey:
> If someone wanted to prevent the overlay from going outside the box they could wrap worldview in a `<div position='relative'>`. By not adding that ourselves to the component we allow the consumer to add it & remove the overflow, or in some cases not add it and keep the overflow...which could be desired behavior. If we don't "fix" this then it can work either way.

My consideration is that by fixing the default behavior, there will be less work at the consumer side to figure out why the overlay doesn't position correctly. Currently the interactivity demo (https://codesandbox.io/s/kxk42rnw05) is broken in codesandbox due to the same lacking container div issue. And by adding the `style` prop, there will be less stying bug, and the user can still set the clipping behavior if desired by using `style={{overflow: 'visible'}}`. I'll update the overlay demo as below to show the difference. What do you think? 

bag

packages/regl-worldview/src/Worldview.js
```diff
@@ -249,15 +251,14 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
...
     const { worldviewContext } = this.state;
-    const style = { width, height };
 
     return (
-      <React.Fragment>
+      <div style={{ position: "relative", overflow: "hidden", ...style }}>
         <CameraListener cameraStore={worldviewContext.cameraStore} keyMap={keyMap}>
           <canvas
-            style={style}
+            style={{ width, height, maxWidth: "100%", maxHeight: "100%" }}
```
robin-pham: does the `maxWidth: "100%", maxHeight: "100%"` do anything?

vidaaudrey: While playing locally, I noticed the canvas width and height can go over the container depending on the parent's layout. Adding this will ensure that the canvas doesn't overflow.

.vscode/settings.json
```diff
@@ -5,5 +5,8 @@
   "javascript.validate.enable": false,
   "files.associations": {
     "*.mdx": "markdown"
+  },
+  "editor.codeActionsOnSave": {
+    "source.organizeImports": false
```
vidaaudrey: I turned this on as editor default for some typescript projects. It automatically sort the imports nicely but it also add two lines at the bottom of import lines which has conflict with our own eslint settings  (only 1 line allowed ).  Thought should just be explicit about disabling default sorting behavior here.

PR #95, opened 2019-02-27
 1️⃣📈 Prevent an off by 1 error for lines without points

3on:
I made the story a bit more fun.

## With the bug 🇦🇹
image

## Without the bug 🏳
image

### DIFF
--- a/packages/regl-worldview/src/commands/Lines.js
+++ b/packages/regl-worldview/src/commands/Lines.js
@@ -389,7 +389,8 @@ const lines = (regl: any) => {
 
     const joined = primitive === "line strip";
     const effectiveNumPoints = numInputPoints + (shouldClose ? 1 : 0);
-    const instances = joined ? effectiveNumPoints - 1 : Math.floor(effectiveNumPoints / 2);
+    const guardedNumber = effectiveNumPoints > 0 ? effectiveNumPoints - 1 : 0;
+    const instances = joined ? guardedNumber : Math.floor(effectiveNumPoints / 2);
 
     render(debug, () => {
       command({
--- a/packages/regl-worldview/src/commands/Lines.js
+++ b/packages/regl-worldview/src/commands/Lines.js
@@ -359,6 +359,10 @@ const lines = (regl: any) => {
     const { debug, primitive = "lines", scaleInvariant = false } = props;
     const numInputPoints = props.points.length;
 
+    if (numInputPoints < 2) {
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


janpaul123:
(We will soon add screenshot testing for this repo so this story is great to prevent future regressions of this bug! 🎉)

packages/regl-worldview/src/commands/Lines.js
```diff
@@ -389,7 +389,8 @@ const lines = (regl: any) => {
 
     const joined = primitive === "line strip";
     const effectiveNumPoints = numInputPoints + (shouldClose ? 1 : 0);
-    const instances = joined ? effectiveNumPoints - 1 : Math.floor(effectiveNumPoints / 2);
+    const guardedNumber = effectiveNumPoints > 0 ? effectiveNumPoints - 1 : 0;
```
vidaaudrey: maybe call it `effectiveJoinedInstances` to be more explicit?

packages/regl-worldview/src/stories/lines.stories.js
```diff
@@ -0,0 +1,57 @@
...
+          position: { x: 0, y: 0, z: 0 },
+          orientation: { x: 0, y: 0, z: 0, w: 1 },
+        },
+        points: [],
+        color: [1, 1, 1, 0.75],
+        scale: { x: 2.2, y: 0, z: 0 },
+      },
+    ];
+    return (
+      <Worldview defaultCameraState={DEFAULT_CAMERA}>
```
vidaaudrey: nit: this story can work without `defaultCameraState={DEFAULT_CAMERA}`

packages/regl-worldview/src/stories/lines.stories.js
```diff
@@ -0,0 +1,57 @@
...
+  targetOffset: [5, 5, 0],
+  thetaOffset: Math.PI / 2,
+  phi: Math.PI / 4,
+  distance: 20,
+  targetOrientation: [0, 0, 0, 1],
+};
+
+storiesOf("Worldview", module)
+  .addDecorator(withKnobs)
+  .add("<Lines> Empty points bug", () => {
```
vidaaudrey: nit: `<Lines> does not render empty points `

packages/regl-worldview/src/commands/Lines.js
jtbandes: Good catch! Can you simply use (after line 360)
```js
if (numInputPoints > 0) {
  return;
}
```
?

3on: Yeap it makes sense to early return.

PR #98, opened 2019-03-11
 Add default hitmap props for FilledPolygons & update docs

vidaaudrey:
- add default hitmap props for FilledPolygons so correct `objectId` is returned without calling `getHitmapId`
- add `FilledPolygonsHitmap.js` example 
- update docs: explicitly indicate that the input marker ids need to be positive integers which is required for correct hitmap id mapping 

Test: 
New api:
id

Compatibility with the old api getHitmapId (notice the polygon information shows up correctly upon click):
old api

### DIFF
--- a/packages/regl-worldview/src/commands/FilledPolygons.js
+++ b/packages/regl-worldview/src/commands/FilledPolygons.js
@@ -46,6 +46,7 @@ function getEarcutPoints(points: Vec3[]): Vec3[] {
 
 type Props = {
   children: PolygonType[],
+  getHitmapId?: (PolygonType) => number,
   getHitmapProps: GetHitmapProps<PolygonType>,
   getObjectFromHitmapId: GetObjectFromHitmapId<PolygonType>,
 };
@@ -53,9 +54,10 @@ type Props = {
 // command to draw a filled polygon
 export default function FilledPolygons({
   children: polygons = [],
+  getHitmapId,
   getHitmapProps = getHitmapPropsForFilledPolygons,
   getObjectFromHitmapId = getObjectFromHitmapIdForFilledPolygons,
-  ...props
+  ...rest
 }: Props) {
   const triangles = [];
   for (const poly of polygons) {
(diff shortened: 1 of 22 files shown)


packages/regl-worldview/src/commands/FilledPolygons.js
```diff
@@ -44,11 +46,17 @@ function getEarcutPoints(points: Vec3[]): Vec3[] {
 
 type Props = {
   children: PolygonType[],
-  getHitmapId?: (TriangleList) => number,
```
brianc: I think this is a backwards incompatible change - is it a lot of work to still support this prop & pass it through to the `<Triangles />` command?

vidaaudrey: Not a lot of work. Let me add it

packages/regl-worldview/src/commands/FilledPolygons.js
```diff
@@ -46,16 +46,18 @@ function getEarcutPoints(points: Vec3[]): Vec3[] {
...
   children: PolygonType[],
+  getHitmapId?: (PolygonType) => number,
   getHitmapProps: GetHitmapProps<PolygonType>,
   getObjectFromHitmapId: GetObjectFromHitmapId<PolygonType>,
 };
 
 // command to draw a filled polygon
 export default function FilledPolygons({
   children: polygons = [],
+  getHitmapId,
```
robin-pham: I don't think this is necessary, `getHitmapId` would have already been passed through

vidaaudrey: I added because I wanted to explicitly support the old api. Guess it's better to implicitly support it so we don't have to remove it in the future. @brianc  what do you think?

brianc: yeah implicit probably the way to go - mostly just leaving it as a valid prop type was my concern so we don't break flow for folks who were using it.

vidaaudrey: Makes sense to make flow work. I misunderstood you. Will add it back.

PR #99, opened 2019-03-11
Add documentation for pose

vidaaudrey:
## Problem 
There was some confusion around how to handle `pose`. It's unclear that the translation happens before the rotation. 

## Fix
- Add a Glossary section to document different terms
- Add documentation for pose

### DIFF
--- a/docs/src/3.7.Glossary.mdx
+++ b/docs/src/3.7.Glossary.mdx
@@ -2,7 +2,7 @@
 
 ## Pose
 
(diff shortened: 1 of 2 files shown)


docs/src/3.7.Glossary.mdx
```diff
@@ -0,0 +1,12 @@
+# Glossary
+
+## Pose
+
+In computer vision and robotics, the **POSE** determines the _position_ and _orientation_ of the specific object in the 3D space. It describes how to bring the target object from a reference pose to the observed pose through translation and rotation. e.g. the pose of A with position `{ x: 1, y: 1, z: 1 }` and orientation `{ x: 0, y: 0, z: 0, w: 1 }` is given by the translation from origin to A's position, and the rotation of A's orientation. The data format follows ROS's [Pose Message](http://docs.ros.org/lunar/api/geometry_msgs/html/msg/Pose.html) definition where position is a point with xyz coordinates and orientation is a [quaternion](https://en.wikipedia.org/wiki/Quaternion).
```
janpaul123: Maybe link to https://en.wikipedia.org/wiki/Pose_(computer_vision) as well? Also maybe not write POSE in all caps, the bold style is probably enough (nice to follow Wikipedia convention here).

vidaaudrey: Good idea

vidaaudrey:
> Should we link to this page from places where we use `pose` to make it easier to find?

We don't mention it anywhere yet except in the upcoming tutorial and the flow type definitions.  Will add later for the specific docs.

PR #100, opened 2019-03-11
Use depth by default for triangles

robin-pham:
Encountered bugs where triangles did not respect depth. I think this `depth` property by default should use the default values that regl uses, and instead can be overridden by the consumer.

### DIFF
--- a/packages/regl-worldview/src/utils/commandUtils.js
+++ b/packages/regl-worldview/src/utils/commandUtils.js
@@ -98,6 +98,14 @@ export const blend = {
   },
 };
 
+// regl default depth values that can be overridden by props
+export const depth = {
+  enable: (context: Object, props: Object) => (props.depthEnable === false ? false : true),
+  mask: (context: Object, props: Object) => (props.depthMask === false ? false : true),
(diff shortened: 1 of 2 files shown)


packages/regl-worldview/src/utils/commandUtils.js
```diff
@@ -98,6 +98,14 @@ export const blend = {
   },
 };
 
+// regl default depth values that can be overridden by props
+export const depth = {
```
jtbandes: I think there might've been discussion about this somewhere already but I can't find it now — did you see about whether you can make the whole `depth` object a function? Would be nicer for `props.depth` to mirror the type supported by regl rather than splitting it into `depthEnable`, `depthMask` etc.

robin-pham: yup I wanted to do the whole 'depth object function' initially but it didn't seem to work, hence why I made this janky `depthEnable` api :(

vidaaudrey:
@robin-pham  going to close this PR now as it's implemented in #128

PR #101, opened 2019-03-13
Don't trigger onClick if mouse moved beyond distance

vidaaudrey:
## Problem 
The Worldview `onClick` handler is being fired even when user is panning around because currently `onClicked` is always fired after `onMouseUp`. 

## Fix 
Track the movement, only fire `onClick` if the mouse has moved a distance within 3 screen pixels. 

## Test
Manually test: press and pan around for long distance, the information display should not change. Will add a drag object demo later in a separate PR. 

demo

### DIFF
--- a/docs/src/3.1.Worldview.mdx
+++ b/docs/src/3.1.Worldview.mdx
@@ -2,23 +2,23 @@
 
 ## Props
 
-| Name                  | Type                                            | Default                | Description                                                                                                            |
-| --------------------- | ----------------------------------------------- | ---------------------- | ---------------------------------------------------------------------------------------------------------------------- |
-| `cameraState`         | `$Shape<CameraState>`                           |                        | control the [cameraState](#/docs/api/camera), must use with `onCameraStateChange`                                      |
-| `onCameraStateChange` | `(CameraState) => void`                         |                        | callback whenever the camera state changes                                                                             |
-| `defaultCameraState`  | `$Shape<CameraState>`                           | `DEFAULT_CAMERA_STATE` | pass in this prop to turn Worldview to an uncontrolled component. `onCameraStateChange` will be ignored                |
-| `keyMap`              | `CameraKeyMap`                                  | {}                     | override default Worldview hotkeys                                                                                     |
-| `backgroundColor`     | `Vec4`                                          | `[0,0,0,1]`            | change the scene background color                                                                                      |
-| `hitmapOnMouseMove`   | `boolean`                                       |                        | if enabled, the clicked objectId will be returned from the mouse event handler                                         |
-| `children`            | `React.Node`                                    |                        | regl components to render inside the Worldview                                                                         |
-| `style`               | `{ [styleAttribute: string]: number | string }` | {}                     | html style object for the Worldview root element                                                                       |
-| `showDebug`           | `boolean`                                       |                        | show debug information                                                                                                 |
-| `mouseClickRadius`    | `number`                                        | 3                      | if the mouse has moved distance beyond the mouseClickRadius limit, the click event handler isn't going to be triggered |
-| `onDoubleClick`       | `MouseHandler`                                  |                        |                                                                                                                        |
-| `onMouseDown`         | `MouseHandler`                                  |                        |                                                                                                                        |
-| `onMouseUp`           | `MouseHandler`                                  |                        |                                                                                                                        |
-| `onMouseMove`         | `MouseHandler`                                  |                        |                                                                                                                        |
-| `onClick`             | `MouseHandler`                                  |                        |                                                                                                                        |
+| Name                  | Type                                            | Default                | Description                                                                                                                                           |
+| --------------------- | ----------------------------------------------- | ---------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `cameraState`         | `$Shape<CameraState>`                           |                        | control the [cameraState](#/docs/api/camera), must use with `onCameraStateChange`                                                                     |
+| `onCameraStateChange` | `(CameraState) => void`                         |                        | callback whenever the camera state changes                                                                                                            |
+| `defaultCameraState`  | `$Shape<CameraState>`                           | `DEFAULT_CAMERA_STATE` | pass in this prop to turn Worldview to an uncontrolled component. `onCameraStateChange` will be ignored                                               |
+| `keyMap`              | `CameraKeyMap`                                  | {}                     | override default Worldview hotkeys                                                                                                                    |
+| `backgroundColor`     | `Vec4`                                          | `[0,0,0,1]`            | change the scene background color                                                                                                                     |
+| `hitmapOnMouseMove`   | `boolean`                                       |                        | if enabled, the clicked objectId will be returned from the mouse event handler                                                                        |
+| `children`            | `React.Node`                                    |                        | regl components to render inside the Worldview                                                                                                        |
+| `style`               | `{ [styleAttribute: string]: number | string }` | {}                     | html style object for the Worldview root element                                                                                                      |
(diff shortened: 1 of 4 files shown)


packages/regl-worldview/src/Worldview.js
```diff
@@ -165,6 +176,16 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
...
+    const { dragStartPos } = this.state;
+    if (dragStartPos) {
+      const deltaX = e.clientX - dragStartPos.x;
+      const deltaY = e.clientY - dragStartPos.y;
+      const distance = Math.sqrt(deltaX * deltaX + deltaY * deltaY);
+      if (distance > this.props.mouseClickRadius) {
+        this._disableOnClick = true;
+      }
+      this.setState({ dragStartPos: null });
+    }
```
janpaul123: If you want to simplify things a bit, you can even just call `onClick` here (if the mouse has not moved beyond the radius). Then you can get rid of `_disableOnClick`.

packages/regl-worldview/src/Worldview.js
```diff
@@ -104,6 +108,7 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
         cameraState: props.cameraState || props.defaultCameraState || DEFAULT_CAMERA_STATE,
         onCameraStateChange: props.onCameraStateChange || undefined,
       }),
+      dragStartPos: null,
```
robin-pham: we don't have to put this in the component `state` if it's not going to be used in the `render` method. Let's just keep it as a class property instead.

docs/src/3.1.Worldview.mdx
```diff
@@ -2,22 +2,23 @@
...
+| `hitmapOnMouseMove`   | `boolean`                                       |                        | if enabled, the clicked objectId will be returned from the mouse event handler                                         |
+| `children`            | `React.Node`                                    |                        | regl components to render inside the Worldview                                                                         |
+| `style`               | `{ [styleAttribute: string]: number | string }` | {}                     | html style object for the Worldview root element                                                                       |
+| `showDebug`           | `boolean`                                       |                        | show debug information                                                                                                 |
+| `mouseClickRadius`    | `number`                                        | 3                      | if the mouse has moved distance beyond the mouseClickRadius limit, the click event handler isn't going to be triggered |
```
robin-pham: Should we set it `Number.MAX_SAFE_INTEGER` by default, as this behavior would be more consistent with the DOM? Or should we just set sensible defaults and people will have to opt-out of this behavior?

docs/src/3.4.MouseEvents.mdx
```diff
@@ -75,3 +75,7 @@ Check out below example for automatic/custom hitmap mapping, and regular/instanc
 ## Full Example
 
 <MouseEvents />
+
+## Other
+
+We use `mouseClickRadius` prop to determine if mouseClick event is going to be triggered, specificially if the mouse has moved beyond the `mouseClickRadius` limit, the `onClick` event handler will be canceled, and `onMouseUp` will still get triggered if present.
```
robin-pham: probably best to also add a description of the common use case, i.e. this helps implement the desired behavior of dragging the camera around not firing `onClick` events

packages/regl-worldview/src/Worldview.js
```diff
@@ -149,14 +154,20 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
...
   };
 
   _onDoubleClick = (e: MouseEvent) => {
+    this._disableOnClick = false;
     this._onMouseInteraction(e, "onDoubleClick");
   };
 
   _onMouseDown = (e: MouseEvent) => {
+    this._disableOnClick = false;
+    this.setState({ dragStartPos: { x: e.clientX, y: e.clientY } });
```
brianc: instead of using state for this what about just setting it as an instance variable on the component? Will remove a render & draw call that I don't think we need.

docs/src/3.1.Worldview.mdx
brianc: might be nice to mention in the docs the number is number of pixels in screen space.  There's a very small potential one could think it means the distance in the scene.

packages/regl-worldview/src/Worldview.js
vidaaudrey: Good idea! Will update

docs/src/3.4.MouseEvents.mdx
vidaaudrey: Good idea! Will add

packages/regl-worldview/src/Worldview.js
vidaaudrey: Good suggestion. Updated

docs/src/3.1.Worldview.mdx
vidaaudrey: As discussed offline, DOM onClick doesn't get triggered if we are dragging objects around.  Here we detect dragging by the moving distance which is somewhat close to what we want to achieve.

packages/regl-worldview/src/Worldview.js
```diff
@@ -32,7 +33,7 @@ export type BaseProps = {|
   cameraState?: CameraState,
   onCameraStateChange?: (CameraState) => void,
   defaultCameraState?: CameraState,
-
+  mouseClickRadius: number,
```
janpaul123: I'm not sure if we should support this as a property or just hardcode it. The fewer properties the better, because each one of these we will have to support indefinitely (until we bump major versions).

janpaul123: (And I don't see any good reason for people to change this property?)

vidaaudrey: You are right. The fewer props the better. Just moved.

docs/src/jsx/utils/WorldviewCodeEditor.js
```diff
@@ -113,7 +114,7 @@ export const scope = {
   Text,
   GLTFScene,
   withPose,
-
+  getRayFromClick,
```
janpaul123: This change seems unrelated to the rest of this PR, did you mean to commit this?

docs/src/3.4.MouseEvents.mdx
```diff
@@ -75,3 +75,7 @@ Check out below example for automatic/custom hitmap mapping, and regular/instanc
 ## Full Example
 
 <MouseEvents />
+
+## Other
+
+We use `DEFAULT_MOUSE_CLICK_RADIUS` (3 pixels in the screen space) to determine if mouse click event is going to be triggered, specificially if the mouse has moved beyond the `DEFAULT_MOUSE_CLICK_RADIUS` limit, the `onClick` event handler will be canceled, and `onMouseUp` will still get triggered if present. This helps implement certain behaviors such as panning around worldview without firing `onClick` event.
```
robin-pham: we don't really need to know that the variable is called `DEFAULT_MOUSE_CLICK_RADIUS`. Either we don't mention anything about this, or we should have a small blurb e.g. "`onClick` events don't fire if the user moves their cursor more than 3 pixels between mouse down and mouse up"

docs/src/jsx/utils/WorldviewCodeEditor.js
vidaaudrey: I started out hoping to build the dragging example, but realized it's better to keep it in a separate PR. Overall it's good to have this function in the global scope of code playground just people can use it.

docs/src/3.4.MouseEvents.mdx
vidaaudrey: Updated: `onClick` events don't fire if the user moves their cursor for more than 3 screen pixels between mouse down and mouse up. This helps implement certain behaviors such as panning around worldview without firing `onClick` event.

PR #105, opened 2019-03-20
Prevent clicking through the Overlay in Interactivity example

vidaaudrey:
Before: clicking the Overlay will cause the objects underneath to be clicked and make the current Overlay disappear 
problem

Fixed: click the Overlay and the Overlay doesn't change
demo

### DIFF
--- a/docs/src/jsx/Hitmap.js
+++ b/docs/src/jsx/Hitmap.js
@@ -167,7 +167,7 @@ function Example() {
           return (
             <div
               onClick={(ev) => {
-                // prevent from clicking the objects underneath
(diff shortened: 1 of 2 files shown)


docs/src/jsx/Hitmap.js
```diff
@@ -163,8 +163,13 @@ function Example() {
             text,
             info: { color, title },
           } = item;
+
           return (
             <div
+              onClick={(ev) => {
+                // prevent from clicking the objects underneath
```
robin-pham: nit: prevents clicking on the objects underneath

PR #106, opened 2019-03-22
Update to latest LTS Node.js and Chrome 73 in Dockerfile

janpaul123:
I already published this as cruise/webviz-ci:0.0.6.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,7 @@ jobs:
             - regl-worldview-deps-{{ checksum "packages/regl-worldview/package-lock.json" }}
             - regl-worldview-deps-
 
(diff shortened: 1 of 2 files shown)


vidaaudrey:
> Which I needed for Puppeteer to work better

Could you explain a little more how Puppeteer is going to work better with the update? Is this also related to the Chrome Flexbox bug?

janpaul123:
> Could you explain a little more how Puppeteer is going to work better with the update? Is this also related to the Chrome Flexbox bug?

I was confused when I wrote that, this has nothing to do with Puppeteer.

Also I talked with @brianc and decided we should stay on the LTS release, so I updated the Dockerfile to point to the latest Node LTS. :)

PR #109, opened 2019-03-26
 Replace CodeSandbox embed with playground in tutorials

vidaaudrey:
## Problem 
Loading CodeSandbox inline makes the page slow and there are lots of console warnings/errors. In addition, updating CodeSandbox content is not very straightforward since they're linked to my personal sandboxes now.  (I started path as I was thinking about making the markdown content work in Medium article and get around the webpack issue for loading GLTF models. )

## Fix and features
- Use code playground like the other examples
- Load the duck model from CodeSandbox storage URL 
- Add tutorial example for orientation follow (thank @jtbandes  for providing the example)


docs/src/1.1.BasicExample.mdx
```diff
@@ -1,7 +1,7 @@
-import { BasicExample } from './jsx/allLiveEditors.js'
+import { BasicExample } from './jsx/allLiveEditors'
 
 # Basic Example
 
-A simple example to show how to set up Worldview to render simple shapes.
+A simple example to show how to set up `Worldview` to render simple shapes.
```
jtbandes: I don't think we should add this everywhere, Worldview is like a proper noun and it looks messy if we have to use `` ` `` all the time

docs/src/2.4.ManagingTheCamera.mdx
```diff
@@ -9,50 +10,25 @@ In this chapter, you'll learn how to:
...
+// ... other existing code
+<GLTFScene model={duckModel}>
+  {{
+    pose: {
+      position: { x: 0, y: 0, z: 0 },
+      orientation: { x: 0, y: 0, z: 0, w: 1 },
+    },
+    scale: { x: 3, y: 3, z: 3 },
+  }}
+</GLTFScene>;
```
jtbandes:
```suggestion
</GLTFScene>
```

docs/src/jsx/allLiveEditors.js
```diff
@@ -35,6 +35,36 @@ export const Arrows = makeCodeComponent(require("!!raw-loader!./Arrows"), "Arrow
 
 export const BasicExample = makeCodeComponent(require("!!raw-loader!./BasicExample"), "BasicExample");
 
+export const C11HelloWorld = makeCodeComponent(require("!!raw-loader!./C11HelloWorld"), "RenderingObjects");
```
jtbandes: what does C stand for? Maybe it's better to organize these by section, such as put them all in a Tutorial folder

docs/src/2.5.AddingInteractivity.mdx
```diff
@@ -33,20 +34,20 @@ Let's also enable manual camera movement, so we can see the duck better.
...
 // use `defaultCameraState` to turn Worldview to uncontrolled component
-<Worldview defaultCameraState={{ ...DEFAULT_CAMERA_STATE, distance: 120 }}>
+<Worldview defaultCameraState={{ ...DEFAULT_CAMERA_STATE, distance: 160 }}>
 ```
 
 Next, let's add a `clickedObjectIds` state to keep track of the clicked spheres. We'll also use this `clickedObjectIds` state to generate cube obstacles.
 
 ```js
-const [clickedObjectIds, setClickedObjectIds] = useState(new Set());
+const [clickedObjectIds, setClickedObjectIds] = React.useState([]);
```
jtbandes: I think it's common to `import React, { useState }` and just use `useState` rather than `React.useState`

docs/src/2.5.AddingInteractivity.mdx
```diff
@@ -82,55 +83,69 @@ Now, let's use the clicked sphere id to update the `clickedObjectIds` state, whi
...
 >
   {obstacleMarkers}
 </Cubes>
 ```
 
-Lastly, setting a `shouldStopDuck` state which will allow us to tell the duck when to stop and when to move forward.
+Now we can interact with the scene:
+
+- whenever we click on a sphere, a red cube will appear at the same clicked position
+- clicking a cube will make itself disappear
```
jtbandes:
```suggestion
- Clicking a cube will make it disappear.
```

docs/src/2.5.AddingInteractivity.mdx
```diff
@@ -82,55 +83,69 @@ Now, let's use the clicked sphere id to update the `clickedObjectIds` state, whi
...
   }}
 >
   {obstacleMarkers}
 </Cubes>
 ```
 
-Lastly, setting a `shouldStopDuck` state which will allow us to tell the duck when to stop and when to move forward.
+Now we can interact with the scene:
+
+- whenever we click on a sphere, a red cube will appear at the same clicked position
```
jtbandes:
```suggestion
- Whenever we click on a sphere, a red cube will appear in its place.
```

docs/src/routes.js
```diff
@@ -112,7 +112,7 @@ export function getHashUrlByComponentName(name) {
       });
     });
   }
-  return nameToUrlMap[name];
+  return nameToUrlMap[name] || "";
```
jtbandes: if the name isn't present in the map, is that a bug? should we throw an exception or return some obvious string like `"<unnamed>"` rather than `""` which might silently fail?

docs/src/jsx/allLiveEditors.js
vidaaudrey: C stands for chapter. Will move them all to a tutorial folder. Thought it's still nice to keep the `C11`, `C21` prefix so they are all in order.

docs/src/routes.js
vidaaudrey: this is mainly for the CodeSandbox in the home page, https://codesandbox.io/s/w0z0y5pyy8 
This change will turn `// docs: https://cruise-automation.github.io/webviz/worldview/#undefined` to `// docs: https://cruise-automation.github.io/webviz/worldview/# ` which is the correct URL

docs/src/jsx/allLiveEditors.js
robin-pham: the number suffix will be a little bit unwieldy if a chapter/subchapter is added before previous ones. Is it necessary to give them an order?

vidaaudrey: I don't have strong opinion about it. Since there are't so many of them right now, I'll just put all under `/tutorials` and remove the prefixes.

vidaaudrey: Thinking about writing more tutorials in the future, so the structure would be something like `/tutorials/duck`, `/tutorials/ballGame` ...

PR #113, opened 2019-03-28
Fix text disappearing after code update in CodeSandbox

vidaaudrey:
## Summary
As shown below, the text disappears after editing the code. We did some [perf improvement](https://github.com/cruise-automation/webviz/blob/master/packages/regl-worldview/src/commands/Text.js#L33)  by dynamically creating shared global styles for Text. However, the CodeSandbox is not persisting the style between renders.   

problem

Fixed by adding the needed styles to CodeSandbox examples, which can be configured through `makeCodeComponent`'s options. 

## Test plan
Manually test that changing the code won't cause the text to disappear. 
demo

## Versioning impact
No impact.

### DIFF
--- a/docs/src/jsx/allLiveEditors.js
+++ b/docs/src/jsx/allLiveEditors.js
@@ -10,7 +10,7 @@ import React from "react";
 
 import WorldviewCodeEditor from "./utils/WorldviewCodeEditor";
 
-function makeCodeComponent(raw, componentName, { isRowView, shouldInsertCodeSandboxStyleFix } = {}) {
+function makeCodeComponent(raw, componentName, { isRowView, insertCodeSandboxStyle } = {}) {
   const code = raw
     .split("// #BEGIN EXAMPLE")[1]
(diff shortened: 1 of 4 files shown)


docs/src/jsx/allLiveEditors.js
```diff
@@ -10,7 +10,7 @@ import React from "react";
 
 import WorldviewCodeEditor from "./utils/WorldviewCodeEditor";
 
-function makeCodeComponent(raw, componentName, isRowView) {
+function makeCodeComponent(raw, componentName, { isRowView, shouldInsertCodeSandboxStyleFix } = {}) {
```
esthersweon: #nit: Not sure if the "Fix" part of this variable is necessary or describing what it is doing. Do you think `insertCodeSandboxStyle` would be sufficient / more descriptive?

vidaaudrey: yeah, totally :) Updated.

PR #103, opened 2019-03-16
Add a quickstart tutorial

vidaaudrey:
Add quickstart tutorial for worldview. Probably easier to review/modify locally.

Outline: 
- Introduction
- Rendering Objects
- Managing the Camera
- Adding Interactivity 

image

image

image

PR #114, opened 2019-03-28
Skip rendering CameraListener if Worldview is fully-controlled

vidaaudrey:
## Summary
Suggested by @jtbandes  [here](https://github.com/cruise-automation/webviz/pull/103/files#r267911290): if the camera if fully controlled, we shouldn't allow the any mouse interactions, because the camera should be managed by the consumer. 

## Test plan
Manually test
**before**: moving the mouse will cause the object to shift a little and snap back right away [demo](https://cruise-automation.github.io/webviz/worldview/#/docs/tutorial/managing-the-camera):
before
**after**:  moving the mouse doesn't cause any shifting when Worldview is fully controlled. 
after

## Versioning impact
No impact if Worldview is correctly used for `controlled` and `uncontrolled` camera state.

### DIFF
--- a/packages/regl-worldview/src/Worldview.js
+++ b/packages/regl-worldview/src/Worldview.js
@@ -264,22 +264,21 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
   render() {
     const { width, height, showDebug, keyMap, shiftKeys, style, cameraState } = this.props;
     const { worldviewContext } = this.state;
-    const canvasHtml = [
-      <canvas
-        key={0}
-        style={{ width, height, maxWidth: "100%", maxHeight: "100%" }}
-        width={width}
-        height={height}
-        ref={this._canvas}
-        onMouseUp={this._onMouseUp}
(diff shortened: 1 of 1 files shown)


packages/regl-worldview/src/Worldview.js
```diff
@@ -262,24 +262,35 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
   }
 
   render() {
-    const { width, height, showDebug, keyMap, shiftKeys, style } = this.props;
+    const { width, height, showDebug, keyMap, shiftKeys, style, cameraState } = this.props;
     const { worldviewContext } = this.state;
+    const canvasHtml = [
```
janpaul123: #nit If you don't want to have to bother with `key` you could use `const canvasHtml = <React.Fragment>`

vidaaudrey: Good idea! I actually tried that first, but was getting some weird lint?/flow? error so I gave up. Now it's all gone, weird.

PR #116, opened 2019-04-01
Add @cruise-automation/hooks package

vidaaudrey:
## Summary
Consolidating a few shared React Hooks into a single package. 

- useAbortable
- useCleanup
- useEventListener
- useAnimationFrame

## Test plan
- unit tests 
- use `npm pack` to test the package locally with Create-React-App 

## Versioning impact
No impact. 

_Will migrate and delete old files in a separate PR._


packages/@cruise-automation/hooks/README.md
```diff
@@ -0,0 +1,50 @@
...
+npm install --save @cruise-automation/hooks
+```
+
+### useAbortable
+
+```js
+import { useAbortable, useConstant, useEventListener, useRequestAnimationFrame } from "@cruise-automation/hooks";
+
+// Abort an fetchData action
+const [remoteData] = useAbortable(
```
jtbandes: Second return value (abort function) should not be ignored.

packages/@cruise-automation/hooks/src/useConstant.js
```diff
@@ -0,0 +1,20 @@
...
+import { useState, useEffect } from "react";
+
+// for sharing the same instance variable during the react life cycle
+export default function useConstant<T>(getValFn: () => T, teardown: (T) => any = () => {}): T {
+  const [value] = useState<T>(getValFn);
+  useEffect(() => {
+    return () => {
+      teardown(value);
+    };
+  }, []);
```
janpaul123: Per discussion in person, we should probably just make this line into a `useCleanup` and then get rid of `useConstant` altogether. Then you would use
```js
const [myThing] = useState(() => new Thing());
useCleanup(() => myThing.destroy());
```

instead of 
```js
const myThing = useConstant(() => new Thing(), (thing) => thing.destroy());
```
(which is a bit less clear IMO)

packages/@cruise-automation/hooks/babel.config.js
```diff
@@ -4,9 +4,6 @@
 //  found in the LICENSE file in the root directory of this source tree.
 //  You may not use this file except in compliance with the License.
 
-const baseConfig = require("../../../babel.config");
+const baseConfig = require("../../regl-worldview/babel.config");
```
janpaul123: Should we pull this out to a different directory, instead of pulling files from a sibling directory? That can be pretty confusing after all.

packages/@cruise-automation/hooks/src/index.js
```diff
@@ -0,0 +1,10 @@
+//  Copyright (c) 2018-present, GM Cruise LLC
+//
+//  This source code is licensed under the Apache License, Version 2.0,
+//  found in the LICENSE file in the root directory of this source tree.
+//  You may not use this file except in compliance with the License.
+
+export { default as useAbortable } from "./useAbortable";
```
jtbandes: Is there any way to do these automatically?

packages/@cruise-automation/hooks/README.md
```diff
@@ -0,0 +1,53 @@
...
+## Usage
+
+```bash
+npm install --save @cruise-automation/hooks
+```
+
+```js
+import { useAbortable, useCleanup, useEventListener, useAnimationFrame } from "@cruise-automation/hooks";
+
+//*************** useAbortable **********************
```
jtbandes: why not use markdown headings `###` for these?

packages/@cruise-automation/hooks/rollup.config.js
```diff
@@ -0,0 +1,10 @@
+//  Copyright (c) 2018-present, GM Cruise LLC
+//
+//  This source code is licensed under the Apache License, Version 2.0,
+//  found in the LICENSE file in the root directory of this source tree.
+//  You may not use this file except in compliance with the License.
+
+import getSharedRollupConfig from "../../../package-shared/getSharedRollupConfig";
```
jtbandes: what about calling this directory `common` or `config`?

also, `get` seems uncommon in JS even for functions like this. consider calling it just `sharedRollupConfig`

package-shared/getSharedRollupConfig.js
```diff
@@ -0,0 +1,83 @@
+//  Copyright (c) 2018-present, GM Cruise LLC
+//
+//  This source code is licensed under the Apache License, Version 2.0,
+//  found in the LICENSE file in the root directory of this source tree.
+//  You may not use this file except in compliance with the License.
+
+import babel from "rollup-plugin-babel";
```
jtbandes: flow?

packages/@cruise-automation/hooks/src/useAnimationFrame.test.js
```diff
@@ -0,0 +1,74 @@
...
+        return cb(MOCK_TIMESTAMP);
+      }
+    });
+  });
+
+  afterEach(() => {
+    window.requestAnimationFrame.mockRestore();
+  });
+
+  it("call the callback at each requestAnimationFrame", () => {
```
jtbandes:
```suggestion
  it("calls the callback upon each animation frame", () => {
```

packages/@cruise-automation/hooks/src/useAnimationFrame.test.js
```diff
@@ -0,0 +1,74 @@
...
+    renderHook(() => useAnimationFrame(cb, false, []));
+    expect(count).toBe(3);
+  });
+
+  it("calls the callback with timestamp", () => {
```
jtbandes: can we just merge this with the previous test using `.mock.calls`?

packages/@cruise-automation/hooks/src/useAnimationFrame.test.js
```diff
@@ -0,0 +1,74 @@
...
+//  You may not use this file except in compliance with the License.
+
+import { renderHook } from "react-hooks-testing-library";
+
+import useAnimationFrame from "./useAnimationFrame";
+
+const MOCK_TIMESTAMP = 10336878.725;
+describe("useAnimationFrame", () => {
+  let rafExecutionCount = 0;
+  let maxExecutionCounnt = 3;
```
jtbandes: typo: counnt

packages/@cruise-automation/hooks/src/useAnimationFrame.test.js
```diff
@@ -0,0 +1,74 @@
...
+  });
+
+  it("can stop and continue to execute when disable and dependency changes", () => {
+    let input = [1];
+    const { rerender } = renderHook(() => useAnimationFrame(cb, false, input));
+    expect(count).toBe(3);
+
+    // stop execution when disable is true
+    maxExecutionCounnt = 6;
+    rerender(cb, true, [cb, true, 1]);
```
jtbandes: I'm confused by this, `rerender` takes props to pass to the callback, but your callback doesn't accept any args.

packages/@cruise-automation/hooks/src/useCleanup.js
```diff
@@ -0,0 +1,18 @@
...
+//  Copyright (c) 2018-present, GM Cruise LLC
+//
+//  This source code is licensed under the Apache License, Version 2.0,
+//  found in the LICENSE file in the root directory of this source tree.
+//  You may not use this file except in compliance with the License.
+
+import { useEffect } from "react";
+
+// for sharing the same instance variable during the react life cycle
+export default function useCleanup(teardown: Function) {
```
jtbandes: don't use `Function`, it's treated as `any`. `() => void` is probably fine

packages/@cruise-automation/hooks/src/useCleanup.test.js
```diff
@@ -0,0 +1,30 @@
...
+//  This source code is licensed under the Apache License, Version 2.0,
+//  found in the LICENSE file in the root directory of this source tree.
+//  You may not use this file except in compliance with the License.
+import { renderHook } from "react-hooks-testing-library";
+
+import useCleanup from "./useCleanup";
+
+describe("useCleanup", () => {
+  class ExpensiveThing {
+    name: string;
```
jtbandes: nit: `destroyed = false;` :)

packages/@cruise-automation/hooks/src/useAbortable.js
```diff
@@ -0,0 +1,52 @@
...
+//
+//  Copyright (c) 2018-present, GM Cruise LLC
+//
+//  This source code is licensed under the Apache License, Version 2.0,
+//  found in the LICENSE file in the root directory of this source tree.
+//  You may not use this file except in compliance with the License.
+
+import { useState, useEffect } from "react";
+
+// translation of https://dom.spec.whatwg.org/#abortcontroller
```
jtbandes: link to https://github.com/flow-typed/flow-typed/issues/1652

packages/@cruise-automation/hooks/README.md
```diff
@@ -0,0 +1,53 @@
...
+  (val) => {}, // clean up work if there is any
+  ["dataName"] // remount when dependencies change
+);
+
+// call abortFn somewhere else, e.g. before component unmounts
+abortFn();
+
+//*************** useCleanup **********************
+// Call the cleanup function before the component unmounts
+const [audioContent] = React.useState(() => new window.AudioContext));
```
jtbandes:
```suggestion
const [audioContext] = React.useState(() => new window.AudioContext));
```

packages/@cruise-automation/hooks/README.md
```diff
@@ -0,0 +1,53 @@
...
+//*************** useAbortable **********************
+// Abort an fetchData action
+const [remoteData, abortFn] = useAbortable(
+  [], // default value
+  async (abortController) => fetchDataFromRemote(props, "dataName"), // action to perform
+  (val) => {}, // clean up work if there is any
+  ["dataName"] // remount when dependencies change
+);
+
+// call abortFn somewhere else, e.g. before component unmounts
```
jtbandes: Doesn't useAbortable already call it when the component is unmounting?

packages/@cruise-automation/hooks/rollup.config.js
vidaaudrey: Good idea. I changed `package-shared/getSharedRollupConfig` to `common/rollup.config.js`

packages/@cruise-automation/hooks/src/useAnimationFrame.test.js
vidaaudrey: It should be `rerender(cb, true, input);`, I forgot to remove `cb` and `true` from the dependencies.  Not sure if this makes it clear

packages/@cruise-automation/hooks/README.md
```diff
@@ -0,0 +1,136 @@
...
+  args: any // dependencies
+): [T, () => void] // result value and abort function
+```
+
+```js
+// sample usage
+import { useAbortable } from "@cruise-automation/hooks";
+
+function Example(props) {
+  const [data, setDatta] = useState(null);
```
jtbandes:
```suggestion
  const [data, setData] = useState(null);
```

packages/@cruise-automation/hooks/README.md
```diff
@@ -0,0 +1,136 @@
...
+  const dateName = 'dateName';
```
jtbandes:
```suggestion
  const dataName = 'foo';
```

packages/@cruise-automation/hooks/README.md
```diff
@@ -0,0 +1,136 @@
...
+  // fetch data from remote when the component is mounted
+  const [remoteData, abortFn] = useAbortable(
+    null,
+    async (abortController) =>  {
+      if(dataName) {
```
jtbandes:
```suggestion
      if (dataName) {
```

packages/@cruise-automation/hooks/README.md
```diff
@@ -0,0 +1,136 @@
...
+  );
+
+  // abort is usually called when the component unmounts, but it can also be called manually
+  function abortManually() {
+    abortFn();
+  }
+
+  return (
+    <div>
+      <button onClick={abortManually}> Abort Manually Now</button>
```
jtbandes:
```suggestion
      <button onClick={abortManually}>Stop Loading</button>
```

packages/@cruise-automation/hooks/README.md
```diff
@@ -0,0 +1,136 @@
...
+    </div>
+  );
+}
+```
+
+### `useCleanup`
+
+A small React Hook to fire the cleanup callback when the component unmounts.
```
jtbandes:
```suggestion
A small React Hook to fire the cleanup callback when the component unmounts. Equivalent to `useEffect(() => () => { teardown(); }, [])`.
```

packages/@cruise-automation/hooks/README.md
```diff
@@ -0,0 +1,136 @@
...
+  const [audioContext] = React.useState(() => new window.AudioContext));
+  // automatically close audioContext when the component unmounts
+  useCleanup(() => audioContext.close());
+  return null;
+}
+```
+
+### `useEventListener`
+
+A handy React Hook for automatically adding and removing event listeners.
```
jtbandes:
```suggestion
A hook for conditionally adding and removing event listeners on DOM elements.
```

packages/@cruise-automation/hooks/README.md
```diff
@@ -0,0 +1,136 @@
...
+```
+
+```js
+// sample usage
+function Example() {
+  const [isDragging, setIsDragging] = useState(false);
+  useEventListener(
+    window,
+    "mousemove",
+    isDragging, // enable during mouse dragging
```
jtbandes:
```suggestion
    isDragging, // add this listener only during mouse dragging
```

packages/@cruise-automation/hooks/README.md
```diff
@@ -0,0 +1,136 @@
...
+    },
+    []
+  );
+  return <div onMouseDown={() => setIsDragging(true)}>demo</div>;
+}
+```
+
+### `useAnimationFrame`
+
+A React Hook that accepts a callback function which will be called before each animation frame.
```
jtbandes:
```suggestion
A React Hook that accepts a callback function which will be called repeatedly, synchronized with the browser's repaints via [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame).
```

packages/@cruise-automation/hooks/README.md
```diff
@@ -0,0 +1,136 @@
...
+// types
+useAnimationFrame(
+  callback: (timestamp: number) => void,
+  disable: boolean,
+  dependencies: any[],
+): void
+```
+
+```js
+// sample useage: a count state that gets updated in every animation frame
```
jtbandes:
```suggestion
// sample usage: a count state that gets updated in every animation frame
```

packages/@cruise-automation/hooks/README.md
```diff
@@ -0,0 +1,136 @@
...
+function Example() {
+  const [count, setCount] = React.useState(0);
+  useAnimationFrame(
+    () => {
```
jtbandes:
```suggestion
    (timestamp) => {
```

packages/@cruise-automation/hooks/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+### `useAnimationFrame`
+
+A React Hook that accepts a callback function which will be called before each animation frame.
+
+```js
+// types
+useAnimationFrame(
+  callback: (timestamp: number) => void,
+  disable: boolean,
```
jtbandes: it's a bit weird to use `enable` in useEventListener but `disable` in useAnimationFrame.

vidaaudrey: For `useEventListener`, the consumer would normally do some sort of `isDragging` state  => `enable` listener mapping, and there is need to think about defaults, so keeping `enable` prop name makes sense.  

For `useAnimationFrame`, if the user doesn't pass in `disable` prop, we'll just enable `useAnimationFrame` by default which is what's expected if people are using this hook. 

So overall, yeah, it's weird, but maybe not enough of a reason to keep them the same?

PR #117, opened 2019-04-01
Update packages/webviz-core from internal repo (update to React 16.8)

jtbandes:
Updating to React 16.8 and removing `react-with-hooks`.

## Test plan

Mostly unit tests with a bit of manual testing.

## Versioning impact

None, not changing public API


package.json
```diff
@@ -97,11 +97,12 @@
   },
   "dependencies": {},
   "scripts": {
-    "bootstrap": "npm install && lerna exec npm install",
-    "install-ci": "npm ci && lerna exec npm ci",
+    "bootstrap": "npm install && lerna bootstrap --hoist react",
```
vidaaudrey: Curious why only hoisting react but not react-dom?

jtbandes: I think it doesn't matter for tests because react-dom isn't used. I tried using a glob like `--hoist 'react|react-dom|...'` but I had some other errors with that

janpaul123: It's unfortunately that comments in json are not really possible because this would be a great place to have a comment…

PR #124, opened 2019-04-09
 Add option to customize blend and depth

vidaaudrey:
## Summary
We've encountered a few [bugs](https://github.com/cruise-automation/webviz/pull/100) where triangles' depth value was causing weird behaviors demonstrated as below: the stop sign is always behind the axes no matter where we place the camera.
https://codesandbox.io/s/qkv1793yrw 

before

## Fix 
Update existing commands to take `blend` and `depth` as prop in order to customize per draw call, e.g. 
```
<Triangles>
{[
  depth: { enable: true, mask: false },
  blend: { enable: false },
  pose: {
      position: { x: 0, y: 0, z: 0 },
      orientation: { x: 0, y: 0, z: 0, w: 1 },
    },
    points: ...,
    colors: ...,
]}
</Triangles>
```
## Test plan
With the change, the `blend` and `depth` props are now transparent, the consumer should see the expected behaviors when they pass in the right props.  
Also manually test that the composition example is fixed: the stop sign does show in front of the axes.  
after

## Versioning impact
Will do a minor release since this PR adds 2 new APIs.

### DIFF
--- a/packages/regl-worldview/src/commands/Triangles.js
+++ b/packages/regl-worldview/src/commands/Triangles.js
@@ -18,7 +18,8 @@ import {
 import { getHitmapPropsForInstancedCommands, getObjectForInstancedCommands } from "../utils/hitmapDefaults";
 import { makeCommand } from "./Command";
 
-const defaultSingleColorDepth = { enable: true, mask: true };
+// TODO(Audrey): default to the actual regl defaults before 1.x release
+const defaultSingleColorDepth = { enable: true, mask: false };
 const defaultVetexColorDepth = {
   enable: true,
   func: "<=",
--- a/packages/regl-worldview/src/commands/Triangles.js
+++ b/packages/regl-worldview/src/commands/Triangles.js
@@ -65,7 +65,7 @@ const singleColor = (regl) =>
         return props.color;
       },
     },
-    // can pass in {enable: true, depth: true } to turn off depth to prevent flicker
+    // can pass in { enable: true, depth: true } to turn off depth to prevent flicker
     // because multiple items are rendered to the same z plane
     depth: (context, props) => props.depth || defaultDepth,
     blend: (context, props) => props.blend || defaultBlend,
--- a/packages/regl-worldview/src/commands/Triangles.js
+++ b/packages/regl-worldview/src/commands/Triangles.js
@@ -9,7 +9,6 @@
 import type { TriangleList, Regl } from "../types";
 import {
   defaultBlend,
-  defaultDepth,
   getVertexColors,
   pointToVec3Array,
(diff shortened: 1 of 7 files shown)


brianc:
Would it be possible to have the api like this?

```js
<Triangles depth={depthConfig} blend={blendConfig}>[triangleObjects]</Triangles>
```

And have the depth and blend be component-level props instead of object level props?  I think it will make batching draw calls easier if we control the blend/depth on a component level versus the instance level.  What do you think?

packages/regl-worldview/src/commands/Triangles.js
```diff
@@ -109,11 +115,12 @@ const vertexColors = (regl) =>
...
-    depth: {
-      enable: true,
-      func: "<=",
-    },
+    blend: (context, props) => props.blend || defaultBlend,
+    depth: (context, props) =>
+      props.depth || {
+        enable: true,
+        func: "<=",
+      },
```
brianc: nit: can we move the inline object into a `const` declared outside the function? Don't need to realloc this object every time we call this function if its unchanging.

packages/regl-worldview/src/commands/Triangles.js
```diff
@@ -57,12 +65,10 @@ const singleColor = (regl) =>
...
       },
     },
-
-    // turn off depth to prevent flicker
+    // can pass in { enable: true, depth: true } to turn off depth to prevent flicker
     // because multiple items are rendered to the same z plane
-    depth: { enable: true, mask: false },
-    blend,
-
+    depth: (context, props) => props.depth || defaultDepth,
```
brianc: the new value here stored in `commandUtils` for `defaultDepth` is actually different than the old value of `{ enable: true, mask: false }`

packages/regl-worldview/src/utils/commandUtils.js
```diff
@@ -98,6 +98,8 @@ export const blend = {
   },
 };
 
+export const defaultDepth = { enable: true, mask: true };
```
brianc: if this value is only used in `triangles.js` it's probably better to colocate it w/ that file if/until it's needed in other places as well.

packages/regl-worldview/src/commands/Triangles.js
vidaaudrey: As chatted offline, the default value produces visual bugs for triangles and having `mask` default to `true` fixes it.

packages/regl-worldview/src/utils/commandUtils.js
vidaaudrey: Good suggestion!

PR #132, opened 2019-04-18
Add mouse interaction to GLTFScene

vidaaudrey:
## Summary
Add hitmap support for `GLTFScene` so the user can interact with the 3D models. 

## Test plan
Add `DuckSceneHitmap.js` example:
now

## Versioning impact
Minor: new api added and it's backward compatible.

### DIFF
--- a/packages/regl-worldview/src/commands/GLTFScene.js
+++ b/packages/regl-worldview/src/commands/GLTFScene.js
@@ -9,8 +9,8 @@
 import { mat4 } from "gl-matrix";
 import React from "react";
 
-import type { Pose, Scale } from "../types";
-import { defaultBlend, pointToVec3, orientationToVec4 } from "../utils/commandUtils";
+import type { Pose, Scale, MouseHandler } from "../types";
+import { defaultBlend, pointToVec3, orientationToVec4, intToRGB } from "../utils/commandUtils";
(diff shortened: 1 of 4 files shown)


packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -250,15 +261,30 @@ export default class GLTFScene extends React.Component<Props, {| loadedModel: ?O
...
       <WorldviewReactContext.Consumer>
         {(context) => {
           this._context = context;
-          return <Command reglCommand={drawModel} drawProps={{ model: loadedModel, ...this.props.children }} />;
+          return (
+            <Command
+              {...rest}
+              reglCommand={drawModel}
+              drawProps={sharedDrawProps}
+              hitmapProps={drawHitmap ? { ...sharedDrawProps, drawHitmap } : undefined}
```
jtbandes: Seems strange that you tell the command about hitmap in 2 different ways: by passing `hitmapProps` _and_ by including a `drawHitmap` key inside the hitmap props. Only one of these should be necessary right? The command renderer can decide what to do based on whether there is an id.

Having to use `(ctx, props) => !!props.x` instead of `regl.prop("x")` is a bit of a code smell too; if you expect a boolean then the user should pass in a boolean.

PR #134, opened 2019-04-23
Improve GLTF hitmap logic

vidaaudrey:
## Summary
Address Jacob's [comment](https://github.com/cruise-automation/webviz/pull/132/files#r277456567): remove redundant `drawHitmap` and always draw hitmap if `id` is present. This is helpful for preventing clicking through objects in the 3d space. (Currently Worldview doesn't automatically draw hitmap for all which may cause some overlay clicking issue. We'll discuss and address it later.)

## Test plan
The existing GLTF hitmap example continues to work as expected. 

## Versioning impact
No impact

### DIFF
--- a/packages/regl-worldview/src/commands/GLTFScene.js
+++ b/packages/regl-worldview/src/commands/GLTFScene.js
@@ -261,18 +261,13 @@ export default class GLTFScene extends React.Component<Props, {| loadedModel: ?O
   }
 
   render() {
-    const {
-      children,
-      children: { id, ...restOfChildren },
-      ...rest
-    } = this.props;
-
+    const { children, ...rest } = this.props;
     const { loadedModel } = this.state;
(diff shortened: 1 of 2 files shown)


packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -280,8 +282,8 @@ export default class GLTFScene extends React.Component<Props, {| loadedModel: ?O
             <Command
               {...rest}
               reglCommand={drawModel}
-              drawProps={sharedDrawProps}
-              hitmapProps={drawHitmap ? { ...sharedDrawProps, drawHitmap } : undefined}
+              drawProps={{ ...restOfChildren, model: loadedModel }}
```
jtbandes: Could just use `...children, id: null` and then you don't need `restOfChildren`.

PR #136, opened 2019-05-14
Allow for multiple GLTFScenes

cfoust:
## Summary

You can only use one model across all `GLTFScene`s. `drawModel` keeps some internal state (`drawCalls`) which is not generated again when you provide different models. This is because `WorldViewContext._commands` is a `Set` and just checks to see whether `drawModel` has ever been registered in `onMount`. Since it has, subsequent `GLTFScene`s keep the same `drawCalls` state and therefore just draw the first mesh that got loaded.

The simplest solution here (if not a bit naive) is to hide `drawModel` behind a lambda in `componentDidMount` so that individual instances of `GLTFScene` all get `reglCommand`s with unique references.

## Test plan

Reproduced locally, I'm not sure how y'all usually test this stuff.

## Versioning impact

No idea, feel free to edit this PR

### DIFF
--- a/packages/regl-worldview/src/commands/GLTFScene.js
+++ b/packages/regl-worldview/src/commands/GLTFScene.js
@@ -232,6 +232,8 @@ export default class GLTFScene extends React.Component<Props, {| loadedModel: ?O
     loadedModel: undefined,
   };
   _context = undefined;
+  _drawModel = undefined;
+
   async _loadModel(): Promise<Object> {
     const { model } = this.props;
     if (typeof model === "function") {
@@ -248,6 +250,7 @@ export default class GLTFScene extends React.Component<Props, {| loadedModel: ?O
   }
 
   componentDidMount() {
+    this._drawModel = (regl) => drawModel(regl);
     this._loadModel()
       .then((loadedModel) => {
         this.setState({ loadedModel });
@@ -276,7 +279,7 @@ export default class GLTFScene extends React.Component<Props, {| loadedModel: ?O
           return (
             <Command
               {...rest}
-              reglCommand={drawModel}
+              reglCommand={this._drawModel}
               drawProps={{ ...children, id: null, model: loadedModel }}
               hitmapProps={drawHitmap ? { ...children, model: loadedModel } : undefined}
               getObjectFromHitmapId={(objId, hitmapProps) => (hitmapProps.id === objId ? hitmapProps : undefined)}


jtbandes:
Great find, thanks! You can tell we didn't test this use case 😅

I think it might be better to just fix `drawModel` so that rather than keeping global mutable state it just uses regular memoization (we already have `reselect` as a dependency, and there's precedent for using `memoize-one` in another package).

The downside to this would be that you end up keeping all the prepared draw calls for every model ever used since there's no opportunity to clear the memoized calls. What do you think @brianc @vidaaudrey?

cfoust:
I tried that approach and ran into problems coming up with a general solution for deciding whether we'd already prepared the `drawCalls` for a given model or not. Ideally you'd just look at the filename (or I suppose since you allow `Promise`s as input, the hash of the byte array?) to determine that, but it didn't seem that clean to me.

It looks like I'm failing Flow anyway, but I'll hold off on fixing that until we get some more input on how to solve this! I don't mind working to get this fixed for everyone rather than using the hack in this PR.

jtbandes:
What about just `===` / reference equality on the model?

vidaaudrey:
What about using the React key to unmount the component each time when model prop changes? Something like:

```
let id = 0;

export default function GLTFWrapper(props: Props) {
  React.useEffect(() => {
    return () => {
      id++;
    };
  }, []);

  return <GLTFScene {...props} key={props.key || id} />;
}
```

cfoust:
I'll give it a go with reference equality some time yet this week, that's a good idea. @vidaaudrey This is a good workaround, too! I should have thought of that. For the time being I'll give the general solution with reference equality a try and we'll see how it goes.

clement91190:
ah thanks was looking for this

clement91190:
Actually I'm still struggling to use 2 GLTFScenes with 2 different models.  @cfoust What is the workaround you are using in the end ? Do you have an example ? . Happy to help.

cfoust:
Heh, I never got around to taking a look at this again. @clement91190 the workaround is the patch this PR introduces. All it does is isolates the draw commands so that the same one doesn't get reused again.

clement91190:
ah thanks, I will take another look, I can't get it to work ( it always load the same object multiple times) . will keep you posted

vidaaudrey:
I created this PR based on the discussions and proposed changes. Let me know what you all think: https://github.com/cruise-automation/webviz/pull/145

vidaaudrey:
Closing in favor of  #145

PR #143, opened 2019-05-21

esthersweon:
Changelog: Add linting for react-hooks

## Summary

Added linting for react-hooks to keep React components code consistent and clean.

## Test plan

Added linting rule and fixed all resulting lint errors.

## Versioning impact

None.


packages/webviz-core/src/players/RandomAccessPlayer.test.js
```diff
@@ -597,6 +597,7 @@ describe("RandomAccessPlayer", () => {
       initialized(): void {
         this._initialized++;
       }
+      isPlaying(): void {}
```
janpaul123: I don't think this should be necessary? Where is this coming from?

packages/webviz-core/src/util/hooks.js
```diff
@@ -1,38 +0,0 @@
-// @flow
```
janpaul123: These probably should not get deleted (even though they're not currently used in this repo). Or do you plan on moving them out of webviz-core?

vidaaudrey: We discussed. Think @esthersweon  is going to delete the file from webviz-core  as well in her current PR. This will help testing her change in CI.

packages/webviz-core/src/players/RandomAccessPlayer.test.js
esthersweon: It was part of Robbie's commit here: https://github.robot.car/cruise/webviz/commit/2abbc7e1ec698522d4440dbcf1659f80dc0c2c7b#diff-8560f843168c34bfbe7d4ee8ed01a653

packages/webviz-core/src/util/hooks.js
esthersweon: Yep - this is the PR that will delete from webviz-core in the internal repo (as well as change all references to these deleted exports): https://github.robot.car/cruise/webviz/pull/1554

PR #145, opened 2019-06-04
Allow GLTFScene to load different models

vidaaudrey:
## Summary
Based on the suggestions and changes on #136, I created this PR to add support for loading multiple GLTF models. Thanks @cfoust for the initial work. 

## Test plan
Add an example to the docs to switch models:
after

## Versioning impact
Will publish a patch version since it's a fix.


clement91190:
Thanks @vidaaudrey @cfoust, working great for multiple GLTFScene loaded at the same time as well ! We plan on using this tool for our robot fleet visualization tools. 
ezgif com-video-to-gif

cfoust:
I looked through this, this is great! Will let me revert the hack we're using in our fork 👍 
@clement91190 That's amazing! It's looking really good -- we're working on similar tooling, too.

vidaaudrey:
@clement91190  thanks for sharing your work! It looks great! Just curious, do you see any performance issue when loading more models?

packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -287,3 +291,19 @@ export default class GLTFScene extends React.Component<Props, {| loadedModel: ?O
     );
   }
 }
+
+// Add unique key to GLTFScene so that it will unmount when the model changes
+export default function GLTFWrapper(props: Props) {
```
jtbandes: Did you consider doing something like `reglCommand = getMemoizedCommand(model, regl)` with simple memoization? I think it would be a simpler solution than the wrapper.

If keeping the wrapper function, please set its `displayName` so that React dev tools work better.

packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -287,3 +291,19 @@ export default class GLTFScene extends React.Component<Props, {| loadedModel: ?O
...
+  const [model, setModel] = useState(props.model);
+  const [id, setId] = useState(0);
+  React.useEffect(
+    () => {
+      if (props.model !== model) {
+        setModel(props.model);
+        setId((id + 1) % 1000000);
```
jtbandes: `% 1000000` is not needed

vidaaudrey:
> I still think memoization might be a simpler solution, I would recommend spending a few minutes to try it and see if you can clean up the code that way. However this is fine as well.

@jtbandes   tried a different approach to cache the drawCalls with a map. The size of the map (# of key-val pairs) will equal to the models we load and we can clear the map when it reaches certain size.  Let me know what you think.

vidaaudrey:
Thanks @jtbandes  for the help. 

@clement91190 @cfoust  regl-worldview@0.1.2 is published with multi-model support. Example:
https://codesandbox.io/s/confident-newton-osi77 Let us know if you see any issues

PR #157, opened 2019-06-18
update github page

jxjshaw:
html landing page to highlight webviz functionality. manually tested on popular browsers and devices


esthersweon:
Build seems to be failing on lint. Will address and ensure build succeeds before merging.

Issue #161, opened 2019-06-20
No explanation of how to run


Hodapp87:
Early on it says, "Webviz is a web-based application for playback and visualization of ROS bag files." and links you to a Webviz demo, however, I cannot find any explanation of how to actually serve this web application myself and get to it from my browser.

After a bit of messing around it looks like the proper way is running `npm run docs` and then browsing to http://localhost:8080/try (just http://localhost:8080 doesn't really work, since the 'view demo' links aren't relative but are to webviz.io) - but this is mostly from guesswork on my part, and it's not at all apparent.

janpaul123:
Yeah, documentation for development is super lacking at this point. I at least added a bit more info for how to run in #160. But if you want to just use the tool itself with your ROS bags (and don't care about development) you can just use https://webviz.io/try/ and drag in your bags.

asimay:
hi, @janpaul123 , I did as you said, I dragged the bagfile into webviz, but why the camera image cannot show out? I checked the message type is right.

![image](https://user-images.githubusercontent.com/8703449/60315096-b8bcc080-9997-11e9-895c-a8bf579330d1.png)

asimay:
and we don't have internet on host-machine, it is setup on car, so how to deploy your tools on our server?

jtbandes:
Hi @asimay,

- For your camera image — the Image panel currently doesn't support the `mono8` encoding. The supported `sensor_msgs/Image` encodings are here (note that we also allow `sensor_msgs/CompressedImage`): https://github.com/cruise-automation/webviz/blob/9be1ffe5db4971a991a018ce6413be76a7a18261/packages/webviz-core/src/panels/ImageView/ImageCanvas.js#L88-L91 (And we're going to add the other `bayer_*8` variants very soon.) If you're interested in contributing to Webviz, we would welcome a PR to add support for mono8! Otherwise, we'll take note of this feature request, and we can definitely add a better error message in the Image panel so the problem is more obvious 😉

- For deploying Webviz, currently the best option is `npm run build` as documented in [the readme](https://github.com/cruise-automation/webviz/blob/master/README.md). That will compile the package with webpack, so it can be run with node. It would be possible to, for example, create a Node.js Docker image that runs the server. Alternatively, the JS bundle could be loaded from a static HTML page, served with any simple web server. Feel free to file a new issue if you get stuck with this.

jtbandes:
@asimay mono8 and mono16 encodings are supported now as of #174. Give it a try!

[...]

PR #165, opened 2019-07-01
Mono8 support

jxjshaw:
Image panel support for mono8, details can be found in this convo: https://github.com/cruise-automation/webviz/issues/161#issuecomment-506621910

mono88

mono888887

source of sample bag: https://www.cse.unr.edu/~kalexis/datasets/thermal-inertial/

Btw encountered some weirdness with /packages/@cruise-automation not being found. Assuming it was an issue with my fork/local npm but flagging just in case

### DIFF
--- a/packages/webviz-core/package-lock.json
+++ b/packages/webviz-core/package-lock.json
@@ -15,19 +15,19 @@
         "regenerator-runtime": {
           "version": "0.12.1",
           "resolved": "https://registry.npmjs.org/regenerator-runtime/-/regenerator-runtime-0.12.1.tgz",
-          "integrity": "sha512-odxIc1/vDlo4iZcfXqRYFj0vpXFNoGdKMAUieAlFYO6m/nl5e9KR/beGf41z4a1FI+aQgtjhuaSlDxQ0hmkrHg=="
+          "integrity": "sha1-+hpxVEdkwDb4xJsToIsllMn4oN4="
         }
       }
     },
     "@mdi/svg": {
       "version": "3.2.89",
       "resolved": "https://registry.npmjs.org/@mdi/svg/-/svg-3.2.89.tgz",
-      "integrity": "sha512-X9DG/qaj5KkNJX/d7G41MyzkXbTWwWVGifs4yWeIZTsGoMZfzjRutz/RO1o09dZpMbcDvU6oy7bcpPWOfiS55Q=="
+      "integrity": "sha1-gkjs2i2eBFI+cq2xcn49YErtzY4="
     },
     "ansi-styles": {
       "version": "3.2.1",
       "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-3.2.1.tgz",
-      "integrity": "sha512-VT0ZI6kZRdTh8YyJw3SMbYm/u+NqfsAxEpWO0Pf9sq8/e94WxxOpPKx9FR1FlyCtOVDNOQ+8ntlqFxiRc+r5qA==",
+      "integrity": "sha1-QfuyAkPlCxK+DwS43tvwdSDOhB0=",
       "requires": {
         "color-convert": "^1.9.0"
       },
@@ -35,7 +35,7 @@
         "color-convert": {
           "version": "1.9.3",
           "resolved": "https://registry.npmjs.org/color-convert/-/color-convert-1.9.3.tgz",
-          "integrity": "sha512-QfAUtd+vFdAtFQcC8CCyYt1fYWxSqAiK2cSD6zDB8N3cpsEBAvRxp9zOGg6G/SHHJYAT88/az/IuDGALsNVbGg==",
+          "integrity": "sha1-u3GFBpDh8TZWfeYp0tVHHe2kweg=",
           "requires": {
             "color-name": "1.1.3"
           }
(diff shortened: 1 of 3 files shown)


packages/webviz-core/src/panels/ImageView/decodings.js
```diff
@@ -114,3 +114,16 @@ export function decodeRGGB(rggb: Uint8Array, width: number, height: number, outp
     }
   }
 }
+
+export function decodeMono8(mono8: Uint8Array, width: number, height: number, output: Uint8ClampedArray) {
+  let inIdx = 0;
+  let outIdx = 0;
+
+  for (let i = 0; i < width * height; i++) {
+    const ch = mono8[inIdx++];
```
jtbandes: For `16UC1` / `mono16`, I think you would want `(data[inIdx++] << 8) | data[inIdx++]` (assuming big endian – not sure where this is documented). There should be twice as many values in the input data because it's 2 bytes per value instead of 1.

packages/webviz-core/src/panels/ImageView/ImageCanvas.js
```diff
@@ -89,6 +89,9 @@ export default class ImageCanvas extends React.Component<Props, State> {
           case "bgr8": decodeBGR(rawData, width, height, image.data); break;
           case "32FC1": decodeFloat1c(rawData, width, height, image.data); break;
           case "bayer_rggb8": decodeRGGB(rawData, width, height, image.data); break;
+          case "mono8": decodeMono8(rawData, width, height, image.data); break;
```
jtbandes: nit: can make it clear these two are treated the same by using a fallthrough:
```suggestion
          case "mono8":
```

packages/webviz-core/src/panels/ImageView/ImageCanvas.js
```diff
@@ -89,6 +89,9 @@ export default class ImageCanvas extends React.Component<Props, State> {
...
+          case "8UC1": decodeMono8(rawData, width, height, image.data); break;
```
jtbandes:
```suggestion
          case "8UC1":
            decodeMono8(rawData, width, height, image.data); break;
```

packages/webviz-core/src/panels/ImageView/decodings.js
jtbandes: actually, just noticed there's an `is_bigendian` on the image: http://docs.ros.org/melodic/api/sensor_msgs/html/msg/Image.html

jtbandes: you could also use DataView like we do for 32FC1

jxjshaw: does that mean we should support is_bigendian = false for all the encodings?

jtbandes: Yeah probably that would be ideal, for example this `true` should be `!is_bigendian`: https://github.com/cruise-automation/webviz/blob/f1d751aa8320e33f4a501a2e4b3a37851423a268/packages/webviz-core/src/panels/ImageView/decodings.js#L59

it doesn't affect the 8-bit encodings though.

packages/webviz-core/src/panels/ImageView/ImageCanvas.js
```diff
@@ -81,15 +81,22 @@ export default class ImageCanvas extends React.Component<Props, State> {
...
       // Binary message processing
       if (msg.datatype === "sensor_msgs/Image") {
-        const { width, height, encoding } = msg.message;
+        const { width, height, is_bigendian, encoding } = msg.message;
         image = new ImageData(width, height);
         // prettier-ignore
         switch (encoding) {
           case "yuv422": decodeYUV(rawData, width, height, image.data); break;
-          case "bgr8": decodeBGR(rawData, width, height, image.data); break;
+          case "bgr8": decodeBGR(rawData, width, height, is_bigendian, image.data); break;
```
jtbandes: not needed for bgr8 since it's only 8-bit

packages/webviz-core/src/panels/ImageView/decodings.js
```diff
@@ -51,12 +51,18 @@ export function decodeBGR(bgr: Uint8Array, width: number, height: number, output
...
+  height: number,
+  is_bigendian: boolean,
+  output: Uint8ClampedArray
+) {
   const view = new DataView(gray.buffer, gray.byteOffset);
 
   let outIdx = 0;
   for (let i = 0; i < width * height * 4; i += 4) {
-    const val = view.getFloat32(i, true) * 255;
+    const val = view.getFloat32(i, !is_bigendian ? true : false) * 255;
```
jtbandes:
```suggestion
    const val = view.getFloat32(i, !is_bigendian) * 255;
```

packages/webviz-core/src/panels/ImageView/decodings.js
```diff
@@ -114,3 +120,36 @@ export function decodeRGGB(rggb: Uint8Array, width: number, height: number, outp
...
+  width: number,
+  height: number,
+  is_bigendian: boolean,
+  output: Uint8ClampedArray
+) {
+  let inIdx = 0;
+  let outIdx = 0;
+
+  for (let i = 0; i < width * height; i++) {
+    const val = !is_bigendian ? mono16[inIdx] : mono16[inIdx - 1];
```
jtbandes: Did you try `... : mono16[inIdx + 1]` instead? I think `inIdx - 1` is wrong because at the beginning, it will be `mono16[-1]`.

packages/webviz-core/package-lock.json
```diff
@@ -15,27 +15,27 @@
         "regenerator-runtime": {
```
jtbandes: Would be nice to revert these changes since they aren't relevant to the PR.

jtbandes:
Merging PR to internal repo, will be out to open source repo in a future push

PR #166, opened 2019-07-01
Add DrawPolygons to Worldview

vidaaudrey:
## Summary
Add the `DrawPolygon` created by @brianc.  We have been using this command internally, and open sourcing it will help add polygon drawing capability to the 3D panel in Webviz. 

Note: use with caution as there might be major api changes before we completely open source it with examples and documentations.

## Test plan
Unit test and manual test.

## Versioning impact
Will publish a patch version now and a patch version later (after documentation is added).


packages/regl-worldview/src/commands/DrawPolygon/PolygonBuilder.js
```diff
@@ -0,0 +1,380 @@
...
+  return x1 === x2 && y1 === y2 && z1 === z2;
+}
+
+function isClosed(polygon: Polygon): boolean {
+  const { points } = polygon;
+  for (let i = 0; i < points.length - 1; i++) {
+    if (areEqual(points[i], points[i + 1])) {
+      return true;
+    }
+  }
```
janpaul123: Huh this for loop seems strange to me. If you create a polygon with two points that sit right on top of each other for some reason we consider it closed?

packages/regl-worldview/src/commands/DrawPolygon/PolygonBuilder.js
```diff
@@ -0,0 +1,380 @@
...
+  return areEqual(firstPoint, lastPoint);
+}
+
+// has listeners you can pass to Worldview for mouse interactions
+// internally builds a list of polygons and modifies the polygons
+// based on mouse & keyboard interactions.  for now we use mututation internally
+// instead of immutability to keep the number of allocations lower and make
+// the implementation a bit more straightforward
+//
+// for an example implementation check stories/Worldview/drawPolygon.js
```
janpaul123: Out of date?

packages/regl-worldview/src/commands/DrawPolygon/PolygonBuilder.js
```diff
@@ -0,0 +1,380 @@
...
+export default class PolygonBuilder {
+  mouseDown: boolean = false;
+  polygons: Polygon[];
+  onChange: () => void = () => {};
+  activePolygon: ?Polygon;
+  activePoint: ?PolygonPoint;
+  mouseDownPoint: Vec3;
```
janpaul123: Might be good to prefix private variables and methods with an underscore.

packages/regl-worldview/src/commands/DrawPolygon/PolygonBuilder.js
vidaaudrey: Not sure what was the original thought. I'll create a ticket for the after-work and add to the todo list.

packages/regl-worldview/src/commands/DrawPolygon/PolygonBuilder.js
vidaaudrey: `mouseDownPoint` and other instance variable seem to be useful for public api as well, e.g. trigger certain external events based mouseDown position

packages/regl-worldview/src/commands/DrawPolygon/PolygonBuilder.js
vidaaudrey: Good catch! Removed.

PR #175, opened 2019-07-16
Fix reading hitmap on every mouse move

davidswinegar:
## Summary

We had some incorrect logic that would cause the Worldview component
to read from the hitmap on every mouse move even if hitmapOnMouseMove
was false, as long as there was no `onMouseMove` handler. Fix it and
make the logic path a bit clearer.

## Test plan

Test manually with interactions on the demo page to make sure they aren't
broken.

## Versioning impact

Requires a patched version.

### DIFF
--- a/packages/regl-worldview/src/Worldview.js
+++ b/packages/regl-worldview/src/Worldview.js
@@ -161,7 +161,7 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
   };
 
   _onMouseMove = (e: MouseEvent) => {
-    this._onMouseInteraction(e, "onMouseMove", this.props.hitmapOnMouseMove);
+    this._onMouseInteraction(e, "onMouseMove");
   };
 
   _onMouseUp = (e: MouseEvent) => {
@@ -178,7 +178,7 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
     }
   };
 
-  _onMouseInteraction = (e: MouseEvent, mouseEventName: MouseEventEnum, readHitmap: boolean = true) => {
+  _onMouseInteraction = (e: MouseEvent, mouseEventName: MouseEventEnum) => {
     const { worldviewContext } = this.state;
     const worldviewHandler = this.props[mouseEventName];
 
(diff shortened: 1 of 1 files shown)


packages/regl-worldview/src/Worldview.js
```diff
@@ -196,8 +196,12 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
       return;
     }
 
-    if (!readHitmap && worldviewHandler) {
-      return handleWorldviewMouseInteraction(0, ray, e, worldviewHandler);
+    // rendering the hitmap on mouse move is expensive, so disable it by default
+    if (mouseEventName === "onMouseMove" && !this.props.hitmapOnMouseMove) {
```
esthersweon: Could you pull out into variable like `const renderHitMapOnMouseMove = mouseEventName === "onMouseMove" && this.props.hitmapOnMouseMove` and say `if (!renderHitmapOnMouseMove)`? Would that do what you want it to do?

packages/regl-worldview/src/Worldview.js
```diff
@@ -196,8 +196,12 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
...
+      if (worldviewHandler) {
+        return handleWorldviewMouseInteraction(0, ray, e, worldviewHandler);
+      }
+      return;
```
esthersweon: #nit for clearer logic path: 
```
if (!this.props.hitmapOnMouseMove) {
  if (mouseEventName === "onMouseMove") {
    return;
  }
  if (worldviewHandler) {
    return handleWorldviewMouseInteraction(0, ray, e, worldviewHandler);
  }
}

Comment (davidswinegar) [reply packages/regl-worldview/src/Worldview.js]: I don't think this is exactly the same logic. I could change it to:
```
    if (mouseEventName === "onMouseMove" && !this.props.hitmapOnMouseMove) {
      if (worldviewHandler) {
        return handleWorldviewMouseInteraction(0, ray, e, worldviewHandler);
      } else {
        return;
      }
    }
```
Would that help?

robin-pham: think the original is fine, this `else` is extraneous

PR #177, opened 2019-07-22
Redesign Worldview command interactivity API

vidaaudrey:
## Summary
Completely redesign the Worldview interactivity/mouse events API. 

New features:
* No longer have to manually manage `hitmapId`s: objects are returned by reference in event callbacks
* Optional support for stacked objects: mouse event callbacks can now return all objects under the mouse, instead of just the top-most object (off by default and currently hardcoded limit of 100 stacked objects for performance reasons)

Other API changes:
* Remove `makeCommand` in favor of simplifying the Command API
* Remove old methods for manually managing `hitmapId`s

Command syntax changes:
```
// Old
export const Triangle = makeCommand<Triangle>("Triangle", triangles);
// New
export function Triangles(props: { ...CommonCommandProps, children: TriangleList[] }) {
  return <Command getChildrenForHitmap={createInstancedGetChildrenForHitmap(3)} {...props} reglCommand={triangles} />;
}
```

Mouse event changes:
```
// Old
<Worldview 
  onClick={(event, { ray, objectId }) => {
    setSelectedObjectId(objectId);
  }}>
  ...
</Worldview>
// New
<Worldview 
  onClick={(event, { ray, objects }) => {
    setSelectedObjects(objects.map(({ object, instanceIndex }) => object));
  }}
  enableStackedObjectEvents>
  ...
</Worldview>
```

## Motivation
We want to support richer interactions with markers in the 3d panel in Webviz. Making these changes gives us a more scalable API for making those changes and unlocks stacked object support.

## Test plan
Unit tests are included for new modules.
Manual testing plan: visit every page in the docs and ensure the examples are working as expected.

## Release plan
This is a major API change that will require a Worldview version bump to 0.2.0.


packages/regl-worldview/src/Worldview.js
```diff
@@ -206,7 +213,7 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
       .readHitmap(canvasX, canvasY)
       .then((objectId) => {
         if (worldviewHandler) {
-          handleWorldviewMouseInteraction(objectId, ray, e, worldviewHandler);
+          handleWorldviewMouseInteraction(worldviewContext.getDrawPropByHitmapId(objectId), ray, e, worldviewHandler);
```
vidaaudrey: We can declare `const object = worldviewContext.getDrawPropByHitmapId(objectId)` here and use directly for `worldviewContext.callComponentHandlers(object)`. Also need to update Command's `handleMouseEvent`

packages/webviz-core/package-lock.json
```diff
@@ -19,11 +19,6 @@
         }
       }
     },
-    "@cruise-automation/hooks": {
```
vidaaudrey: Leave this commit out? The `webviz-core` directory is updated by mirroring the changes from our internal repo.

docs/src/3.3.Command.mdx
```diff
@@ -2,14 +2,22 @@ import CommandDemo from './jsx/api/Command.js';
...
+
+| Name          | optional | type                        | Default | Description                                                                 |
+| ------------- | -------- | --------------------------- | ------- | --------------------------------------------------------------------------- |
+| `reglCommand` | no       | `(Regl) => (Props) => void` |         | The command that takes the `regl` function and children and draws the shape |
+
+##### Properties generally passed through wrapping classes
+
+| Name         | optional | type               | Default | Description                                                                       |
+| ------------ | -------- | ------------------ | ------- | --------------------------------------------------------------------------------- |
+| `children`   | no       | `Array<any> | any` |         | Arbitrary properties passed into reglCommand.                                     |
```
janpaul123: The `Array<any> | any` bit doesn't get rendered properly by the Github markdown render, because it sees the `|` as a cell divider. Just make sure that in our docs this actually looks correct.

docs/src/3.4.MouseEvents.mdx
```diff
@@ -19,56 +20,124 @@ There are two kinds of event handlers:
 **Worldview Level Handlers**
 
 - `ray`: the raw raycasting information including `dir` (direction), `origin`, and `point` (all in `Vec3` format).
-- `objectId`: the id of the object that you interacted with. The value is `undefined` if no object is found.
+- `objects`: an array of objects (`{ object, instanceIndex }`) that you interacted with. The value is `[]` if no object is found. The `enableStackedObjectEvents` option must be set to `true` on `<Worldview` for this handler to return more than one object.
```
janpaul123:
```suggestion
- `objects`: an array of objects (`{ object, instanceIndex }`) that you interacted with. The value is `[]` if no object is found. The `enableStackedObjectEvents` option must be set to `true` on `<Worldview>` for this handler to return more than one object.
```

#nit

docs/src/3.4.MouseEvents.mdx
```diff
@@ -19,56 +20,124 @@ There are two kinds of event handlers:
...
-### Hitmap for Instanced Objects
+##### Turning off interactivity for a layer
 
-Usually using instanced objects can help improve rendering performance, e.g. an instanced sphere can contain thousands of spheres. When the user clicks a single sphere, the mapped instanceId (returned as `objectId`) which was generated through `getHitmapProps` function will be returned from the event.
+All built in commands allow overriding the `getHitmap` function, and we recommend that you also allow users to override it. If a user passes in `null` or `undefined` for `getHitmap` in a command, any objects drawn by that command instance will not be visible for interaction purposes. Disabling interaction in this way may help improve performance.
 
-_Commands that currently support instanced rendering: Cubes, Points, Spheres, Triangles, Cones, Cylinders. More to be added later._
+##### Customizing the getHitmap function
+
+Each Command is passed the `reglCommand` function and `children`. Worldview calls compiled version of the `reglCommand` function with `children`. `getHitmap` takes in those `children` and returns new version of them that will be rendered into the hitmap.
```
janpaul123:
```suggestion
Each Command is passed the `reglCommand` function and `children`. Worldview calls a compiled version of the `reglCommand` function with `children`. `getHitmap` takes in those `children` and returns new versions of them that will be rendered into the hitmap.
```

docs/src/3.4.MouseEvents.mdx
```diff
@@ -19,56 +20,124 @@ There are two kinds of event handlers:
...
+
+Example using composition of the default `mapHitmap` functions:
+
+```js
+<Worldview>
+  <Points
+    onClick={(evt, { objects }) => console.log(`You clicked on ${objects[0].object.id}`)}
+    mapHitmap={(props, assignNextIds) => {
+      const hitmapProps = nonInstancedGetHitmap(props, assignNextIds);
+      // Points are hard to click, so expand their hitmap so that clicking slightly outside of a point still selects it.
```
janpaul123: Love this example!!

packages/regl-worldview/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "regl-worldview",
-	"version": "0.1.2",
+	"version": "0.1.4",
```
janpaul123: Should we just do this version bump separately outside of this PR?

packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -71,15 +77,18 @@ function compile<T>(regl: any, cmd: RawCommand<T>): CompiledReglCommand<T> {
   return typeof src === "function" ? src : regl(src);
 }
 
+const MAX_NUMBER_OF_LAYERS = 100;
```
janpaul123: Maybe a comment to explain what this is for?

packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -240,63 +244,104 @@ export class WorldviewContext {
...
+          } while (currentObjectId !== 0);
+          resolve(seenIds);
         });
       });
     });
   }
 
-  callComponentHandlers = (objectId: number, ray: Ray, e: MouseEvent, mouseEventName: MouseEventEnum) => {
-    this._hitmapCalls.forEach((_, component) => {
+  callComponentHandlers = (hitmapIds: Array<HitmapId>, ray: Ray, e: MouseEvent, mouseEventName: MouseEventEnum) => {
```
janpaul123: It feels a bit roundabout for `<Worldview>` to call this. Why not call this internally from `readHitmap` (need to rename probably though)? Then we can probably find the appropriate `component` more easily anyway.

janpaul123: Not super big deal though, maybe the concerns are split kinda nicely now? 🤷‍♂

packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -240,63 +244,104 @@ export class WorldviewContext {
       regl({ framebuffer: _fbo })(() => {
         // clear the framebuffer
         regl.clear({ color: [0, 0, 0, 1], depth: 1 });
+        let currentObjectId = 0;
+        const seenObjects = [];
+        const seenIds: Array<HitmapId> = [];
```
jtbandes: should we call this type ObjectId or HitmapObjectId so it matches the "objects" terminology?

packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -240,63 +244,104 @@ export class WorldviewContext {
...
-  _drawInput = (isHitmap?: boolean) => {
-    const drawCallsMap = isHitmap ? this._hitmapCalls : this._drawCalls;
-    const sortedDrawCalls = Array.from(drawCallsMap.values()).sort((a, b) => (a.layerIndex || 0) - (b.layerIndex || 0));
+  _drawInput = (isHitmap?: boolean, seenObjects?: Array<MouseEventObject>) => {
+    if (isHitmap) {
+      this._hitmapIdManager.reset();
+    }
 
+    const drawCalls = Array.from(this._drawCalls.values());
+    const sortedDrawCalls = drawCalls.sort((a, b) => (a.layerIndex || 0) - (b.layerIndex || 0));
```
jtbandes: nit: while you are here, since sort is mutating, no need for 2 separate variables

packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -240,63 +244,104 @@ export class WorldviewContext {
...
       }
       const cmd = this._compiled.get(command);
       if (!cmd) {
-        return console.warn("could not find draw command for", instance.constructor.displayName);
+        return console.warn("could not find draw command for", instance ? instance.constructor.displayName : "Unknown");
+      }
+      // draw hitmap
+      if (isHitmap && getHitmap) {
+        const assignNextIdsFn: AssignNextIdsFn = (...rest) => {
+          return this._hitmapIdManager.assignNextIds(instance, ...rest);
```
jtbandes: nit: maybe just use `bind`

packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -240,63 +244,104 @@ export class WorldviewContext {
...
+                y,
+                width: 1,
+                height: 1,
+                data: pixel,
+              });
+
+              currentObjectId = getIdFromColor(pixel);
+              if (currentObjectId && this.getDrawPropByHitmapId(currentObjectId).object) {
+                seenIds.push(currentObjectId);
+                seenObjects.push(this.getDrawPropByHitmapId(currentObjectId));
```
jtbandes: maybe use a Set instead of Array for these?

packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -272,31 +275,51 @@ export class WorldviewContext {
...
+      }
+      if (getActive) {
+        const drawPropsWithActive = getActive(drawProps);
+        return cmd(drawPropsWithActive);
       }
       cmd(drawProps);
     });
   };
 
+  getDrawPropByHitmapId = (hitmapId: HitmapId): MouseEventObject => {
```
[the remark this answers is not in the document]
jtbandes: ♻️ It seems like this doesn't need to be a public function if we were to automatically call it in `readHitmap` to get the objects rather than just returning the ids?

packages/regl-worldview/src/commands/Command.js
```diff
@@ -9,45 +9,45 @@
...
-export type Props<T> = {
+type Props<T> = {
   [MouseEventEnum]: ComponentMouseHandler,
-  children?: T[],
-  getObjectFromHitmapId?: GetObjectFromHitmapId<HitmapProp<T>>,
-  hitmapProps?: HitmapProp<T>,
+  children: T[],
   layerIndex?: number,
   reglCommand: RawCommand<T>,
+  // Interactivity
```
jtbandes:
```suggestion
```

packages/regl-worldview/src/commands/Command.js
```diff
@@ -9,45 +9,45 @@
...
+  getHitmap?: GetHitmap,
```
jtbandes: Should use `...OptionalCommandProps` here. And maybe call those `CommonProps` or something?

packages/regl-worldview/src/commands/Command.js
```diff
@@ -9,45 +9,45 @@
...
   points?: Point[],
   color?: Color,
 };
-export type HitmapMarkerDefault = {
-  id?: number,
-  points?: Point[],
-  color?: Vec4,
+
+export type OptionalCommandProps = {
+  [MouseEventEnum]: ComponentMouseHandler,
```
jtbandes: this one doesn't look optional?

packages/regl-worldview/src/commands/Command.js
```diff
@@ -84,46 +84,26 @@ export default class Command<T> extends React.Component<Props<T>> {
...
+    const { children, reglCommand, layerIndex, getHitmap } = this.props;
+    if (children == null) {
       return;
     }
     context.registerDrawCall({
       instance: this,
-      command: this.props.reglCommand,
-      drawProps,
-      layerIndex: this.props.layerIndex,
+      command: reglCommand,
```
jtbandes: nit: maybe rename this to reglCommand since all the others are keeping their original names too

packages/regl-worldview/src/commands/Cones.js
```diff
@@ -6,19 +6,17 @@
...
 const cones = fromGeometry(points, sideFaces.concat(endCapFaces));
 
-const Cylinders = makeCommand<BaseShape>("Cylinders", cones, {
-  getHitmapProps,
-  getObjectFromHitmapId,
-});
-
-export default Cylinders;
+export default function Cones(props: { ...OptionalCommandProps, children: Array<BaseShape> }) {
+  return <Command getHitmap={createInstancedGetHitmap({ pointCountPerInstance: 1 })} {...props} reglCommand={cones} />;
```
jtbandes: nit: ordering
```suggestion
  return <Command reglCommand={cones} getHitmap={createInstancedGetHitmap({ pointCountPerInstance: 1 })} {...props} />;
```

packages/regl-worldview/src/commands/DrawPolygon/index.js
```diff
@@ -72,11 +72,19 @@ type Props = {
   children: DrawPolygonType[],
 };
 
+const polygonLinesGetHitmap: GetHitmap = <T>(props: T, assignNextIds, alreadySeenObjects) => {
```
jtbandes: nit: consistent naming throughout please
```suggestion
const polygonLinesGetHitmap: GetHitmap = <T>(props: T, assignNextIds, seenObjects) => {
```

packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -141,12 +142,15 @@ const drawModel = (regl) => {
       for (const primitive of mesh.primitives) {
         const material = model.json.materials[primitive.material];
         const texInfo = material.pbrMetallicRoughness.baseColorTexture;
+        if (!accessors || !textures) {
+          throw new Error("Error decoding GLB file");
```
jtbandes: curious why include these changes in this PR? when did you run into this error?

packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -267,9 +271,17 @@ function useModel(model: string | (() => Promise<Model>)): ?Model {
   );
 }
 
+// Override the default mapHitmap with our own implementation.
+const getHitmap: GetHitmap = <T>(prop: T, assignNextIds, alreadySeenObjects) => {
+  const [hitmapProp] = nonInstancedGetHitmap([prop], assignNextIds, alreadySeenObjects);
```
jtbandes: Can you just use the non-array version here?
```suggestion
  const hitmapProp = nonInstancedGetHitmap(prop, assignNextIds, alreadySeenObjects);
```

packages/regl-worldview/src/commands/Grid.js
```diff
@@ -58,13 +59,10 @@ type Props = {
...
-  };
 
-  render() {
-    return <Command reglCommand={grid} drawProps={this.props} />;
-  }
+export default function Grid(props: Props) {
+  return <Command getHitmap={nonInstancedGetHitmap} {...props} reglCommand={grid} />;
 }
+
+Grid.displayName = "Grid";
```
jtbandes: shouldn't be necessary since the function itself is named

packages/regl-worldview/src/types/index.js
```diff
@@ -174,11 +164,11 @@ export type Cube = BaseShape & {
 export type Cylinder = BaseShape;
 
 export type Line = BaseShape & {
-  points: (Point | Vec3)[],
+  points: Point[] | Vec3[],
 };
 
 export type PointType = BaseShape & {
-  points: (Point | Vec3)[],
+  points: (Point[] | Vec3)[],
```
jtbandes: ⚠️ Seems like a typo? Wasn't this supposed to be the same as Line? Bonus points: why didn't you get any Flow errors as a result?
```suggestion
  points: Point[] | Vec3[],
```

packages/regl-worldview/src/types/index.js
```diff
@@ -194,6 +184,23 @@ export type PolygonType = BaseShape & {
...
+
+export type HitmapId = number;
+/*
+ * type: "instanced" if mapping multiple IDs to a single callback object, "single" if 1:1.
+ * callbackObject: the object to pass to event callbacks when this object is interacted with.
+ * count: If "instanced", how many IDs to map to the callback object.
+ * return type: an array of the IDs assigned.
+ */
+export type AssignNextIdsFn = (
+  { type: "single", callbackObject: BaseShape } | { type: "instanced", callbackObject: BaseShape, count: number }
```
jtbandes: 📜

- Strange to have "callback" in the name; why not just call it `object` since that is the terminology we use elsewhere?

- Is there any need for the object to be BaseShape (doesn't seem like it's really used inside the HitmapIdManager)? Are Commands allowed to have children that don't derive from BaseShape?

- Maybe make this into a new type so you can use it again in HitmapIdManager.js

- Is there actually anything related to instancing here, or should we just rename it `"multiple"`?

  - Do we really need a `type` or could we instead just rely on `count === 1`?


- What if we also/instead return colors from this function? Consumers are pretty much always going to be converting this into a color with intToRGB, right?

packages/regl-worldview/src/utils/HitmapIdManager.js
```diff
@@ -0,0 +1,74 @@
...
+
+import type { HitmapId, MouseEventObject, BaseShape } from "../types";
+
+function fillArray(start: number, length: number): Array<number> {
+  return new Array(length).fill(0).map((_, index) => start + index);
+}
+
+type CommandInstance = React.Component<any>;
+
+export default class HitmapIdManager {
```
jtbandes: :memo: comment

packages/regl-worldview/src/utils/HitmapIdManager.js
```diff
@@ -0,0 +1,74 @@
...
+  _hitmapIdMap: { [HitmapId]: BaseShape } = {}; // map hitmapId to the original marker object
```
jtbandes: Can make the code self-documenting here
```suggestion
  _objectsByHitmapId: { [HitmapId]: BaseShape } = {};
```

packages/regl-worldview/src/utils/getHitmapDefaults.js
```diff
@@ -0,0 +1,89 @@
...
+  const hitmapArray: Array<T> = propsArray
+    .map((prop) => {
+      if (alreadySeenObjects.some(({ object }) => object === prop)) {
+        return null;
+      }
+      const hitmapProp = { ...prop };
+      const [id] = assignNextIds({ type: "single", callbackObject: prop });
+      const hitmapColor = intToRGB(id);
+      hitmapProp.color = hitmapColor;
+      if (hitmapProp.points) {
```
jtbandes: should we check points.length > 0? I don't remember what the commands do in the case of `point: foo,  points: []`

packages/regl-worldview/src/commands/Cylinders.js
```diff
@@ -47,9 +49,8 @@ const { points, sideFaces, endCapFaces } = createCylinderGeometry(30, false);
...
 
-const Cylinders = makeCommand<Cylinder>("Cylinders", cylinders, {
-  getHitmapProps,
-  getObjectFromHitmapId,
-});
-
-export default Cylinders;
+export default function Cylinders(props: { ...OptionalCommandProps, children: Array<Cylinder> }) {
+  return (
+    <Command getHitmap={createInstancedGetHitmap({ pointCountPerInstance: 1 })} {...props} reglCommand={cylinders} />
```
jtbandes: 🛑 Is this actually right? Each instance has `sideFaces.length + endCapFaces.length` elements (or 3 * that number of points). I don't see how that's being accounted for.

Screenshot-based tests for the hitmap logic would help with issues like this.

packages/regl-worldview/src/utils/getHitmapDefaults.js
```diff
@@ -0,0 +1,89 @@
...
+
+export const createInstancedGetHitmap = ({ pointCountPerInstance }: { pointCountPerInstance: number }) => <T: Object>(
+  props: Array<T> | T,
+  assignNextIds: AssignNextIdsFn,
+  alreadySeenObjects: Array<MouseEventObject>
+) => {
+  const propsArray = Array.isArray(props) ? props : [props];
+  const hitmapArray: Array<T> = propsArray
+    .map((prop: T) => {
+      const filteredIndicies = alreadySeenObjects
```
jtbandes:
```suggestion
      const filteredIndices = alreadySeenObjects
```

packages/regl-worldview/src/utils/getHitmapDefaults.js
```diff
@@ -0,0 +1,89 @@
...
+//  found in the LICENSE file in the root directory of this source tree.
+//  You may not use this file except in compliance with the License.
+
+import type { AssignNextIdsFn, MouseEventObject } from "../types";
+import { intToRGB } from "./commandUtils";
+
+export const nonInstancedGetHitmap = <T: Object>(
+  props: Array<T> | T,
+  assignNextIds: AssignNextIdsFn,
+  alreadySeenObjects: Array<MouseEventObject>
```
jtbandes: 📝 I think it would make more sense to call this `excludedObjects` so it's clear why we care about having "seen" them.

packages/regl-worldview/src/utils/getHitmapDefaults.js
```diff
@@ -0,0 +1,89 @@
...
+    })
+    .filter(Boolean);
+
+  if (Array.isArray(props)) {
+    return hitmapArray;
+  }
+  return hitmapArray[0];
+};
+
+export const createInstancedGetHitmap = ({ pointCountPerInstance }: { pointCountPerInstance: number }) => <T: Object>(
```
jtbandes: since there is only one param, maybe don't bother with the object

docs/src/3.4.MouseEvents.mdx
```diff
@@ -4,13 +4,14 @@ import { MouseEvents, MouseEventsInstanced } from './jsx/allLiveEditors';
...
+| `onMouseDown`   | `MouseHandler`    |         |                                                                                                                                                                          |
+| `onMouseUp`     | `MouseHandler`    |         |                                                                                                                                                                          |
+| `onMouseMove`   | `MouseHandler`    |         | only available if the `hitmapOnMouseMove` prop is set to `true` on the Worldview component                                                                               |
+| `onClick`       | `MouseHandler`    |         |                                                                                                                                                                          |
+| `getHitmap`     | `AssignNextIdsFn` |         | Maps the `children` to props to be drawn on the hitmap. Optional, but required for any mouse interactions. Provided by default by all included Commands. See more below. |
```
jtbandes: 📝 Naming: usually in our terminology the "hitmap" is the whole offscreen canvas, so calling this "getHitmap" is a bit weird. Maybe `getChildrenForHitmap` or something that relates it more to the original object/children?

docs/src/3.4.MouseEvents.mdx
```diff
@@ -19,56 +20,124 @@ There are two kinds of event handlers:
...
+ * callbackObject: the object to pass to event callbacks when this object is interacted with.
+ * count: If "instanced", how many IDs to map to the callback object.
+ * return type: an array of the IDs assigned.
+ */
+type AssignNextIdsFn = (
+  { type: "single", callbackObject: BaseShape } | { type: "instanced", callbackObject: BaseShape, count: number }
+) => Array<HitmapId>;
+```
+ * Each unique object that you want to event on should get a unique ID. If you want to map parts of a single object to multiple IDs (instancing), use the `type: "instanced"` call to `assignNextIds`.
+ * Once you have this ID, pass it to `intToRGB()` to get the color to paint your object. Don't apply any transparency or shading to your object or this will mess up the hitmap.
```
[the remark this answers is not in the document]
davidswinegar: we could pass the next color instead of the next id. Passing it into the fragment shader is an option but I had earlier opted not to do that because I thought it made the API more confusing and less explicit.

packages/regl-worldview/package-lock.json
davidswinegar: probably, since it should be 0.1.4 in master

packages/regl-worldview/src/WorldviewContext.js
davidswinegar: I think `HitmapObjectId` makes sense because it is an object in the hitmap

packages/regl-worldview/src/WorldviewContext.js
davidswinegar: Bind unfortunately caused some weird flow errors, so I went with this

packages/regl-worldview/src/commands/Command.js
davidswinegar: Because of the `[]` these mouse events do become optional

packages/regl-worldview/src/commands/GLTFScene.js
davidswinegar: I didn't run into an issue but this was caught when fixing flow errors in this file. Adding this error allows us to guarantee `accessors` and `textures` are defined.

packages/regl-worldview/src/commands/Cones.js
davidswinegar: We don't want users of this command to overwrite the reglCommand ever though, do we? I thought that was the one prop we don't want to allow

jtbandes: Ah gotcha, yeah that makes sense. I think I suggested this because it matches the order of props in other places we use Command. Not a big deal

packages/regl-worldview/src/utils/getHitmapDefaults.js
davidswinegar: I'm pretty sure they just create a single point at the default position, but I'll add this line

packages/regl-worldview/src/commands/Cylinders.js
davidswinegar: I am not sure what is correct but I can check.

docs/src/3.4.MouseEvents.mdx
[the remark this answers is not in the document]
vidaaudrey: I love @jtbandes's suggestion.  I wonder if we can have a general api for reglCommand: `hitmapColor`. For some of the complex reglCommand (e.g. MapTiles, GLTFScene, VirtualTileMaps), we need to pass down `isHitmap` anyhow to distinguish if we are drawing regular objects or hitmap objects.  This will also simplify hitmapProp generation since we can just use the original `drawProps`.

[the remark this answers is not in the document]
vidaaudrey: On a second thought, we can't just use the original `drawProps` for `hitmapProps`, because we might need to do something special such as changing `scale` when drawing hitmap objects, and these kind of changes are normally specified at the consumer side.  So agree with @davidswinegar, we should not pass the color(s) to the fragment shader.

docs/src/3.4.MouseEvents.mdx
davidswinegar: I think that makes sense

packages/regl-worldview/src/types/index.js
davidswinegar: We don't really have any usage of these components in tests, so we don't have any flow assertions. That's something I'll have to fix

packages/regl-worldview/src/utils/getChildrenForHitmapDefaults.test.js
```diff
@@ -0,0 +1,226 @@
...
+        object,
+        assignNextIds,
+        toexcludedObjects([object], [0])
+      );
+      expect(hitmapProps).toEqual(undefined);
+    });
+
+    it("handles arrays correctly", () => {
+      const objects = [
+        { some: "garbage", points: fillArray(6), colors: fillArray(6) },
```
janpaul123: 😂 🗑

packages/regl-worldview/src/WorldviewContext.js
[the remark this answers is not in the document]
davidswinegar: Refactored this to be a bit more understandable

packages/regl-worldview/src/WorldviewContext.js
davidswinegar: Did this in a different way

packages/regl-worldview/src/WorldviewContext.js
davidswinegar: Unfortunately we can't use a set, because we would need access to the original mouse event object. Instead we're reading from the sub-object inside of it and the instance index separately, and there isn't really an elegant way to get around this. The total number of seen objects should be <10 in the common case anyways

packages/regl-worldview/src/types/index.js
davidswinegar: * We could just call it object
* We currently don't support it in our typing system because of these interaction changes, but I could change that
* This type is almost identical to that type, but can't be used exactly because it's wrapped. Sharing a type signature would be a huge pain in Flow
* Kind of? The main reason you'd assign multiple IDs to the same object is for instancing. I chose to make it explicit based on the name but I could change it to be based on the count.
* We could also return colors with this function

packages/regl-worldview/src/commands/Cylinders.js
davidswinegar: For both cones and cylinders each point is just the centerpoint, so this is correct. Added some examples to make sure.

packages/regl-worldview/src/commands/Arrows.js
```diff
@@ -7,17 +7,50 @@
...
+  children: Arrow[],
+  ...CommonCommandProps,
+};
+
+export default class Arrows extends React.PureComponent<Props> {
+  constructor(props: Props) {
+    super(props);
+
+    // Wrap mouse events so that any cylinder/cone events are propagated correctly.
+    for (const eventName of SUPPORTED_MOUSE_EVENTS) {
```
vidaaudrey: Nice. I imagine we'd use this mapping logic a lot for composite components. Let's extract it to a util function and export it? Also would be nice to convert `Arrows` to a function.

docs/src/3.4.MouseEvents.mdx
```diff
@@ -19,56 +20,130 @@ There are two kinds of event handlers:
...
+<Worldview>
+  <Points
+    onClick={(evt, { objects }) => console.log(`You clicked on ${objects[0].object.id}`)}
+    getChildrenForHitmap={(props, assignNextColors, excludedObjects) => {
+      // This is a simplified version of nonInstancedGetChildrenForHitmap, for explanation purposes.
+      return propsArray.map((prop) => {
+        // Don't render objects that have been excluded.
+        // Objects are excluded when we enable stacked object events. If this object has already been evented on, we
+        // want to event on the objects underneath it, so don't allow drawing it.
+        if (excludedObjects.find({ object }) => object === prop)) {
```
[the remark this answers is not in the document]
davidswinegar: Unfortunately we can't because we're looking at an object inside, not the top-level object. There are always < 100 objects

docs/src/jsx/examples/Hitmap.js
```diff
@@ -139,15 +138,17 @@ function Example() {
       onCameraStateChange={(cameraState) => {
         setCameraState(cameraState);
       }}
-      onClick={(e, arg) => {
-        if (clickedId === arg.objectId) {
-          setClickedId(undefined);
-        } else {
-          setClickedId(arg.objectId);
+      onClick={(e, { objects }) => {
+        if (objects[0]) {
```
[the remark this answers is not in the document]
davidswinegar: Any onClick on a Command (not the top-level worldview) is guaranteed to have at least one object, otherwise that event wouldn't trigger at all

packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -239,61 +245,93 @@ export class WorldviewContext {
       // tell regl to use a framebuffer for this render
       regl({ framebuffer: _fbo })(() => {
         // clear the framebuffer
-        regl.clear({ color: [0, 0, 0, 1], depth: 1 });
+        regl.clear({ color: intToRGB(0), depth: 1 });
+        let currentObjectId = 0;
+        const excludedObjects = [];
```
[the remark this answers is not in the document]
davidswinegar: Explained above why we can't do this

packages/regl-worldview/src/Worldview.js
```diff
@@ -199,20 +202,26 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
...
-      .readHitmap(canvasX, canvasY)
-      .then((objectId) => {
+      .readHitmap(canvasX, canvasY, !!this.props.enableStackedObjectEvents)
+      .then((mouseEventsWithCommands) => {
         if (worldviewHandler) {
-          handleWorldviewMouseInteraction(objectId, ray, e, worldviewHandler);
+          const mouseEvents = mouseEventsWithCommands.map(([mouseEventObject]) => mouseEventObject);
+          handleWorldviewMouseInteraction(mouseEvents, ray, e, worldviewHandler);
+        }
+        const mouseEventsByCommand: Map<CommandComponentInstance, Array<MouseEventObject>> = aggregate(
```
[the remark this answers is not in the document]
davidswinegar: We removed that and instead have Worldview call commands directly. The context layer wasn't doing a whole lot other than passing through calls so it didn't make much sense to add another layer

packages/regl-worldview/package-lock.json
davidswinegar: Done in #190

packages/regl-worldview/src/commands/Arrows.js
davidswinegar: Changed this to instead write a new version of `getChildrenForHitmap`

docs/src/4.13.DrawPolygons.mdx
```diff
@@ -0,0 +1,13 @@
...
+
+# DrawPolygons
+
+`<DrawPolygons />`, combined with `PolygonBuilder`, allows drawing polygons on the X-Y plane. The example below shows how to combine the two to create a drawing tool. This tool works best when the user's camera is in non-perspective mode (2D).
+
+## Props
+
+| Name       | Type     | Default | Description                     |
+| ---------- | -------- | ------- | ------------------------------- |
+| `children` | `[]Polygon` | []   | Polygon class instances from DrawPolygons. |
```
jtbandes:
```suggestion
| `children` | `Polygon[]` | []   | Polygon class instances from DrawPolygons. |
```

packages/regl-worldview/src/commands/Arrows.js
```diff
@@ -7,87 +7,104 @@
...
+      pose: {
+        position: vec3ToPoint(headPosition),
+        orientation: vec4ToOrientation(orientation),
+      },
+    });
   }
+
+  return (
+    <Fragment>
+      <Cylinders getChildrenForHitmap={getChildrenForHitmapWithOriginalMarker} {...passedProps}>
```
jtbandes: Thinking about alternative APIs for this. Some ideas:
- `getChildrenForHitmap={makeGetChildrenForHitmap(originalMarker)}` (don't have to mutate the marker itself, although I guess the child needs to pass it down somehow — could do this by keeping a map from draw call to original marker)
- what if we override the original marker with a context, such as `return <Worldview.OriginalMarkerProvider value={originalMarker}> <Cylinders> {... as normal ... } </Cylinders> ... </Worldview.OriginalMarkerProvider>`

packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -197,8 +201,8 @@ const drawModel = (regl) => {
           props.scale ? pointToVec3(props.scale) : [1, 1, 1]
         ),
       globalAlpha: (context, props) => (props.alpha == null ? 1 : props.alpha),
-      hitmapColor: (context, props) => intToRGB(props.id),
-      drawHitmap: (context, props) => props.id != null,
+      hitmapColor: (context, props) => props.color || [0, 0, 0, 1],
+      drawHitmap: (context, props) => !!props.isHitmap,
```
jtbandes: nit: rename this to isHitmap?

packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -267,9 +271,17 @@ function useModel(model: string | (() => Promise<Model>)): ?Model {
   );
 }
 
+// Override the default getChildrenForHitmap with our own implementation.
+const getChildrenForHitmap: GetChildrenForHitmap = <T: any>(prop: T, assignNextColors, excludedObjects) => {
+  const hitmapProp = nonInstancedGetChildrenForHitmap(prop, assignNextColors, excludedObjects);
+  if (hitmapProp) {
+    hitmapProp.isHitmap = true;
```
jtbandes: Probably better to use `return {...hitmapProp, isHitmap: true}`?

packages/regl-worldview/src/types/index.js
```diff
@@ -187,13 +182,30 @@ export type SphereList = BaseShape & {
...
+  object: BaseShape,
+  instanceIndex: ?number,
+};
+
+export type ObjectHitmapId = number;
+/*
+ * type: "instanced" if mapping multiple IDs to a single callback object, "single" if 1:1.
+ * object: the object to pass to event callbacks when this object is interacted with.
+ * count: If "instanced", how many IDs to map to the callback object.
+ * return type: an array of the IDs assigned.
```
jtbandes: comment out of date: seems to return colors now

packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -239,61 +242,110 @@ export class WorldviewContext {
...
-      }
-    });
-  };
-
-  _drawInput = (isHitmap?: boolean) => {
-    const drawCallsMap = isHitmap ? this._hitmapCalls : this._drawCalls;
-    const sortedDrawCalls = Array.from(drawCallsMap.values()).sort((a, b) => (a.layerIndex || 0) - (b.layerIndex || 0));
+  _drawInput = (isHitmap?: boolean, excludedObjects?: MouseEventObject[]) => {
+    if (isHitmap) {
+      this._hitmapObjectIdManager.reset();
```
jtbandes: Maybe just create a new manager object each time? otherwise there's some risk of bugs if we add some behavior and forget to clear it in reset().

packages/regl-worldview/src/utils/HitmapObjectIdManager.js
```diff
@@ -0,0 +1,75 @@
...
+  return new Array(length).fill(0).map((_, index) => start + index);
+}
+
+/*
+ * This object manages the mapping between objects that are rendered into the scene and their IDs.
+ * It supplies an API for generating IDs for a rendered object and then accessing those objects based on their ID.
+ */
+export default class HitmapObjectIdManager {
+  _objectsByObjectHitmapIdMap: { [ObjectHitmapId]: Object } = {};
+  _objectToCommandMap: Map<Object, Command> = new Map();
```
jtbandes: naming nit: you have `_foosByBarMap` above, and `_barToFooMap` here. Should change these to be consistent — personally I'd advocate for just `_foosByBar`, but it's up to you, just want to see them be consistent

packages/regl-worldview/src/utils/aggregate.js
```diff
@@ -0,0 +1,31 @@
...
+
+//  Copyright (c) 2018-present, GM Cruise LLC
+//
+//  This source code is licensed under the Apache License, Version 2.0,
+//  found in the LICENSE file in the root directory of this source tree.
+//  You may not use this file except in compliance with the License.
+
+// Takes an array of [value, key] and aggregates across the keys. Results in a Map of [key, values[]], in order of the
+// keys as seen in the array.
+export default function aggregate<T, K>(array: Array<[T, K]>): Map<K, T[]> {
```
jtbandes: Isn't this pretty much lodash `groupBy()`?

packages/regl-worldview/src/utils/aggregate.js
```diff
@@ -0,0 +1,31 @@
...
+  const aggregationMap = new Map<K, T[]>();
+  array.forEach(([item, key]) => {
+    const existingItems = aggregationMap.get(key) || [];
+    existingItems.push(item);
+    aggregationMap.set(key, existingItems);
+  });
+  // Since later `set` calls overwrite the earlier ones in ordering, we have to use a new map to ensure that the order
```
jtbandes: Seems like you don't need to call `set()` every time, only if it's the first time you're adding a key, so you could avoid this second map.

packages/regl-worldview/src/utils/getChildrenForHitmapDefaults.test.js
```diff
@@ -0,0 +1,226 @@
...
+
+import type { MouseEventObject } from "../types";
+import { intToRGB } from "./commandUtils";
+import { nonInstancedGetChildrenForHitmap, createInstancedGetChildrenForHitmap } from "./getChildrenForHitmapDefaults";
+
+function fillArray(length: number) {
+  return new Array(length).fill(null).map(() => []);
+}
+
+function toexcludedObjects(objects: Object[], instanceIndicies?: Array<?number>): MouseEventObject[] {
```
jtbandes: caps: toExcluded? typo: instanceIndices

packages/regl-worldview/src/utils/aggregate.js
davidswinegar: It is, except groupBy does not allow arbitrary object keys (it returns a plain JS object)

PR #180, opened 2019-07-31
Add dependencies necessary for future Node Playground panel

esthersweon:
## Summary
- Add dependencies necessary for development of future Node Playground panel

### DIFF
--- a/packages/webviz-core/package.json
+++ b/packages/webviz-core/package.json
@@ -24,7 +24,7 @@
     "moment": "2.22.2",
     "moment-duration-format": "2.2.2",
     "moment-timezone": "0.5.23",
-    "monaco-editor": "^0.17.1",
+    "monaco-editor": "0.17.1",
     "natsort": "2.0.0",
     "nearley": "2.15.1",
     "promise-queue": "2.2.5",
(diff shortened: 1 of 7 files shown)


packages/webviz-core/package.json
```diff
@@ -65,7 +67,8 @@
     "eslint-plugin-react-hooks": "^1.5.0",
     "fake-indexeddb": "2.0.4",
     "fetch-mock": "7.2.5",
-    "flow-bin": "0.95.1"
+    "flow-bin": "0.95.1",
+    "monaco-editor-webpack-plugin": "^1.7.0"
```
troygibb: ^^ :hocho:  semantic versioning.

packages/webviz-core/package.json
```diff
@@ -24,6 +24,7 @@
     "moment": "2.22.2",
     "moment-duration-format": "2.2.2",
     "moment-timezone": "0.5.23",
+    "monaco-editor": "0.17.1",
```
janpaul123: Since we have this already in the global package.json, we should probably not also have it here, because then we might get multiple instances of this package, which can cause confusing issues.

PR #182, opened 2019-07-31
Add webviz related package dependencies for deterministic build

vidaaudrey:
## Summary
Problem: `webviz-core`  depends on the internal packages such as `regl-worldview` and `react-key-listerner`, and currently we run the CI build against the latest package src code, not on the actual npm packages. Ideally, we'd like to update any package, publish, and update `webviz-core` in separate PRs. 

Fix: add the packages as dependencies and modify webpack.config to not look for node modules in `/packages` directory, unless `USE_LOCAL_PACKAGES` env variable is set. 

## Test plan

CI should pass. 

## Versioning impact

No impact.


vidaaudrey:
Closing this as it's not worth the effort to change.

PR #194, opened 2019-08-19
Fix grid not rendering

davidswinegar:
## Summary

Because of the changes to hitmap rendering, grid rendering broke.
Fix it by updating the shader to handle new hitmapId mapping, and
also add a doc page for grids.

## Test plan

Added grid doc route, manually tested

## Versioning impact

May require a 0.2.1 package for fixes.


davidswinegar:
Yeah, I've started to research that today and will work on it in a separate PR

PR #196, opened 2019-08-19
Add engines to package.json

davidswinegar:
Add "engines" section because our build currently doesn't work with
node v8. Prints a warning when using the wrong node version.

Test plan: manual

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,10 @@
 {
   "name": "root",
   "private": true,
+  "engines": {
+    "node": ">=10.0.0 <12.0.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,10 @@
 {
   "name": "root",
   "private": true,
+  "engines": {
```
vidaaudrey: Should we share the settings for all package.json files just in case people are trying to build each one separately?

davidswinegar: I thought about it, but I'm afraid they'll all get out of date and all of our commands start in this package and use lerna anyways

PR #200, opened 2019-08-21
Add FOV and frusta to cameraState

nlittau-cruise:
## Summary

Add `fov`, `nearFrustum`, and `farFrustum` to `cameraState` so they can be customized.

## Test plan

Unit tests pass with no regressions; added the new properties to `CameraStore` tests. Manual inspection confirms that the properties behave as expected.

## Versioning impact

Minor update due to the backward-compatible new features.


docs/src/3.2.Camera.mdx
```diff
@@ -24,6 +24,9 @@ You can move the camera around using the mouse/trackpad, or by using the keyboar
 | `distance`          | `number`  | `75`           | distance between the camera and target                                                          |
 | `phi`               | `number`  | `Math.PI/4`    | angle in radians from +z pole to -z pole, range: [0, π]. Ignored when `perspective` is false    |
 | `thetaOffset`       | `number`  | `0`            | offset azimuth angle in radians, +θ corresponds to +x→+y, range: [0, 2π]                        |
+| `fov`               | `number`  | `Math.PI/4`    | field of view in radians                                                                        |
+| `nearFrustum`       | `number`  | `0.01`         | distance to the near frustum                                                                    |
```
jtbandes: Re naming: I think "frustum" refers to the whole viewable box, not the near/far values themselves. These are called near/far _planes_, but this is pretty standard, so maybe we can just call them `near` and `far`? Also it looks like `fovy` is the standard name for fov, with `y` indicating that it's in the vertical direction.

references:
https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/gluPerspective.xml (uses fovy/zNear/zFar)
https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glFrustum.xml (uses nearVal/farVal)
http://glmatrix.net/docs/module-mat4.html#.perspective (uses fovy/near/far)

nlittau-cruise: 👍️ Fixed

nlittau-cruise:
> Thanks! Could you also rebase with master since we just pushed massive changes, and add a screenshot to PR description for the update CameraStateControls UI.

@vidaaudrey Rebased and added a screenshot. Unfortunately, the sliders don't line up anymore, since 0 isn't a valid value for the near/far planes in a perspective projection. I couldn't think of an obvious way to fix that in CSS without changing the HTML structure, but perhaps someone else can weigh in.

vidaaudrey:
> @vidaaudrey Rebased and added a screenshot. Unfortunately, the sliders don't line up anymore, since 0 isn't a valid value for the near/far planes in a perspective projection. I couldn't think of an obvious way to fix that in CSS without changing the HTML structure, but perhaps someone else can weigh in.

Will take a look first thing tomorrow morning as I'm a little caught up today.

jtbandes:
We should also make sure the API will be backwards compatible and not break existing code that's unaware of these new fields :)

nlittau-cruise:
> We should also make sure the API will be backwards compatible and not break existing code that's unaware of these new fields :)

That's true. I think the way we're handling the default camera state could be improved—if I pass `{perspective: false}` as `cameraState`, I would expect the missing fields to be filled in from `DEFAULT_CAMERA_STATE`, but they're not. I would have to pass in `{...DEFAULT_CAMERA_STATE, perspective: false}`.

If the unit tests here are any indication, there are a lot of uses of Worldview that aren't speading `DEFAULT_CAMERA_STATE` for forward compatibility. What would you think about making this change?

PR #202, opened 2019-08-21
Add docs for migrating from 0.1.x to 0.2.x

davidswinegar:
## Summary

Add a new doc for migrating worldview from 0.1.x to 0.2.x.

## Test plan

Manually ran docs

## Versioning impact

None

### DIFF
--- a/docs/src/5.1.MigratingToVersion0.2.x.mdx
+++ b/docs/src/5.1.MigratingToVersion0.2.x.mdx
@@ -1,0 +1,91 @@
+# Migrating from 0.1.x to 0.2.x
+
+The 0.2.1 Worldview release brought a new API for managing interactions and a variety of changes to how interactions are done. This is a breaking change - code that worked under 0.1.x may silently fail under 0.2.x. Below is a summary of what changed and how to update your own code.
+
+## Mouse event changes
+
+Previously, the Worldview-level event handler only passed an `objectId`, not the full object being clicked. Now both the Worldview-level event handlers and Command-level event handlers pass the same information about the click: the ray, and an array of objects that contain the object being clicked and the `instanceIndex`, if the clicked object is part of an instanced command.
+
+We now pass an array of objects because stacked object events can now optionally be enabled: passing `enableStackedObjectEvents={true}` to the `Worldview` component allows seeing all objects at the point the user clicked on, instead of just the top-level object.
+```js
+// New type
+type ClickedObject = {
+  object: Object,
+  instanceIndex?: ?number,
+};
+
+export type ReglClickInfo = {
+  ray: Ray,
+  objects: Array<ClickedObject>,
+};
+
+export type MouseHandler = (MouseEvent, ReglClickInfo) => void;
+```
+
+Usage:
+
+```js
+// Old
+<Worldview
+  onClick={(event, { ray, objectId }) => {
+    setSelectedObjectId(objectId);
+  }}>
+  ...
+</Worldview>
+// New
+<Worldview
+  onClick={(event, { ray, objects }) => {
+    setSelectedObjects(objects.map(({ object, instanceIndex }) => object));
+  }}
+  enableStackedObjectEvents>
+  ...
+</Worldview>
+```
+
+## Hitmap changes
+
+How do we detect which object was interacted with? We use a process called picking, in which we generate an off-screen canvas, or "hitmap", to paint our objects into, assign each object a unique color, and then paint each object into this hitmap. We then detect which object was clicked on by testing which color was underneath the cursor.
+
+For stacked objects, we run this hitmap rendering multiple times. We note which object was picked, then exclude it from the next rendering cycle, until we stop hitting objects.
+
+Previously, the assignment of these colors (also referred to as `hitmapIds`) was passed into each Command. Now there is an API for drawing these colors in a more scalable way.
+
+This means that we no longer support passing the `getObjectFromHitmapId`, `getHitmapId`, `getHitmapProps`, or `hitmapProps` props into a Command instance.
+
+Instead we have a single function, `getChildrenForHitmap`. This function takes as arguments the children passed into the Command, another function that helps assign colors for the hitmap, and a list of objects that we want to exclude from rendering (used for stacked object hitmaps). It returns new children that we use for rendering to the hitmap.
+
+Similarly to the previous API, there are a few default `getChildrenForHitmap` functions available to you to use.
+
+```js
+// Old
+<Cubes getHitmapId={getHitmapId} getObjectFromHitmapId={getObjectFromHitmapId}>{cubes}</Cubes>
+
+// New
+<Cubes getChildrenForHitmap={createInstancedGetChildrenForHitmap(1)}>{cubes}</Cubes>
+```
+
+See [MouseEvents](#/docs/api/mouse-events) for further info on the new hitmapId API.
+
+## Command syntax changes
+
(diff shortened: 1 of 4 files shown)


docs/src/routes.js
```diff
@@ -115,6 +121,7 @@ export function getHashUrlByComponentName(name) {
     ROUTE_CONFIG.forEach(({ name, subRouteNames }) => {
       subRouteNames.forEach((subRouteName) => {
         const componentName = getComponentName(subRouteName);
+        console.log(componentName)
```
janpaul123: Stray console.log

docs/src/5.1.MigratingToVersion0.2.x.mdx
```diff
@@ -0,0 +1,91 @@
+# Migrating from 0.1.x to 0.2.x
+
+The 0.2.1 Worldview release brought a new API for managing interactions and a variety of changes to how interactions are done. This is a breaking change - code that worked under 0.1.x may silently fail under 0.2.x. Below is a summary of what changed and how to update your own code.
+
+## Mouse event changes
+
+Previously, the Worldview-level event handler only passed an `objectId`, not the full object being clicked. Now both the Worldview-level event handlers and Command-level event handlers pass the same information about the click: the ray, and an array of objects that contain the object being clicked and the `instanceIndex`, if the clicked object is part of an instanced command.
```
jtbandes:
```suggestion
Previously, the Worldview-level event handler was only passed an `objectId`, not the full object being clicked. Now both the Worldview-level event handlers and Command-level event handlers pass the same information about the click: the ray, and an array of objects that contain the object being clicked and the `instanceIndex`, if the clicked object is part of an instanced command.
```

docs/src/5.1.MigratingToVersion0.2.x.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+We now pass an array of objects because stacked object events can now optionally be enabled: passing `enableStackedObjectEvents={true}` to the `Worldview` component allows seeing all objects at the point the user clicked on, instead of just the top-level object.
```
jtbandes:
```suggestion
We now pass an array of objects because stacked object events can now optionally be enabled: passing `enableStackedObjectEvents={true}` to the `Worldview` component allows seeing all objects at the point the user clicked on, instead of just the top-level object. (Enabling this feature may reduce performance, as the scene is rendered repeatedly while objects are removed until nothing is left under the clicked point.)
```

docs/src/5.1.MigratingToVersion0.2.x.mdx
```diff
@@ -0,0 +1,91 @@
...
+    setSelectedObjects(objects.map(({ object, instanceIndex }) => object));
+  }}
+  enableStackedObjectEvents>
+  ...
+</Worldview>
+```
+
+## Hitmap changes
+
+How do we detect which object was interacted with? We use a process called picking, in which we generate an off-screen canvas, or "hitmap", to paint our objects into, assign each object a unique color, and then paint each object into this hitmap. We then detect which object was clicked on by testing which color was underneath the cursor.
```
jtbandes: Probably intentional, but just noting this duplicates content from the MouseEvents page. Would it be better to just link there so we don't risk them getting out of sync?

docs/src/5.1.MigratingToVersion0.2.x.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+For stacked objects, we run this hitmap rendering multiple times. We note which object was picked, then exclude it from the next rendering cycle, until we stop hitting objects.
+
+Previously, the assignment of these colors (also referred to as `hitmapIds`) was passed into each Command. Now there is an API for drawing these colors in a more scalable way.
```
jtbandes:
```suggestion
Previously, each Command was responsible for assigning a unique identifier to each object (also referred to as a `hitmapId`) and converting this to a color. Now there is a more batteries-included API for assigning these colors.
```

docs/src/5.1.MigratingToVersion0.2.x.mdx
vidaaudrey: Have you thought about using a different prop `maxStackedObjectCount` to replace `enableStackedObjectEvents` . e.g. default `maxStackedObjectCount` to 0, which is the equivalent of `enableStackedObjectEvents=false`

docs/src/5.1.MigratingToVersion0.2.x.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+```js
+// Old
+<Cubes getHitmapId={getHitmapId} getObjectFromHitmapId={getObjectFromHitmapId}>{cubes}</Cubes>
+
+// New
+<Cubes getChildrenForHitmap={createInstancedGetChildrenForHitmap(1)}>{cubes}</Cubes>
+```
+
+See [MouseEvents](#/docs/api/mouse-events) for further info on the new hitmapId API.
```
jtbandes: did you mean "the new `getChildrenForHitmap` API"?

docs/src/5.1.MigratingToVersion0.2.x.mdx
```diff
@@ -0,0 +1,91 @@
...
+
+## Command syntax changes
+
+Because the Command has been simplified, there wasn't any need for a `makeCommand` wrapper any longer. The `makeCommand` function has been removed. Simply wrap the `Command` in another React component that passes it's props through, while specifying the `getChildrenForHitmap` function and the `reglCommand`. It's recommended that the `getChildrenForHitmap` function be overrideable and the `reglCommand` not be overrideable through props - see [MouseEvents](#/docs/api/mouse-events) for more on this.
```
jtbandes:
```suggestion
Because the Command has been simplified, there wasn't any need for a `makeCommand` wrapper any longer. The `makeCommand` function has been removed. Simply wrap the `Command` in another React component that passes its props through, while specifying the `getChildrenForHitmap` function and the `reglCommand`. It's recommended that the `getChildrenForHitmap` function be overrideable and the `reglCommand` not be overrideable through props — see [MouseEvents](#/docs/api/mouse-events) for more on this.
```

docs/src/5.1.MigratingToVersion0.2.x.mdx
```diff
@@ -0,0 +1,91 @@
...
+// Old
+<Cubes getHitmapId={getHitmapId} getObjectFromHitmapId={getObjectFromHitmapId}>{cubes}</Cubes>
+
+// New
+<Cubes getChildrenForHitmap={createInstancedGetChildrenForHitmap(1)}>{cubes}</Cubes>
+```
+
+See [MouseEvents](#/docs/api/mouse-events) for further info on the new hitmapId API.
+
+## Command syntax changes
```
jtbandes:
```suggestion
## Command syntax changes and removal of `makeCommand()`
```

docs/src/5.1.MigratingToVersion0.2.x.mdx
vidaaudrey: Maybe mention the `objects` are the exact ones that been passed in as `children` or `originalMarker`

docs/src/5.1.MigratingToVersion0.2.x.mdx
davidswinegar: Yeah, I didn't get around to merging the two concepts. Even though this one is more verbose, it might be more understandable? not sure.

jtbandes: What about just calling it `maxObjectCount` to make it clear it's related to the `objects` that you get back in an event handler? Then it make sense to have it default to 1.

Issue #197, opened 2019-08-19
Issues with decoding .glb files


spjy:
I am attempting to load .glb files other than the Duck and CesiumMan that are provided in the [GLTFScene example](https://webviz.io/worldview/#/docs/commands/gltfscene). I have tried to import my own models as well as ones provided in the example model repository [here](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0) such as `Box.glb` and `RiggedSimple.glb`, but I always get the error "Error decoding GLB file".

I have forked your GLTFScene example and tried importing my own model to it, namely `rectangle.glb` and I get the same error as above. All the models are located in `/utils/Models`.

Example: https://codesandbox.io/s/agitated-star-ruxr4?fontsize=14

jtbandes:
Thanks for the report. This looks like a regression caused by #198; we'll look into it soon. cc @davidswinegar

jtbandes:
Tested out your CodeSandbox again and it looks like it's working now, after David's fix:

![image](https://user-images.githubusercontent.com/14237/64647684-4f612100-d3ce-11e9-9378-4cd7f094b73e.png)

PR #203, opened 2019-08-22
Fix GLTF regression

davidswinegar:
## Summary
Fix GLTFScene regression causing bug report. We unnecessarily
checked for an attribute that wasn't needed for rendering.

## Test plan
Manually checked that GLTFScene works with a file that previously failed.

## Versioning impact
This will require a minor version bump.

### DIFF
--- a/packages/regl-worldview/src/commands/GLTFScene.js
+++ b/packages/regl-worldview/src/commands/GLTFScene.js
@@ -116,13 +116,12 @@ const drawModel = (regl) => {
   // build the draw calls needed to draw the model. This will happen whenever the model changes.
   const getDrawCalls = memoizeWeak((model: GLBModel) => {
     // upload textures to the GPU
-    const { images, accessors } = model;
+    const { accessors } = model;
     const textures =
       model.json.textures &&
-      images &&
       model.json.textures.map((textureInfo) => {
         const sampler = model.json.samplers[textureInfo.sampler];
-        const bitmap: ImageBitmap = images[textureInfo.source];
+        const bitmap = model.images && model.images[textureInfo.source];
         const texture = regl.texture({
           data: bitmap,
           min: glConstantToRegl(sampler.minFilter),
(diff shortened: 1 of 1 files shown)


davidswinegar:
Fixes #197

packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -116,13 +116,12 @@ const drawModel = (regl) => {
...
     // upload textures to the GPU
-    const { images, accessors } = model;
+    const { accessors } = model;
     const textures =
       model.json.textures &&
-      images &&
       model.json.textures.map((textureInfo) => {
         const sampler = model.json.samplers[textureInfo.sampler];
-        const bitmap: ImageBitmap = images[textureInfo.source];
+        const bitmap = model.images && model.images[textureInfo.source];
```
jtbandes: Is this just a revert or why is it changing? (I don't remember whether `images` always exists; I think it does but might be empty?)

packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -142,7 +141,7 @@ const drawModel = (regl) => {
       for (const primitive of mesh.primitives) {
         const material = model.json.materials[primitive.material];
         const texInfo = material.pbrMetallicRoughness.baseColorTexture;
-        if (!accessors || !textures) {
+        if (!accessors) {
           throw new Error("Error decoding GLB file");
```
jtbandes: I think we could just remove this error entirely. The error message will be more obvious/debuggable if it says `cannot access property [x] of undefined` rather than just "Error decoding GLB". unless you want to reword the error message

packages/regl-worldview/src/commands/GLTFScene.js
davidswinegar: this is kind of a revert - I just added the `model.images &&` to ensure that there wasn't a flow error. `images` does not always exist.

packages/regl-worldview/src/commands/GLTFScene.js
davidswinegar: I'll reword it, I wrote it this way is so that we can maintain flow typing

PR #212, opened 2019-08-26
Add Worldview interactivity integration tests

davidswinegar:
## Summary

Add a bunch of integration tests for worldview interactivity. These integration tests take the form of Storybook stories. The assertions run in the browser immediately upon loading the page and, if they fail, display the stack trace instead of the component being rendered. We test these in CI using our screenshot tests, so we will know if one of the tests have failed.

Tests cover the `onClick` handler for each component and the worldview-level `onClick`, and test the most common cases: clicking an object, clicking an object when `enableStackedObjectEvents` is enabled, and instanced clicking if applicable. There are also basic tests for other events.

This testing enabled discovery and fixing of the following bugs, mostly involving errors around stacked object events:
* Sometimes we would complain that the `getChildrenFromHitmap` function was not respecting `excludedObjects` in stacked object hitmaps when it really was
* Ensure that we print errors when we encounter a bitmap color that we don’t know about
* GLTFScene: did not return the exact marker passed into children when evented
* Grid: could not be clicked, incorrect shader code
* `getChildrenForHitmapWithOriginalMarker` would not respect `excludedObjects` in stacked object hitmaps at all
* `createInstancedGetChildrenForHitmap` would not respect `excludedObjects` in stacked object hitmaps in many cases

## Test plan

These are tests. There is also a deliberately "failing" screenshot test and jest tests for our internally built expect matchers.

## Versioning impact

This will require bumping the minor version of the package.


webpack.config.js
```diff
@@ -63,6 +63,9 @@ module.exports = {
     devtoolModuleFilenameTemplate: (info) => path.resolve(info.absoluteResourcePath),
   },
   resolve: {
+    alias: {
+      common: path.resolve(__dirname, "common"),
+    },
```
janpaul123: Hm, why is this necessary when we have `modules: [path.resolve(".")` below?

.flowconfig
```diff
@@ -19,6 +19,8 @@ module.name_mapper.extension='scss' -> '<PROJECT_ROOT>/packages/webviz-core/src/
 module.name_mapper.extension='md' -> '<PROJECT_ROOT>/packages/webviz-core/src/types/modules/Markdown.js.flow'
 module.name_mapper.extension='ne' -> '<PROJECT_ROOT>/packages/webviz-core/src/types/modules/Nearley.js.flow'
 module.name_mapper.extension='worker.js' -> '<PROJECT_ROOT>/packages/webviz-core/src/types/modules/WorkerLoader.js.flow'
+module.name_mapper='stories/\(.*\)' -> '<PROJECT_ROOT>/stories/\1'
+module.name_mapper='common/\(.*\)' -> '<PROJECT_ROOT>/common/\1'
```
janpaul123: Should we instead add `module.system.node.resolve_dirname=.` for consistency with https://github.com/cruise-automation/webviz/pull/212/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR69?

stories/config.js
```diff
@@ -19,6 +19,7 @@ installChartjs();
 
 addDecorator(initScreenshot());
 setScreenshotOptions({
+  waitFor: "waitFor",
```
janpaul123: Maybe a quick comment to explain what we use this for?

stories/assertionTestUtils.test.js
```diff
@@ -0,0 +1,89 @@
...
+      });
+
+      it("fails correctly", () => {
+        let error;
+        try {
+          assertionTestExpect({ some: "otherthing" }).toEqual({ some: "thing" });
+        } catch (e) {
+          error = e;
+        }
+        expect(error).not.toEqual(undefined);
```
janpaul123:
```suggestion
        expect(() => assertionTestExpect({ some: "otherthing" }).toEqual({ some: "thing" })).toThrow();
```

#nit. Can do something similar below.

stories/config.js
janpaul123: Also, just to confirm, is this just a no-op for screenshot tests that don't define `waitFor`?

packages/regl-worldview/src/stories/assertionStories/Worldview.stories.js
```diff
@@ -0,0 +1,155 @@
...
+    });
+    element.dispatchEvent(mouseEvent);
+    await timeout(100);
+  };
+}
+
+const mouseMove = createClickFunction("mousemove");
+const mouseUp = createClickFunction("mouseup");
+const mouseDown = createClickFunction("mousedown");
+const doubleClick = createClickFunction("dblclick");
```
janpaul123: Maybe rename the function above `emitClickEvent(eventName: "mousemove" | "mouseup" | "mousedown" | "dblclick"): Promise<void>`, remove the function nesting, and then just call it as `emitClickEvent("mousemove");`? Then you don't need the indirection.

stories/config.js
davidswinegar: It is a no-op when `waitFor` is not defined

webpack.config.js
davidswinegar: I'll remove this

PR #213, opened 2019-08-27
Add FOV and frustum plane distances to cameraState

nlittau-cruise:
## Summary

Add `fov`, `near`, and `far` to `cameraState` so they can be customized. Also add these new properties to the interactive documentation:

CameraStateControls2

## Test plan

Unit tests pass with no regressions; added the new properties to `CameraStore` tests. Manual inspection confirms that the properties behave as expected.

## Versioning impact

Minor update due to the backward-compatible new features.

### DIFF
--- a/packages/regl-worldview/src/camera/CameraStore.test.js
+++ b/packages/regl-worldview/src/camera/CameraStore.test.js
@@ -127,7 +127,7 @@ describe("camera store", () => {
       targetOffset: [0, 0, 0],
       targetOrientation: [0, 0, 0, 1],
       perspective: false,
-      fov: Math.PI / 3,
+      fovy: Math.PI / 3,
       near: 0.2,
       far: 1000,
     };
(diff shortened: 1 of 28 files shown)


packages/regl-worldview/src/camera/CameraStore.test.js
```diff
@@ -135,6 +138,26 @@ describe("camera store", () => {
         store = new CameraStore(undefined, cameraState);
         expect(store.state).toEqual(cameraState);
       });
+
+      it("has missing properties filled in from DEFAULT_CAMERA_STATE", () => {
```
janpaul123: #nit tip for writing test descriptions: usually they're most readable and compact when starting with an active verb, e.g. `it("fills in missing properties from DEFAULT_CAMERA_STATE"`.

nlittau-cruise: I considered that phrasing, but it didn't make semantic sense to me. Together with the component being described, the test description would be `initial store state fills in missing properties from DEFAULT_CAMERA_STATE`. But the initial store state doesn't fill in missing properties—the camera store does that.

PR #221, opened 2019-09-04
Unify inScreenshotTests

davidswinegar:
We have three copies of the same function. Unfortunately we can't
really share them since they have very different contexts and one
is inside the webviz-core package, but instead make sure they have
the same definition.

Test plan: tests pass


janpaul123:
What's up with those screenshots? Do we have flakey ones?

jtbandes:
We should really start using that `waitFor` feature...

PR #223, opened 2019-09-05
Replace custom expect library with Jest

davidswinegar:
I hadn't seen that Jest exports their own expect library that works
inside the dom before completing the integration tests. Replace
ours with theirs in the assertion stories because their error
messages are better and library is more complete.

Test plan: no screenshot changes


jtbandes:
Sweet 😄

PR #227, opened 2019-09-11
Fix mouse event argument mix-up

### LINKED ISSUE #226: MouseEvent arguments mix-up
Hello, I've just upgraded to v0.2.2, and two of the fields seem to have been swapped unintentionally. Previously, the arguments were `event, { ray, object(s) }`. 

Now, I'm seeing that the second argument is still an object with the field ray, but this field then points to a mouse event object. Meanwhile, the first argument sent turns out to be a ray. To explain a little more this what the second argument looks like now: 
```
{
  objects: [{…}]
  ray: Class
    altKey: false
    bubbles: true
    button: 0
    buttons: 0
    cancelable: true
    clientX: 640
    clientY: 233
    ctrlKey: false
    ...
  }
}
```

That Ray looks a lot like a MouseEvent...

I'm guessing this was just an accident. I'll switch these out in my code, in the meantime, and leave myself a note.

davidswinegar:
## Summary
Fix bug of switching event and ray in object callbacks. Fixes #226.

## Test plan
Included integration test.

## Versioning impact
Upgraded to version 0.2.3.

### DIFF
--- a/packages/regl-worldview/src/commands/Command.js
+++ b/packages/regl-worldview/src/commands/Command.js
@@ -98,7 +98,7 @@ export default class Command<T> extends React.Component<Props<T>> {
     });
   }
 
-  handleMouseEvent(objects: MouseEventObject[], e: MouseEvent, ray: Ray, mouseEventName: MouseEventEnum) {
+  handleMouseEvent(objects: MouseEventObject[], ray: Ray, e: MouseEvent, mouseEventName: MouseEventEnum) {
     const mouseHandler = this.props[mouseEventName];
     if (!mouseHandler || !objects.length) {
       return;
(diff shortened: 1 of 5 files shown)


packages/regl-worldview/src/commands/Command.js
```diff
@@ -98,7 +98,7 @@ export default class Command<T> extends React.Component<Props<T>> {
     });
   }
 
-  handleMouseEvent(objects: MouseEventObject[], e: MouseEvent, ray: Ray, mouseEventName: MouseEventEnum) {
+  handleMouseEvent(objects: MouseEventObject[], ray: Ray, e: MouseEvent, mouseEventName: MouseEventEnum) {
```
jtbandes: Why didn't Flow catch this at the call site?

davidswinegar: This file currently has flow disabled, which I didn't realize until after finishing the v0.2 release (turns out nothing has errors!). We have a ticket to fix this soon.

PR #246, opened 2019-10-04
Make lines truly instanced

davidswinegar:
Make the lines component able to take multiple poses per marker.
This will help us implement some markers that have different poses
instead of different points. While we're at it, make the loading
of colors into lines more efficient by passing a typed array
directly to regl.

Test plan: added a story, test that existing tests continue working.

Versioning impact: this will require a minor version bump

### DIFF
--- a/packages/regl-worldview/src/commands/Lines.js
+++ b/packages/regl-worldview/src/commands/Lines.js
@@ -7,8 +7,9 @@
 //  You may not use this file except in compliance with the License.
 
 import * as React from "react";
+import flatten from "lodash/flatten";
 
-import type { Line } from "../types";
+import type { Line, Vec4, Vec3, Color, Pose } from "../types";
 import { defaultBlend, withPose, toRGBA, shouldConvert, pointToVec3 } from "../utils/commandUtils";
 import { nonInstancedGetChildrenForHitmap } from "../utils/getChildrenForHitmapDefaults";
 import Command, { type CommonCommandProps } from "./Command";
@@ -87,6 +88,9 @@ attribute vec3 positionA;
 attribute vec3 positionB;
 attribute vec3 positionC;
 attribute vec3 positionD;
+// per-instance pose attributes
+attribute vec3 posePosition;
+attribute vec4 poseRotation;
 
 uniform mat4 projection, view;
 uniform float viewportWidth;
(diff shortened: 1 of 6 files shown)


packages/regl-worldview/src/commands/Lines.js
```diff
@@ -346,7 +368,96 @@ const lines = (regl: any) => {
...
+        positionArray = new Float32Array(VERTICES_PER_INSTANCE * 3);
+        rotationArray = new Float32Array(VERTICES_PER_INSTANCE * 4);
+      }
+      positionArray.fill(0);
+      rotationArray.fill(0);
+      // Rotation array identity is [x: 0, y: 0, z: 0, w: 1]
+      for (let index = 0; index < VERTICES_PER_INSTANCE; index++) {
+        rotationArray[index * 4 + 3] = 1;
+      }
+    }
```
brianc: You could statically allocate the non-instanced posePositionBuffer and poseRotationBuffer for one the command does not have instanced poses.  I think that's the happy path / default cause & in that situation instead of doing any of this work you could just pass the non-instanced buffers [right here](https://github.com/cruise-automation/webviz/pull/246/files#diff-60b9c6ba8357cc8cd56a2306263d0644R318-R325)

packages/regl-worldview/src/commands/Lines.js
```diff
@@ -375,52 +486,52 @@ const lines = (regl: any) => {
...
-        primitive: "triangle strip",
-        alpha: debug ? 0.2 : 1,
-        monochrome,
-        instances,
-        scaleInvariant,
-      });
-      if (debug) {
-        command({
-          ...props,
+      // Use Object.assign because it's actually faster than babel's object spread polyfill.
```
brianc: is it possible to use neither? Just construct a single object w/ all known keys, taking any existing key from `props` and providing the default in the `{ }` part of the `Object.assign` call?  I reckon that's far faster than either more dynamic route.

davidswinegar: It is possible to use neither but I tested object.assign and it seems to take essentially no time so I feel pretty comfortable about using it - the tradeoff of requiring developers to enumerate keys here kinda sucks

davidswinegar:
@brianc thanks for the look over, should be ready for another round

PR #249, opened 2019-10-07
Fix minor Worldview issues and release new version

davidswinegar:
## Summary

Fix two minor issues:
* Don't wrap Grid in a `React.memo` call. No other commands do this.
* Don't require "colors" in a command prop for instancing to be enabled for that command in our default `createInstancedGetChildrenForHitmap`. Instead we always populate the `colors` array if there is a `points` array. The current behavior is confusing, since many people expect instancing to be enabled even if they're using a single color, and the workaround (provide an empty color array) is a bit weird.

## Test plan

Automated testing/flow should be able to catch all issues here.

## Versioning impact

Release new version of Worldview that includes the changes to Lines.

### DIFF
--- a/packages/regl-worldview/src/utils/getChildrenForHitmapDefaults.js
+++ b/packages/regl-worldview/src/utils/getChildrenForHitmapDefaults.js
@@ -72,7 +72,7 @@ function instancedGetChildrenForHitmapFromSingleProp<T: any>(
   const idColors = assignNextColors(prop, instanceCount);
   const startColor = idColors[0];
   // We have to map these instance colors to `pointCountPerInstance` number of points
-  if (hitmapProp.colors && hitmapProp.points && hitmapProp.points.length) {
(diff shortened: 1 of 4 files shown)


packages/regl-worldview/src/utils/getChildrenForHitmapDefaults.js
```diff
@@ -72,7 +72,7 @@ function instancedGetChildrenForHitmapFromSingleProp<T: any>(
   const idColors = assignNextColors(prop, instanceCount);
   const startColor = idColors[0];
   // We have to map these instance colors to `pointCountPerInstance` number of points
```
vidaaudrey: nit: update the comment as well

davidswinegar: I think this comment still applies, it's about the action we're taking instead of the condition that I'm changing

PR #261, opened 2019-10-29
Add monaco editor fix

troygibb:
## Summary

There was a duplicate `monaco-editor` dependency in the `webviz-core` package.json, causing that instantiation to not load the Typescript language service. I'll take out the corresponding dependency out of the private repo. 

## Test plan

Tested locally with docs.

## Versioning impact

None

### DIFF
--- a/packages/webviz-core/package-lock.json
+++ b/packages/webviz-core/package-lock.json
@@ -952,11 +952,6 @@
         "moment": ">= 2.9.0"
       }
     },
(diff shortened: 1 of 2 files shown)


troygibb:
The longer term fix here is to not depend on the `monaco-editor-webpack-plugin` so we don't have to have a top level `monaco-editor` dependency and can appropriately nest it in `packages/webviz-core/package.json`.

PR #263, opened 2019-11-01
Show /tf in the 3D panel topicTree

vidaaudrey:
## Summary
Currently `/tf` topic and visualization don't show up in the 3D panel because there is no default topic tree configured to show this topic. 
image

Fixed by adding TF to the default topic tree config.

## Test plan
Manually test http://localhost:8080/try/?demo and the `/tf` should show up after the bag is loaded:
image

## Versioning impact
No impact.


janpaul123:
@vidaaudrey can you please also make this change to our internal repo so it won't get overwritten during the next open source release?

PR #265, opened 2019-11-05
Capture error in requestAnimationFrame

troygibb:
## Summary

This particular block of code throws a lot of exceptions (`(regl) context lost
`) when the OS occasionally resets the GPU (More [info here](https://www.khronos.org/webgl/wiki/HandlingContextLost)). Since it was running in `requestAnimationFrame`, our React Error Boundaries were not catching the error. 

## Test plan

I've been having a hard time reproducing this particular error. You basically have to intentionally over-utilize the GPU to get this to occur, but it's very inconsistent.

## Versioning impact

Needs a minor bump.


packages/regl-worldview/src/Worldview.js
```diff
@@ -153,7 +153,11 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
     if (!this._tick) {
       this._tick = requestAnimationFrame(() => {
         this._tick = undefined;
-        worldviewContext.paint();
+        try {
+          worldviewContext.paint();
+        } catch (e) {
+          console.error(e);
```
davidswinegar: maybe throw this if it doesn't match `(regl) context lost`? We don't necessarily want errors in this look to be thrown away if they would be useful. Otherwise users might see a black 3d panel and we'd have no error notices

janpaul123:
Btw looks like there's some flakey node playground screenshots — might be worth looking into

packages/regl-worldview/src/Worldview.js
```diff
@@ -158,7 +158,17 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
...
       this._tick = requestAnimationFrame(() => {
         this._tick = undefined;
-        worldviewContext.paint();
+        try {
+          worldviewContext.paint();
+        } catch (error) {
+          // Regl automatically tries to reconnect when losing the canvas 3d context.
+          // We should log this error, but it's not important to throw it.
+          if (error.message === "(regl) context lost") {
+            console.error(error);
```
janpaul123: `console.warn` might be better if it's not critical.

PR #266, opened 2019-11-08
Add option to only render triangles in the hitmap

davidswinegar:
## Summary

For performance reasons we sometimes want to only render triangles when clicking. Add an option to allow this. Modify the command API somewhat to enable this. Also do a small cleanup: move all of the `createInstancedGetChildrenForHitmap` calls out of the render functions of commands.

## Test plan

Added a story with this new behavior.

## Versioning impact

Minor package increment needed.

### DIFF
--- a/docs/src/4.11.Triangles.mdx
+++ b/docs/src/4.11.Triangles.mdx
@@ -35,6 +35,8 @@ type Triangle = {
   colors?: Color[],
   points: [{ x: number, y: number, z: number }]
   // Pass true to not render the triangles to the screen - just the hitmap.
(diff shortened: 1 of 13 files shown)


docs/src/4.11.Triangles.mdx
```diff
@@ -34,6 +34,8 @@ type Triangle = {
   color: Color,
   colors?: Color[],
   points: [{ x: number, y: number, z: number }]
+  // Pass true to not render the triangles to the screen - just the hitmap.
+  onlyRenderInHitmap?: boolean,
```
vidaaudrey: Would be good to add some use cases  to the doc to explain why this is needed.

PR #268, opened 2019-11-19
A first take on stopPropagation within worldview

robin-pham:
## Summary

We want to mimic how the browser [Event.stopPropagation](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation) works in Worldview, e.g. calling `event.stopPropagation()` in a click handler on a nested div would prevent any click event handlers from firing off in ancestor divs. This PR accomplishes a similar behavior for Worldview by changing the order in which global vs command level mouse handlers are fired as well as using [React's synthetic event's isPropagationStopped](https://reactjs.org/docs/events.html#overview) property.

## Test plan

Added two storybook tests

## Versioning impact

This probably necessitates a major/minor patch as it can be a breaking change to those who use `e.stopPropagation` in command level mouse handlers or those who expect the worldview level mouse handlers to fire off before the command ones.

### DIFF
--- a/packages/regl-worldview/src/stories/assertionStories/Worldview.stories.js
+++ b/packages/regl-worldview/src/stories/assertionStories/Worldview.stories.js
@@ -26,6 +26,15 @@ const cube = {
   color: { r: 1, g: 0, b: 1, a: 0.5 },
 };
 
+const underCube = {
+  pose: {
+    orientation: { x: 0, y: 0, z: 0, w: 1 },
+    position: { x: 0, y: -30, z: 0 },
+  },
+  scale: { x: 10, y: 10, z: 10 },
+  color: { r: 0, g: 1, b: 1, a: 0.5 },
+};
+
 async function emitMouseEvent(eventName: "mousemove" | "mousedown" | "mouseup" | "dblclick"): Promise<void> {
   const [element] = document.getElementsByTagName("canvas");
   if (!element) {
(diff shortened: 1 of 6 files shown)


davidswinegar:
One thing I forgot about - you should add this to the Worldview documentation. I think `3.4.MouseEvents.mdx` would be a good place. A small section with an example (doesn't have to run, your test code seems fine for this) and an explanation should be sufficient.

robin-pham:
Screenshot_2019-11-19_17-15-35

are we cool with ^ this copy?

davidswinegar:
looks good to me!

packages/regl-worldview/src/stories/assertionStories/Worldview.stories.js
```diff
@@ -175,4 +184,54 @@ stories
...
+      ),
+      assertions: async (getTestData) => {
+        await emitMouseEvent("mousedown");
+        const result = await getTestData();
+        expect(result.length).toEqual(1);
+        expect(result[0].object).toEqual(cube);
+      },
+    })
+  )
+  .add(
```
vidaaudrey: Would be nice to add more tests with different `layerIndex` and order , e.g. 
```
<Cube layerIndex={2} {... stopPropgation()}><Cubes>
<Cube layerIndex={1}><Cubes>
```
or:
```
<Cube layerIndex={1}><Cubes>
<Cube layerIndex={2} {... stopPropgation()} ><Cubes>
```

robin-pham:
> Curious what you think about removing command level handlers

I like having them. Having only handlers at the global level became this giant mess of 20+ `if` conditionals trying to figure out what was being clicked and which logic to fire.

> I think it's a little confusing to have to deal with sibling propagations

Yeah the ordering of events can be tricky but not too different from managing the ordering of `if` logic in the global handler. For the most part I don't see why sibling propagations will be an issue as the majority of the time users will just be interested in interacting with the object that's on top. Although I guess there can be cases similar to the DOM where someone wants multiple mouse handlers to fire while clicking on something that's a composite. 

> simplify the code with potential perf improvement

Yeah hitmap is definitely the biggest perf hit right now and I wish there was a faster way to go about it. But I don't see why removing the component level handlers makes this any better? It still seems like the problem is rendering the hitmap in the first place as well as `enableStackedObjectEvents` making it so that there are multiple renders depending on how many objects are under the mouse. Everytime I look at performance flame charts it seems like the biggest lag comes from the `readHitmap` method and its nested culprits `getChildrenForHitmapWithOriginalMarker`, `readPixels` and `REGLCommand`.

> allow Worldview to do some internal optimization such as command batching without worrying about the individual command handlers, e.g. batch 20 topics with Lines together

What's the reason for why command batching can't be done? Can we abstract React 'commands' from regl 'commands' such that the regl commands can still do batching but there's still the mapping between objects -> React 'commands' in order to figure out the appropriate mouse handler?

davidswinegar:
@vidaaudrey @robin-pham I think one of the biggest hitmap optimizations we could make right now in Webviz is that we call it multiple times on a click - for mousedown, mouseup, click - even when most of the time we only care about the click event. I'm not sure if you're seeing the same in Scenebuilder. We just haven't gotten around to making fine-grained control of these handlers work in Webviz yet.

robin-pham:
Ya that's an issue although the `mouseMove` is still just as big, if not bigger, of a performance problem

Issue #271, opened 2019-11-21
Colors issue on State transitions panel


shaeqahmed:
When the value for a topic is -1, seems that the color is a greyed out very hard to see color. Also for some topics say the message goes from 'AUTONOMY' to 'NAVIGATE' for a single message, the color remains the same. Would be nice if we could improve the the colors to make sure it is always good enough contrast and also to make sure that the color changes whenever the message value is different.

jtbandes:
Can you post screenshots of the problems you're seeing, especially for the second issue you mentioned where it doesn't change colors when there's a new value in the message?

shaeqahmed:
pic for second issue:
![image](https://user-images.githubusercontent.com/13088492/69374257-8b302680-0c5a-11ea-900d-2ecd90971c18.png)

janpaul123:
This should be an easy fix. I'll look into it.

shaeqahmed:
Thanks! Would be nice to adjust the color scheme as well to make sure it is easy to see for everyone including those with disabilities.

See: https://webaim.org/resources/contrastchecker/

janpaul123:
We did use a colour scheme that should be pretty distinctive for vision disabilities, see https://github.com/cruise-automation/webviz/blob/master/packages/webviz-core/src/util/plotColors.js#L14 but of course we're always open to improvements!

shaeqahmed:
Great to hear. Looks like it is a bug then in the code that determines color for value, as when the value is -1 it is very hard to see:

![image](https://user-images.githubusercontent.com/13088492/69381601-f59c9300-0c69-11ea-90c9-2e38bc0bc6a4.png)

janpaul123:
Yeah it's a bug. I have a fix incoming! 😄

shaeqahmed:
Pulled in latest changes, and the issue with the transparent color for -1 is fixed but the color still stays same when the value changes, just as shown in the first picture.

janpaul123:
We cycle through a fixed number of colours so you can get unlucky with which value is assigned to which colour.

shaeqahmed:
I think the most important part even if colors are to be reused, is to ensure that no two consecutive state messages, within the same topic, will have the same color unless they are of the same value. That shouldn't be too difficult, would be a good enough solution.

janpaul123:
[...]

PR #273, opened 2019-11-21

janpaul123:
Changelog:
- Added a new "Flat" mode to the 3D panel topic picker. Now when playing multiple bags in the "Flat" mode, bag1 topics are grouped under `Bag` and bag2 topics are grouped under `Bag 2 /webviz_bag_2`. In the "Flat" mode, you can temporarily toggle the topic's visibility by clicking the eye icon next to the edit icon. This is the first step in the ongoing work to improve multi-bag comparison in the 3D panel.
- Sped up local bag reading by decompressing bag chunks ahead of playback.
- Added support for negative slices in the topic path syntax, e.g. to get the last element of an array.
- Added the ability to plot numerical values in Diagnostics - Detail panel.
- Added ability to pan, zoom and do auto zoom in imageView panel.
- Clicking on stacked objects is now automatically turned on in the 3D panel when playback is paused. If there are multiple clicked objects, the object id will be displayed in the context menu if available.
- Added ability to toggle visibility of Plot panel's legends.
- Implemented a workaround to runaway disk usage due to a Chrome bug, which has been causing very slow Webviz start up times. If possible, please clear out your local instance of IndexedDB on Webviz.
- Fixed image panel with synchronized markers not loading data when seeking, by increasing the amount of data we load when seeking.
- Fixed a `Time should never move backwards` error when seeking.
- Fixed negative numbers showing incorrect colors in the State Transitions panel. #271


vidaaudrey:
Not sure if the test is flaky or actually failing, may worth check if something is wrong:
- IntegrationGLTFScene-Clicks-on-a-GLTFSCene-object-with-an-object-behind-it---worldview-event-handler.png

janpaul123:
@davidswinegar can you look at that test? Both the "before" and "after" look incorrect.

PR #279, opened 2019-12-02
fix jest config for CSS files imported from node_modules

jtbandes:
CSS files directly imported from node_modules would cause an error when jest failed to parse them. CSS files were being handled in the `transforms`, but these are ignored for node_modules, so use `moduleNameMapper` instead.


jtbandes:
Mostly I did that because I read about it in the docs and was confused why we didn't do it that way, so I tried to see if it worked and it did!

PR #280, opened 2019-12-03
Add explanation of how to view a rosbag

lucasw:
From #161

## Summary

I couldn't figure out how to load a rosbag until coming across #161

## Test plan

I ran npm run docs and browsed to try, then dragged in a rosbag.

## Versioning impact

No.

### DIFF
--- a/packages/webviz-core/README.md
+++ b/packages/webviz-core/README.md
@@ -2,6 +2,8 @@
 
 This is the open source version of Webviz, the browser-based rosbag analysis tool.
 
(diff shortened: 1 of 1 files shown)


jtbandes:
FYI, we also have this in the UI already:

image

lucasw:
I knew about the drag n drop part, it was running 'docs' and changing the url to get to the page that allows it that isn't obvious.

jtbandes:
A lot of this info is also in the main readme: https://github.com/cruise-automation/webviz/blob/master/README.md#developing  So I wonder if we can link there instead so they don't get out of date?

PR #282, opened 2019-12-03
[WIP] Node Playground Flakey Screenshots Fix

troygibb:
## Summary

_What's the motivation for this change, what approach did you take, and why?_

## Test plan

_How was this change tested & verified?_

## Versioning impact

_Does this change necessitate a [major, minor, or patch](https://semver.org/) semver update for the affected package?_

### DIFF
--- a/packages/webviz-core/src/panels/NodePlayground/Editor.js
+++ b/packages/webviz-core/src/panels/NodePlayground/Editor.js
@@ -21,7 +21,7 @@ import {
 
 const VS_WEBVIZ_THEME = "vs-webviz";
 
-type Props = { script: string, setScript: (script: string) => void, vimMode: boolean };
+export type Props = { script: string, setScript: (script: string) => void, vimMode: boolean };
 const Editor = ({ script, setScript, vimMode }: Props) => {
   const editorRef = React.useRef(null);
   const vimModeRef = React.useRef(null);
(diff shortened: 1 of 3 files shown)


jtbandes:
Looks like it didn't work:

image

Can we make it actually wait until it finishes loading?

janpaul123:
Maybe first try something crazy like 10 seconds to see if slow loading is indeed the problem

troygibb:
@jtbandes ya I have been thinking along that track.. the really odd thing is that our internal CI has not exhibited any flakeyness for these screenshots.

If it is indeed a loading issue, we could add a condition that does not lazy load the `Editor` component if it's in screenshots, since I'm assuming that's where we're running into an issue here (we don't lazy load anywhere else). Open to more suggestions here as well.

jtbandes:
There is a `waitFor` option in https://github.com/reg-viz/storycap that can be customized.

troygibb:
Ya I was looking into that... we use it currently for `regl-worldview`. Seems like a pretty good pattern that @davidswinegar has implemented for awaiting different interactions.

The main issue I couldn't think around was how to await a promise nested in the React tree from the storybook level. I know React.Suspense has a clever use of error boundaries in order to determine whether to render a pending/resolved component, maybe we could hook into that somehow.

Or we could stub out the React.Suspense component in the storybook config so that it invokes the global `waitFor` promise. One problem there is that `storycap` allows only one `waitFor` promise. We could potentially overwrite temporarily.

PR #289, opened 2019-12-17

davidcrawford:
Changelog:
- Update license header from "GM Cruise LLC" to "Cruise LLC"
- In the 3D panel, moved "Flatten markers" from the topic selector to the panel settings since it's not associated with any topics.
- Added ability to set min and max values for both x and y axes in the 2D Plot panel.
- Added support for slices using global variables in the topic path syntax, e.g. to get $my_idx of an array.
- Added ability to diff messages from 2 different topics in Raw Messages panel.
- Added a warning when a topic in the 3D panel is in the non-root frame. Webviz has a known bug with applying transforms correctly when they are not in the root frame, so this change alerts users to that fact when it is affecting their view.
- Fixed edge case where wrapping around playback to the start of the bag would miss messages which timestamp are at the very start of the bag.
- Fixed inability to copy large messages to the clipboard in JSON format from the Raw Messages panel.
- Fixed a Plot panel bug where looping through the bag caused it to crash.
- Allow skip publishing in Node Playground by returning 'undefined'.
- Fixed topics disappearing in the 3D panel flat topic list.
- Fixed inability to play markers from webviz_bag_2 in the image panel. The webviz_bag_2 image topics in the image panel topic selector menu are also now grouped the same way that image topics from the first bag are.
- Now we always show some data in panels upon first load, even if a seek-to URL query parameter is not given.
- Various improvements to playback performance, including using threads for rendering images, a Playback Performance panel, more playback speed options, and other more minor improvements.


vidaaudrey:
Gonna merge this and so it's clear to see the chart.js related screenshots in #292

PR #291, opened 2019-12-18
Worldview: add GLText command

jtbandes:
## Summary

Adds a new `<GLText />` command which is API-compatible with `<Text />`. Instead of overlaying DOM nodes on top of Worldview, the text is rendered in GL so it can participate in occlusion. Rendering in GL is also much faster than managing many separate DOM elements; see the GLText docs page for a performance demo.

All the necessary characters are rendered into a texture atlas using a [signed distance field](https://github.com/libgdx/libgdx/wiki/Distance-field-fonts) representation, then all characters from all children are drawn in a single draw call. More implementation details are explained in GLText.js.

GLText demo

Many thanks to the following resources/implementations/references which were helpful in informing this implementation!
- Valve: [Improved Alpha-Tested Magnification for Vector Textures and Special Effects](https://steamcdn-a.akamaihd.net/apps/valve/2007/SIGGRAPH2007_AlphaTestedMagnification.pdf)
- Mapbox: [Drawing Text with Signed Distance Fields in Mapbox GL](https://blog.mapbox.com/drawing-text-with-signed-distance-fields-in-mapbox-gl-b0933af6f817)
- Otto Chrons: [Text rendering in OpenGL](http://hack.chrons.me/opengl-text-rendering/)
- Stack Overflow: [Better Quality Text in WebGL](https://stackoverflow.com/q/25956272/23649)
- https://github.com/mapbox/tiny-sdf
- [deck.gl TextLayer](https://github.com/uber/deck.gl/tree/78891632763fa655bec04d320fec3c66c85f6365/modules/layers/src/text-layer)
- [mapbox-gl-js GlyphManager](https://github.com/mapbox/mapbox-gl-js/blob/26a47e4e11dec83dc6d6190a904224c290c74f96/src/render/glyph_manager.js)
- https://github.com/dy/font-atlas-sdf

## Test plan
Added stories/screenshot tests and docs page demo. Unfortunately the code is not too easily amenable to unit testing (other than perhaps the font atlas memoization).

## Versioning impact
This PR is purely additive, so doesn't necessitate a major version bump. The GLText command might eventually replace Text, but we'd like to get some mileage on it first — we'll be testing it in Webviz, and would love for any external users to give it a try as well. A few areas for improvement would probably need to be addressed before deprecating the DOM-based Text command: solid rectangular background instead of outline (or at least improved visibility at small sizes); highlight ranges to replace browser-native ⌘F; hitmap support.


packages/regl-worldview/src/commands/GLText.js
```diff
@@ -0,0 +1,393 @@
...
+// (and whenever new characters are being rendered). Then one instanced draw call is made with an instance
+// per character which reads from the corresponding place in the texture atlas.
+//
+// Possible future improvements
+// ============================
+// - Add hitmap support.
+// - Allow customization of font style, maybe highlight ranges.
+// - Add a scaleInvariant option.
+// - Consider a solid rectangular background instead of an outline. This is challenging because the
+//   instances currently overlap, so there will be z-fighting, but might be possible using the stencil buffer and multiple draw calls.
```
brianc: if you used a fixed width font would instances still overlap?

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -0,0 +1,393 @@
...
+    });
+
+    return (props: $ReadOnlyArray<TextMarker & { billboard?: boolean }>) => {
+      const prevNumChars = charSet.size;
+      for (const { text } of props) {
+        for (const char of text) {
+          charSet.add(char);
+        }
+      }
+      const charsChanged = charSet.size !== prevNumChars;
```
brianc: I like the concept of lazy generating/updating the texture only when new characters arrive from an ease of use standpoint...but it feels like there would be a lot of wins if you were able to specify the universe of characters up front instead of doing a neasted loop & set construction on ever render.  With thousands of text markers this could potentially add up to quite a bit of overhead over the course of a long scene, particularly once all the characters have been encountered.  This goes double for ros as text markers only support ASCII.  Maybe at the minimum would be nice to allow for a pre-specified universe or maybe just `ascii: true` and bypassing the set construction on every render of every instance of the command.  Nothing that needs to be added in this PR but wanted to call it out as a future perf win.

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -0,0 +1,393 @@
...
+const BG_COLOR_LIGHT = Object.freeze({ r: 1, g: 1, b: 1, a: 1 });
+const BG_COLOR_DARK = Object.freeze({ r: 0, g: 0, b: 0, a: 1 });
+
+// Build a single font atlas: a texture containing all characters and position/size data for each character.
+const createMemoizedBuildAtlas = () =>
+  memoizeOne(
+    (charSet: Set<string>): FontAtlas => {
+      const tinySDF = new TinySDF(FONT_SIZE, BUFFER, SDF_RADIUS, CUTOFF, "sans-serif", "normal");
+
+      const fontStyle = `${FONT_SIZE}px sans-serif`;
```
brianc: future optimization would be nice to allow font specification to the command (likely not on the marker level since you'd need memoized build atlases for each marker)

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -0,0 +1,393 @@
...
+      const canvas = document.createElement("canvas");
+      const ctx = canvas.getContext("2d");
```
brianc: I suppose you could memoize this separately with `memoizeOne` taking no arguments - then you could share a single reference to a canvas versus 1 for each instance of the text command and then creating a new one each time the character universe changes.

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -0,0 +1,393 @@
...
+          height: textureHeight,
+          format: "alpha",
+          wrap: "clamp",
+          mag: "linear",
+          min: "linear",
+        });
+      }
+
+      let totalInstances = 0;
+      const estimatedInstances = props.reduce((sum, marker) => sum + marker.text.length, 0);
```
brianc: you could save a loop and do this when you go through each char on line 252

jtbandes: point here is to know what size buffer to allocate up front

packages/regl-worldview/src/commands/GLText.js
jtbandes: Yeah because there is a radius + buffer around each character in the sdf which you can see on https://mapbox.github.io/tiny-sdf/. I think these demos don't have any z-fighting because they simply don't use the depth buffer

packages/regl-worldview/src/commands/GLText.js
brianc: yah - aren't you computing the entire character count of all characters in each text marker?  I was saying you're _also_ looping through each text marker around line 252.  You could keep the count up there & not have to loop again.  It's a very small optimization by removing a loop.

packages/regl-worldview/src/commands/GLText.js
jtbandes: Given that we have to iterate over each character anyway to build the attribute arrays, do you think checking whether the character is in the set really adds much extra overhead? (Chrome perf tools were giving me some trouble and not showing amount of time spent per line.)  I made sure not to actually allocate a new Set unless we found some new characters.

packages/regl-worldview/src/commands/GLText.js
jtbandes: oh gotcha! Should've checked the line number 🤣

packages/regl-worldview/package.json
```diff
@@ -26,10 +26,12 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.3.1",
+    "@mapbox/tiny-sdf": "^1.1.1",
     "distance-to-line-segment": "0.2.0",
     "earcut": "^2.1.3",
     "gl-matrix": ">=2.6.1",
     "lodash": "^4.17.11",
+    "memoize-one": "5.1.1",
```
vidaaudrey: why locking the version to 5.1.1 ?

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -0,0 +1,397 @@
...
+  varying float vEnableOutline;
+  varying vec4 vForegroundColor;
+  varying vec4 vOutlineColor;
+  void main() {
+    float dist = texture2D(atlas, vTexCoord).a;
+
+    // fwidth(dist) is used to provide some anti-aliasing. However it's currently only used
+    // between outline and text, not on the outer border, because the alpha blending and
+    // depth test don't work together nicely for partially-transparent pixels.
+    if (vEnableOutline > 0.5) {
```
troygibb: clarifying question -- `vEnableOutline` is only either 1 or 0 right?

jtbandes: Yup. When it was a uniform (before I combined all the markers into 1 draw call) I just had it as a boolean. But when I combined them I needed to pass it into the fragment shader with a `varying`, so it started doing interpolation, and some floating point rounding errors meant that `== 1.0` was giving glitchy results.

packages/regl-worldview/package.json
jtbandes: Good call, will change this to `^`

packages/regl-worldview/src/camera/cameraStateSelectors.js
```diff
@@ -189,9 +189,21 @@ const viewSelector: (CameraState) => Mat4 = createSelector(
...
+    return m;
+  }
+);
+
 export default {
   orientation: orientationSelector,
   position: positionSelector,
   targetHeading: targetHeadingSelector,
   view: viewSelector,
+  billboardRotation,
```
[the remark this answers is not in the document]
jtbandes: I think the naming convention in this file is kind of silly...less complicated to just export names directly

packages/regl-worldview/src/commands/GLText.js
troygibb: crazy. thanks!

PR #292, opened 2019-12-18
Fix ColorPicker screenshots by upgrading styled-components

vidaaudrey:
## Summary 
The inconsistent style-components version is causing a few ColorPicker related screenshots to fail in #289.  This is mainly due to the ColorPicker use `ref` which is only supported in styled-components v4 ([innerRef deprecated](https://www.styled-components.com/docs/api)). Fixed by updating styled-components to v4. 

## Test 
The existing screenshots should pass.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,5 @@
 {
   "editor.formatOnSave": true,
-  "eslint.autoFixOnSave": true,
   "flow.useNPMPackagedFlow": true,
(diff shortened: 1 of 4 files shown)


vidaaudrey:
> Any idea why the chart.js screenshots are changing? Is it related to this change or are they flaky?

The changes come from #289, likely not related to flakiness. Just merged #289. Gonna trigger a rebuild and see if the chart.js related screenshots will change

PR #294, opened 2019-12-18
regl-worldview version 0.3.2

vidaaudrey:
Adding GLText from #291


vidaaudrey:
I forgot to merge. Looks it's already late. @davidswinegar  has another PR to bump the version.

PR #296, opened 2019-12-20
<GLText /> Highlight

troygibb:
## Summary

Adds a `highlightedIndices` prop to the `<GLText />` command.

After reviewing an original implementation of this feature with @jtbandes, where I kept the `BUFFER` value and dynamically calculated the highlight box, we agreed it might be better to just take that value out altogether, since the background highlighting feature, currently available in the `<Text />` command, is more readable than outlining.

So the main change is really just taking out this buffer value out of all the calculations, and adding a separate `vEnableHighlight` variable to the fragment shader, which takes an opinionated stance on how highlighting is done. That is, we make the font black and the background color yellow. This highlighting UI is how Chrome does it at least. 

I also bumped up the `CUTOFF` constant significantly since there were some weird smudges at certain zoom levels. Since this value is used in a variety of places, I don't have a good theory on why it seemed to make things better. Since we no longer have a buffer in the atlas, I figured some pixels from neighboring characters might be getting mixed into the character textures. Bumping it seemed to work though with no visible regression.

One feature I did not include is a background color for the focused highlight element, which would mimic the behavior you see on a webpage as you iterate through `ctrl + f` search results (the focused highlighted element has an orange background color). This may/may not be necessary, I just figured I'd keep it simple for now.  

gl-text-highlight

## Test plan

Existing screenshots and one new screenshot with some dynamically highlighted text markers.

## Versioning impact

Minor.


packages/regl-worldview/src/commands/GLText.js
```diff
@@ -177,23 +178,25 @@ const frag = `
...
+  varying float vEnableBackground;
   varying vec4 vForegroundColor;
-  varying vec4 vOutlineColor;
+  varying vec4 vBackgroundColor;
+  varying float vEnableHighlight;
   void main() {
     float dist = texture2D(atlas, vTexCoord).a;
 
     // fwidth(dist) is used to provide some anti-aliasing. However it's currently only used
     // between outline and text, not on the outer border, because the alpha blending and
```
jtbandes:
```suggestion
    // when the solid background is enabled, because the alpha blending and
```

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -345,12 +346,14 @@ function makeTextCommand() {
...
-          outlineColor[4 * index + 0] = bgColor.r;
-          outlineColor[4 * index + 1] = bgColor.g;
-          outlineColor[4 * index + 2] = bgColor.b;
-          outlineColor[4 * index + 3] = bgColor.a;
+          backgroundColor[4 * index + 0] = bgColor.r;
+          backgroundColor[4 * index + 1] = bgColor.g;
+          backgroundColor[4 * index + 2] = bgColor.b;
+          backgroundColor[4 * index + 3] = bgColor.a;
+
+          enableHighlight[index] = marker.highlightedIndices ? +marker.highlightedIndices.includes(i) : 0;
```
jtbandes: nit: clarity
```suggestion
          enableHighlight[index] = marker.highlightedIndices && marker.highlightedIndices.includes(i) ? 1 : 0;
```

packages/regl-worldview/src/stories/GLText.stories.js
```diff
@@ -86,5 +86,51 @@ storiesOf("Worldview/GLText", module)
...
+          });
+        }
+        return { ...marker, highlightedIndices: Array.from(highlightedIndices) };
+      });
+
+      return (
+        <>
+          <div style={{ width: "100%", height: "95%" }}>
+            <Container cameraState={{ perspective: true, distance: 40 }} backgroundColor={[0.2, 0.2, 0.4, 1]}>
+              <GLText autoBackgroundColor>{markers}</GLText>
```
jtbandes: Might be nice to have a demo without autoBackgroundColor to make sure it works. Maybe having both is best.

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -53,11 +54,9 @@ type FontAtlas = {|
 
 // Font size used in rendering the atlas. This is independent of the `scale` of the rendered text.
 const FONT_SIZE = 40;
-const BUFFER = 10;
 const MAX_ATLAS_WIDTH = 512;
 const SDF_RADIUS = 8;
-const CUTOFF = 0.25;
-const OUTLINE_CUTOFF = 0.6;
+const CUTOFF = 0.75;
```
jtbandes: Rather than changing the cutoff, I think it might be better to keep the buffer around characters in the atlas, to avoid them accidentally bleeding into each other. You would keep your changes to destOffsets/srcWidths, but revert the changes to rowHeight and dx when building the atlas, and then just offset info.x/info.y by the buffer amount.

packages/regl-worldview/src/stories/GLText.stories.js
```diff
@@ -86,5 +86,51 @@ storiesOf("Worldview/GLText", module)
...
+        const highlightedIndices = new Set();
+        let match;
+        let regex;
+        try {
+          regex = new RegExp(searchText, "ig");
+        } catch (e) {
+          return marker;
+        }
+        while ((match = regex.exec(marker.text)) !== null) {
+          Array.from(Array(searchText.length).keys()).forEach((i) => {
```
jtbandes: TIL about this technique for getting a range 😮  But also, feel free to just use `_.range()` :)

Should probably use `match[0].length` instead though.

packages/regl-worldview/src/stories/GLText.stories.js
troygibb: will do

packages/regl-worldview/src/commands/GLText.js
troygibb: nice suggestion -- done!

packages/regl-worldview/src/stories/GLText.stories.js
```diff
@@ -86,5 +87,51 @@ storiesOf("Worldview/GLText", module)
...
+        const highlightedIndices = new Set();
+        let match;
+        let regex;
+        try {
+          regex = new RegExp(searchText, "ig");
+        } catch (e) {
+          return marker;
+        }
+        while ((match = regex.exec(marker.text)) !== null) {
+          range(0, searchText.length).forEach((i) => {
```
jtbandes: still want to use match[0].length

troygibb: woops missed that.

PR #305, opened 2019-12-31
Title capitalization

janpaul123:
Tiny consistency fix, it just bugged me that for webviz.io the title was "Webviz" but for webviz.io/app it's "webviz". We use the lowercase version in the title bar so I figured I'd just use that for the main page as well. 🙃

### DIFF
--- a/docs/public/index.html
+++ b/docs/public/index.html
@@ -24,9 +24,9 @@
     <meta name="viewport" content="width=device-width, initial-scale = 1.0,
 maximum-scale=1.0, user-scalable=no" />
     <meta name="description" content="Visualizing robotics data in the browser">
-    <title>Webviz</title>
+    <title>webviz</title>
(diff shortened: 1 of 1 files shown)


janpaul123:
Internally we also use lowercase "webviz" for the title. I dunno, I kind of like it lowercased. 😛 But happy to change it to something else too as long as it's consistent. For now I'm just going to merge this so at least it's consistent.

jtbandes:
>  I dunno, I kind of like it lowercased. 😛

Yeah me too... at some point in the past though, I just had to accept that it was awkward to begin sentences with lowercase webviz, and WebViz was ugly, so I should just learn to use Webviz, and that's how I've been capitalizing it in prose. That said, I think it's still valid for us to use an all-lowercase page title as a stylistic thing. (We also use an all-uppercase version on the webviz.io landing page.) 

So 👍 on this PR

janpaul123:
Yeah I feel exactly the same for exactly the same reasons. 😂

PR #308, opened 2020-01-02
Add flow typing to Worldview Command

davidswinegar:
This flow type was missing for a while, causing a bunch of flow
errors to pile up. Remove so that we can have Command fully typed.

Test plan: existing stories and functional integration tests should
be enough

Versioning impact: This can go out with the next version release

### DIFF
--- a/packages/regl-worldview/src/commands/Axes.js
+++ b/packages/regl-worldview/src/commands/Axes.js
@@ -45,8 +45,10 @@ const zAxis = {
   color: { r: 0.11, g: 0.51, b: 0.92, a: 1 },
 };
 
+type Axis = Line;
+
 type Props = {
-  children: Line[],
+  children: Axis[],
 };
 
 // Renders lines along the x, y, and z axes; useful for debugging.
--- a/packages/regl-worldview/src/commands/Axes.js
+++ b/packages/regl-worldview/src/commands/Axes.js
@@ -8,7 +8,7 @@
 
 import React from "react";
 
-import type { Pose, Color, Scale, Point, Vec3 } from "../types";
+import type { Point, Vec3, Line } from "../types";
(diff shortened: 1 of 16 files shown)


packages/regl-worldview/src/commands/Axes.js
```diff
@@ -45,15 +45,8 @@ const zAxis = {
   color: { r: 0.11, g: 0.51, b: 0.92, a: 1 },
 };
 
-type Axis = {
```
vidaaudrey: Maybe do type alias instead? `type Axis = Line`

packages/regl-worldview/src/commands/Command.js
```diff
@@ -37,25 +36,26 @@ export type CommonCommandProps = {
...
+  ...CommonCommandProps,
+  children?: T[] | T,
   // Deprecated, but here for backwards compatibility
   drawProps?: T[],
   reglCommand: RawCommand<T>,
-  ...CommonCommandProps,
 };
 
-export type CommandProps = Props;
+export type CommandProps<T> = Props<T>;
```
jtbandes: nit: maybe just remove Props and use CommandProps everywhere in this file?

packages/regl-worldview/src/commands/FilledPolygons.js
```diff
@@ -32,7 +32,7 @@ function flatten3D(points: Vec3[]): Float32Array {
   return array;
 }
 
-function getEarcutPoints(points: Vec3[]): Vec3[] {
+function getEarcutPoints(points: Vec3[]): (Vec3 | Point)[] {
```
jtbandes: Might help to use `$ReadOnlyArray<Vec3|Point>` elsewhere to avoid having to change this type.

PR #309, opened 2020-01-04
Updated follow TF to include Map

cjds:
## Summary

When I was viewing bags, I couldn't see the bags in the map frame. I was looking through the functions and I found a bug in the `getDescendants` function. It was ignoring the possibility of viewing the bag in the `rootTransformFrame`
image

## Test plan
Tested the function using `npm run docs` with and without bag files. This always traverses down the tree producing the list expected including the root TF
## Versioning impact
Not as far as I know

### DIFF
--- a/packages/webviz-core/src/panels/ThreeDimensionalViz/FollowTFControl.js
+++ b/packages/webviz-core/src/panels/ThreeDimensionalViz/FollowTFControl.js
@@ -89,13 +89,15 @@ type Props = {
   onFollowChange: (tfId?: string | false, followOrientation?: boolean) => void,
 };
 
-function* getDescendants(nodes: TfTreeNode[]) {
-  for (const node of nodes) {
-    if (node.tf.id !== getGlobalHooks().perPanelHooks().ThreeDimensionalViz.rootTransformFrame) {
-      yield node;
-    }
-    yield* getDescendants(node.children);
+function getDescendants(roots: TfTreeNode[]) {
+  const toVisit = [...roots]
+  const visited = []
+  while (toVisit.length >0){
+    const node = toVisit.pop();
+    visited.push(node);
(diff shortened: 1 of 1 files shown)


janpaul123:
This makes sense to me. Can you think of any reason not to do this @jtbandes or @brianc?

jtbandes:
I think the original idea was that "viewing the bag in the `rootTransformFrame`" is the same as "not following anything". When the button is off, that should mean the root frame is being used as the camera's origin. But maybe we should show it in the list anyway to make that more obvious?

janpaul123:
I do wonder, in this screenshot, there are multiple root frames, in which case it does seem useful to have them in the list. Because I think we don't just use this for following, but also for rendering, but I might be mistaken.

cjds:
I believe in the screenshot the robot was mislocalized at the start of the bag. So there was no association of the `/odom` frame with the `/map` frame which causes the two roots (solid I think on that one, I could go back and double check)

cjds:
@janpaul123 in other news could you help me with the CI ? Is it flaky?

It currently says:

```
Too long with no output (exceeded 10m0s): context deadline exceeded
```

But I don't think anything in this PR could have caused that. Or am I reading it wrong?

janpaul123:
Yeah it's a bit flakey since the free circleci machines are very slow. 😬 I'll take a look at this today. 👍

janpaul123:
This PR broke the tree representation a bit, but in the end the proper fixed turned out to be really simple, and I included it here: https://github.com/cruise-automation/webviz/pull/313/files#diff-b9112208bba7b2f237d9ef5fcedff919 This should get deployed onto webviz.io in an hour or so. Thanks for your contribution!!

cjds:
Yep that fix looks perfect. Thanks for the merge

PR #315, opened 2020-01-08
<GLText /> Add custom highlight color prop

troygibb:
## Summary

In order to further approximate native browser highlighting in `<GLText />`, we need to be able to specify a highlight color instead of the default yellow. This PR simply adds `highlightColor` as a prop to each `<GLText />` command. 

Note about performance: I haven't benchmarked this change quite yet, but will be doing so in my broader performance tasks in the next week. I was 90% done with this feature when we re-prioritized performance. 

custom-highlight-color-2

## Test plan

Added a new story which allows you to input custom colors.

## Versioning impact

Minor.


packages/regl-worldview/src/stories/GLText.stories.js
```diff
@@ -115,22 +160,60 @@ storiesOf("Worldview/GLText", module)
...
-          />
-        </>
+    return <Example />;
+  })
+  .add("custom highlight color", () => {
+    const Example = () => {
+      const [highlightColor, setHighlightColor] = useState<Color>({ r: 1, b: 0.5, g: 0.5, a: 1 });
+      const markers = textMarkers({ text: "Hello\nWorldview" }).map((marker) => ({
+        ...marker,
+        highlightedIndices: [0, 1, 2, 3, 4],
```
janpaul123: As said in person, I'd change this list a bit to make sure we actually test this properly.

packages/regl-worldview/src/stories/GLText.stories.js
```diff
@@ -41,6 +42,50 @@ function textMarkers({
   });
 }
 
+// Originally from: https://css-tricks.com/converting-color-spaces-in-javascript/
```
vidaaudrey: Maybe we can add [tinycolor2](https://www.npmjs.com/package/tinycolor2)
  as dependency since we use in webviz-core anyhow?

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -241,6 +249,7 @@ function makeTextCommand() {
         billboard: (ctx, props) => ({ buffer: props.billboard, divisor: 1 }),
         foregroundColor: (ctx, props) => ({ buffer: props.foregroundColor, divisor: 1 }),
         backgroundColor: (ctx, props) => ({ buffer: props.backgroundColor, divisor: 1 }),
+        highlightColor: (ctx, props) => ({ buffer: props.highlightColor, divisor: 1 }),
```
vidaaudrey: With this change, maybe we could remove `enableHighlight` React prop since we can tell by the presence of `highlightColor` prop?  In addition, maybe add an `autoHighlight` prop which computes the contrasted color based on the text marker's color so that the consumers don't have to manually compute highlight colors?

packages/regl-worldview/src/stories/GLText.stories.js
troygibb: oh cool! did not know we used this lib -- will refactor.

packages/regl-worldview/src/commands/GLText.js
troygibb: `enableHighlight` is a webgl prop (not react) -- the only prop react is passing is the indices in the string it needs to highlight. I may be able to co-op the existence of the highlight color in the frag shader, not sure if GLSL has a good mechanism for truthiness, slash not sure if this will make things more readable. 

RE: auto-computing highlight colors -- that's an interesting idea! I'll take a look and see if it's an easy enough lift for this PR

troygibb: > RE: auto-computing highlight colors -- that's an interesting idea! I'll take a look and see if it's an easy enough lift for this PR

I forgot to mention -- by default I set the font color (`foregroundColor`) to black when it's highlighted, so in a sense it is autocomputed already for consumers. This is how it is done for web pages in Chrome when you `control - f`, so I think I'd like to keep it the way it is. Lemme know what you think!

packages/regl-worldview/src/stories/GLText.stories.js
```diff
@@ -115,22 +117,60 @@ storiesOf("Worldview/GLText", module)
       });
 
       return (
-        <>
-          <div style={{ width: "100%", height: "95%" }}>
+        <div style={{ width: "100%", height: "100%" }}>
```
vidaaudrey: nit: could put the search ui on top of the Worldview, e.g. 
```
<div style={{width: '100vw', height: '100vh'}}>
<Container>
  <div>... your input ui</div>
</Container>
</div>
```

packages/regl-worldview/src/stories/GLText.stories.js
```diff
@@ -115,22 +117,60 @@ storiesOf("Worldview/GLText", module)
...
+              <GLText>{markers}</GLText>
+              <Axes />
+            </Container>
+          </div>
+          <div style={{ position: "absolute", top: "10px", right: "10px" }}>
+            <label htmlFor="highlight-color">Highlight Color: </label>
+            <input
+              type="color"
+              name="highlight-color"
+              value={tinyColor.fromRatio(highlightColor).toHexString()}
```
vidaaudrey: ❤️

PR #316, opened 2020-01-08
Add bz2 decompression support with compressjs

jtbandes:
We already proved that [compressjs](https://github.com/cscott/compressjs) works as a bz2 decoder in the rosbag.js tests: https://github.com/cruise-automation/rosbag.js/blob/68526faa242bb76943b6e42fa452f1eadb6cbc73/src/bag.test.js#L175-L178 But it wasn't imported in webviz, so bz2-compressed bags (such as the ones at https://vision.in.tum.de/data/datasets/rgbd-dataset/download#freiburg3_cabinet) didn't work by default.

It was almost very straightforward to add, except that a webpack/AMD module issue (https://github.com/webpack/webpack/issues/5316) prevents us from actually using compressjs in our webpack build. I worked around it by using a `string-replace-loader` to just remove the offending line. It's a pretty horrible hack, but it works...let me know if you have any better suggestions.


jtbandes:
I'm not sure if there's a good structured way to do performance testing on this. But also I think slow bz2 support is probably better than no bz2 support. If you're happy with this without further testing, feel free to merge!

janpaul123:
Cool I'll merge this when I also have time to pull this code into our internal version. 👍

Issue #312, opened 2020-01-06
Live viewer bogs down after time and data


defied:
Currently, while running webviz connected to a rosbag websocket, the data collected remains on screen from time of start and beings to slow the web display down as time increases.

It would be nice to have an option for live telemetry allowing the time scale to slide along a specified range of time, reducing the amount of data being kept.

janpaul123:
That's a great idea! Which panels in particular do you have this issue with?

defied:
It's most easily recognized with the plotters, as they maintain the full history. I haven't tried the others, including raw data.

janpaul123:
Cool, yeah, makes sense. Most panels keep a fairly limited set of data, but the Plot panel currently always keeps an indefinite set. I believe that the rosout panel also keeps a large amount of data, but not unlimited. The 3d panel keeps an indefinite set if you just keep adding markers and never delete them, but that's in your control to fix. 

So perhaps if we just add a maximum number of datapoints to the Plot panel, that might be enough?

defied:
And possibly make it switchable? It's nice in small rosbags to see the whole trail of data, but when you get larger bags, it would bog down just like live data.

defied:
Although, I guess if you let us set max datapoints, we could do that ourselves.

orshefi:
Is there any progress with a solution for this issue? limit X axis entries?

troygibb:
Check out the plot panel settings:

![Screen Shot 2020-07-28 at 8 56 02 AM](https://user-images.githubusercontent.com/18633748/88690273-54939100-d0b0-11ea-80fb-bc70f4af4abb.png)

There's an `X range` field for at least limiting the # of x axis entries from a visual standpoint.

As far as performance is concerned, we still don't have a way of limiting the number of points in the Plot panel. We've been shipping a set of performance improvements lately though, I'd try the latest release of Webviz if you haven't already! Let us know if you're still running into issues.

PR #317, opened 2020-01-09

janpaul123:
Changelog:
- Better error message when bag messages are not sorted.
- Allow limiting the number of messages in the Plot panel. #312
- Fix bug in TF follow button in 3d panel.
- More progress on new topic picker in 3d panel.
- Bunch more tiny fixes.

### DIFF
--- a/packages/webviz-core/package-lock.json
+++ b/packages/webviz-core/package-lock.json
@@ -50,6 +50,11 @@
         }
       }
     },
+    "@cruise-automation/hooks": {
+      "version": "0.0.1",
+      "resolved": "https://registry.npmjs.org/@cruise-automation/hooks/-/hooks-0.0.1.tgz",
+      "integrity": "sha512-sUwYShqRJAyg2XS7+uc9ZwdNkRArLdyjPTF6Ebl1GphXHX5QYU14aMt3SO3QtVg0fd/KSWM4FinRosxZDq5uZg=="
+    },
     "@mdi/svg": {
       "version": "3.2.89",
       "resolved": "https://registry.npmjs.org/@mdi/svg/-/svg-3.2.89.tgz",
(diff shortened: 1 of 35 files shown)


defied:
Checked this out. Looks great. I just wonder if the 'scaling window' idea should be added in this one, or should another issue be requested? 

By scaling window, I mean a window where I can say I only want to see the past <n> time in the plot (As it appears to still display from the beginning of monitoring to the latest data received).

janpaul123:
This should be implementing a sliding window. You will only see the latest N messages in the plot panel.

defied:
Okay, I'm not seeing that. I've tried refreshing, I see the start of the plot as 0.8 seconds. The end of the plot is over 100 seconds and still climbing. The beginning remains at 0.8.

defied:
image

janpaul123:
You have to set "N" in the panel settings :)

defied:
That's what I get for having it in the lower portion of my web browser. Lol!

Yeah, it's looking good now!

defied:
I vertically split a plot graph, changed the right panel type to a State Transitions page.
Suddenly all of the timing went weird.
Screenshot from 2020-01-09 16-38-25

Removed the State Transition panel and the plots resumed working as expected.

Split a plot vertically again and waited for the new plot screen to load, and there was no issues found.

Added (not split from plot) a new State Transitions and the same issue occured.

defied:
It happened again, but with only the plots. I sent data from the second topic on the plot, and when I stopped, it began to display the same unexpected result:

image

janpaul123:
Yeah we keep all the x-axes synced so you can compare them more easily. Maybe we should add a checkbox to toggle that, too? Or just add this same feature to the State Transitions panel?

defied:
It looks like when that topic stopped, it switched to display that node in full (orange), and the blue node was still only carrying my specified message N (400).

janpaul123:
Haha yeah, in our systems it's not very common for topics to stop publishing. Is this something that happens often? We can also consider only keeping the last M seconds, instead of the last N messages. Or to have the N messages spread out in total over the different topics (right now it's per-topic).

defied:
It does happen for development and testing, yes. Just hopefully not in the real world. 0]

defied:
I think time would make more sense.
I only want to see the last N seconds, ideally.

janpaul123:
I just realised that this doesn't only break when you have a topic that stops publishing, but also when you have topics that publish at different rates. So I agree that the time-based solution is the better one. Unfortunately it's also a bit harder to implement with the current architecture. Luckily I'm currently working on changing the architecture a bit, which should allow this to be done, but it might take a few weeks, so stay tuned. I'm not going to release this change since I think it's too broken. 😄

PR #326, opened 2020-01-14
Worldview: fix hitmap bug

davidswinegar:
I'm seeing an error in our logs: "(regl) can not resize a framebuffer
which is currently in use" in the readHitmap function. Fix by debouncing
the readHitmap function because the framebuffer can only be used in
one function at a time. Also add caching to our readHitmap function: if
we haven't repainted or changed the pixel that we're looking at, we return
the previous result.

Test plan: test added for queuePromise for caching. Testing that caching
actually works (an implementation detail) seems a bit difficult right now
given our test harness.

Release plan: minor release required

### DIFF
--- a/packages/regl-worldview/src/WorldviewContext.js
+++ b/packages/regl-worldview/src/WorldviewContext.js
@@ -9,6 +9,7 @@
 import debounce from "lodash/debounce";
 import * as React from "react";
 import createREGL from "regl";
+import shallowequal from "shallowequal";
 
 import { camera, CameraStore } from "./camera/index";
 import Command from "./commands/Command";
@@ -84,6 +85,10 @@ export class WorldviewContext {
   _drawCalls: Map<React.Component<any>, DrawInput> = new Map();
   _paintCalls: Map<PaintFn, PaintFn> = new Map();
   _hitmapObjectIdManager: HitmapObjectIdManager = new HitmapObjectIdManager();
+  _cachedReadHitmapCall: ?{
+    arguments: Array<any>,
+    result: Array<[MouseEventObject, Command<any>]>,
+  } = null;
   // store every compiled command object compiled for debugging purposes
   reglCommandObjects: { stats: { count: number } }[] = [];
   counters: { paint?: number, render?: number } = {};
@@ -207,6 +212,7 @@ export class WorldviewContext {
     if (!this.initializedData) {
       return;
     }
+    this._cachedReadHitmapCall = null; // clear the cache every time we paint
     const { regl, camera } = this.initializedData;
     this._clearCanvas(regl);
     camera.draw(this.cameraStore.state, () => {
@@ -232,6 +238,21 @@ export class WorldviewContext {
     ): Promise<Array<[MouseEventObject, Command<any>]>> => {
       if (!this.initializedData) {
         return new Promise((_, reject) => reject(new Error("regl data not initialized yet")));
+      }
+      const args = [canvasX, canvasY, enableStackedObjectEvents, maxStackedObjectCount];
+
+      const cachedReadHitmapCall = this._cachedReadHitmapCall;
+      if (cachedReadHitmapCall) {
+        if (shallowequal(cachedReadHitmapCall.arguments, args)) {
+          // Make sure that we aren't returning the exact object identity of the mouseEventObject - we don't know what
+          // callers have done with it.
+          const result = cachedReadHitmapCall.result.map(([mouseEventObject, command]) => [
+            { ...mouseEventObject },
+            command,
+          ]);
+          return Promise.resolve(result);
+        }
+        this._cachedReadHitmapCall = null;
       }
 
       const { regl, camera, _fbo } = this.initializedData;
@@ -326,6 +347,10 @@ export class WorldviewContext {
               // eslint-disable-next-line no-unmodified-loop-condition
             } while (currentObjectId !== 0 && enableStackedObjectEvents);
 
+            this._cachedReadHitmapCall = {
+              arguments: args,
+              result: mouseEventsWithCommands,
+            };
             resolve(mouseEventsWithCommands);
           });
         });
--- a/packages/regl-worldview/src/WorldviewContext.js
+++ b/packages/regl-worldview/src/WorldviewContext.js
@@ -86,9 +86,9 @@ export class WorldviewContext {
   _paintCalls: Map<PaintFn, PaintFn> = new Map();
   _hitmapObjectIdManager: HitmapObjectIdManager = new HitmapObjectIdManager();
   _cachedReadHitmapCall: ?{
-    arguments: Array<any>,
+    arguments: any[],
     result: Array<[MouseEventObject, Command<any>]>,
-  } = null;
+  } = undefined;
   // store every compiled command object compiled for debugging purposes
   reglCommandObjects: { stats: { count: number } }[] = [];
   counters: { paint?: number, render?: number } = {};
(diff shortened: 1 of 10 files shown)


packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -222,12 +223,12 @@ export class WorldviewContext {
 
   _debouncedPaint = debounce(this.paint, 10);
 
-  readHitmap(
+  readHitmap = queuePromise((
```
vidaaudrey: What do you think about dropping some early promises if the queued promises get too large or if the waiting time is too long? Could get a lot of queued promises if the user is rapidly clicking stacked objects in a changing scene.

packages/regl-worldview/src/utils/queuePromise.test.js
```diff
@@ -0,0 +1,79 @@
...
+  });
+  promise.resolve = resolve;
+  promise.reject = reject;
+  return promise;
+}
+
+/* eslint-disable jest/valid-expect-in-promise */
+
+describe("queuePromise", () => {
+  it("queues with resolved and rejected promises", async () => {
```
vidaaudrey: Nice test coverage

packages/regl-worldview/src/WorldviewContext.js
davidswinegar: We can't drop early promises because every `readHitmap` function has to resolve, and they could have different arguments. However, what I could add here is a cache of the `readHitmap` result that is busted every time there are new arguments or commands updates.

packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -83,6 +85,10 @@ export class WorldviewContext {
   _drawCalls: Map<React.Component<any>, DrawInput> = new Map();
   _paintCalls: Map<PaintFn, PaintFn> = new Map();
   _hitmapObjectIdManager: HitmapObjectIdManager = new HitmapObjectIdManager();
+  _cachedReadHitmapCall: ?{
+    arguments: Array<any>,
```
janpaul123: #nit I think `any[]` is a bit more canonical?

packages/regl-worldview/src/stories/assertionStories/Worldview.stories.js
```diff
@@ -234,4 +238,80 @@ stories
...
+        const result1 = await getTestData();
+        expect(result1.objects[0].object).toEqual(cube);
+        await emitMouseEvent("mousedown");
+        const result2 = await getTestData();
+        expect(result2.objects[0].object).toEqual(cube);
+      },
+    })
+  )
+  .add(
+    `(cached hitmap test) repainting busts the cache`,
```
vidaaudrey:
```suggestion
    `(cached hitmap test) Repainting busts the cache`,
```

packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -83,6 +85,10 @@ export class WorldviewContext {
...
+    result: Array<[MouseEventObject, Command<any>]>,
+  } = null;
```
janpaul123: #nit I prefer avoiding `null` since by default it's already `undefined` and it's weird that JS has multiple concepts for essentially the same thing, so might be simpler to just always use `undefined`.

packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -222,113 +229,134 @@ export class WorldviewContext {
...
-    const { width, height } = this.dimension;
+  readHitmap = queuePromise(
+    (
+      canvasX: number,
+      canvasY: number,
+      enableStackedObjectEvents: boolean,
+      maxStackedObjectCount: number
+    ): Promise<Array<[MouseEventObject, Command<any>]>> => {
+      if (!this.initializedData) {
+        return new Promise((_, reject) => reject(new Error("regl data not initialized yet")));
```
janpaul123:
```suggestion
        return Promise.reject(new Error("regl data not initialized yet"));
```

packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -222,113 +229,134 @@ export class WorldviewContext {
...
+      }
+      const args = [canvasX, canvasY, enableStackedObjectEvents, maxStackedObjectCount];
+
+      const cachedReadHitmapCall = this._cachedReadHitmapCall;
+      if (cachedReadHitmapCall) {
+        if (shallowequal(cachedReadHitmapCall.arguments, args)) {
+          // Make sure that we aren't returning the exact object identity of the mouseEventObject - we don't know what
+          // callers have done with it.
```
janpaul123: 👍

packages/regl-worldview/src/stories/assertionStories/Worldview.stories.js
```diff
@@ -234,4 +238,80 @@ stories
...
+      assertions: async (getTestData) => {
+        emitMouseEvent("mousedown");
+        await emitMouseEvent("mousedown");
+        const result1 = await getTestData();
+        expect(result1.objects[0].object).toEqual(cube);
+      },
+    })
+  )
+  .add(
+    `(cached hitmap test) A second event at the same point correctly selects the same object`,
```
janpaul123: Is there some way to assert that this actually uses the cache, instead of only testing that you get the right data? Not super big deal, but if there's an easy way then let's add that.

packages/regl-worldview/src/utils/queuePromise.js
```diff
@@ -0,0 +1,50 @@
...
+
+type QueuedFn = ((...args: Array<any>) => Promise<any>) & { currentPromise: ?Promise<any> };
+
+// Wait for the previous promise to resolve before starting the next call to the function.
+export default function queuePromise(fn: (...args: Array<any>) => Promise<any>): QueuedFn {
+  // Whether we are currently waiting for a promise returned by `fn` to resolve.
+  let calling = false;
+  // Whether another call to the function was made while a call was in progress.
+  const nextCallsArguments = [];
+  const nextCallPromises = [];
```
janpaul123: Slightly nicer might be `const nextCalls: {| args: any[], promise: Promise<any> |}[] = [];` since that way it's impossible for these arrays to get out of sync.

packages/regl-worldview/src/utils/queuePromise.js
```diff
@@ -0,0 +1,50 @@
...
+      nextCallPromises.push(returnPromise);
+      return returnPromise;
+    }
+    return start(...args);
+  }
+
+  function start(...args) {
+    calling = true;
+
+    const promise = fn(...args).finally(() => {
```
janpaul123: Good call on using `finally` here instead of `then`!

packages/regl-worldview/src/utils/queuePromise.js
janpaul123: Also if you then call it `calls` instead of `nextCalls` you can get rid of `calling` (which becomes implicit through `calls.length > 0` as well as reducing the amount of branching in this function since everything would go through the same code path.

packages/regl-worldview/src/stories/assertionStories/Worldview.stories.js
davidswinegar: I looked into it and unfortunately it isn't easy to do.

packages/regl-worldview/src/utils/queuePromise.js
davidswinegar: I copied half of this straight from `debouncePromise` :)

PR #328, opened 2020-01-14
[WIP] chore: update rollup and some plugins

### LINKED ISSUE #320: Update rollup and plugins
Hello!

We have recently finished migrating rollup plugins into our new plugins [monorepo](https://github.com/rollup/plugins) and consolidated packages names (ie: `rollup-plugin-commonjs` -> `@rollup/plugin-commonjs`)

Would this sort of change be welcome? If so I'd love to make this contribution!

wuweiweiwu:
## Summary

partially fixes #320 as `babel` and `copy` are not migrated yet

- seems like the styled-components flow types are now installed and is breaking some typechecks
- noticed that packages are declaring their own different dependencies on flow-bin, as flow is configured from monorepo root, it should be shared and hoisted. I don't believe this is causing the type issue though

### Check list

- [ ] fix flow type checks for monorepo packages
- [ ] fix flow type checks for styled-components
- [ ] single flow-bin dependency

## Test plan

- build and test work

## Versioning impact

- AFAIK this does not impact the distribution of any packages


wuweiweiwu:
For follow up purposes I believe this is the culprit https://github.com/cruise-automation/webviz/pull/328/files#diff-20d78e7d6fbae9db27f62afefaae7334R336

doesn't seem like flow-bin @ 0.106.2 handles this correctly

PR #331, opened 2020-01-16
Basic fast 2D collision

smallfx:
## Summary
My team works on a tool that uses `webviz` to make a 2d top-down application.
We noticed the current mouse collision detection is sometimes laggy, especially when `hitmapOnMouseMove` is `true`. I think this is because `regl.readPixel` calls are blocking.
This change adds an optional "2D" top-down mode for collision detection which doesn't depend on rendering / reading from GPU.

This change currently requires some additional data from the implementing application side.
For example, if you want to interact with a circle in 2D, you might use webviz's `Sphere` command.
For each child object of `Sphere`, to enable collisions with a 2D circle, you would now require:
```js
{
  ...sphereChildProps,
  collisionData: {
    type: "circle",
    zLayer: desiredZLayer,
    point: [circlePointX, circlePointY],
    radius: desiredCircleRadius,
   }
}
```
And for polygons (both convex and concave):
```js
{
  ...polygonChildProps,
  collisionData: {
    type: "polygon",
    zLayer: desiredZLayer,
    points: [
      [point1.x, point1.y],
      ...etc
    ],
  }
}
```

Additionally, the root `Webviz` component now has an extra prop `collisionMode2D` to enable this functionality.

Part of the reason I chose to separate collision data and the points/scale that are used for rendering is that we sometimes want to change the interactions with items, such as making circles easier to interact with / click.
 `collisionData` takes care of this use case by allowing you to just set the collision area to whatever you need.
I also separated the `z` because these are just 2D planes which we want to selectively give precedence over each other. This amounts to just sorting the collided objects / events after collisions are calculated. "Selecting" the `z` layer from an arbitrary polygon point didn't feel right, so I thought this would be the clearest way to specify a z-ordering for the entire polygon plane. (We often just use the "first" item in a list of stacked collisions to know which one we want to interact with)

I also made some assumptions which i'm unsure of. Pulling the command / child data from the draw call data feels odd, though I wasn't sure where else I could effectively get that data from.

## Test plan
A lot of manual testing of expected mouse-point collision behavior and framerate improvements.

## Versioning impact
Minor


packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -226,12 +281,64 @@ export class WorldviewContext {
...
+        if (!children) {
+          return console.debug(`2d hitmap calculation skipped, no draw children`, drawInput);
+        }
+        if (getChildrenForHitmap) {
+          let hitmapChildren = getChildrenForHitmap(children, () => [], []);
+          if (!Array.isArray(hitmapChildren)) {
+            hitmapChildren = [hitmapChildren];
+          }
+
+          hitmapChildren.forEach((item) => {
```
vidaaudrey: Would be nice if you could make the api more generic so it handles more than spheres and polygons. What do you think about having a prop like `testIs2DCollided(item, {clickedX, clickedY, layerIndex})` and let the consumer decide however they'd like to handle the collision? 

We could also avoid adding `zLayer` by using `layerIndex`. This way, the consumer only need to add a single prop to be able to use the 2D collision detection feature.

smallfx: @vidaaudrey Didn't know about the `layerIndex` prop, yeah that works!

davidswinegar:
A couple issues we should figure out with this API:
* Does this belong inside Worldview? It seems as though it might be possible for this to exist outside of the library by storing object state separately + using the ray. We could expose a few more helper methods on the camera that allow you to do this. Right now, the Worldview library doesn't have any logic around where elements actually are in relation to each other and ideally we wouldn't introduce it.
* I agree with Audrey's point if we are to do this that we probably want to just take a function that hit tests every item. This is likely to be very slow if there are a lot of items - does your usecase have a lot of rendered items (> 100)? There are alternative ways of storing objects that are more efficient but they'd require more work to implement.
* Right now, this API is triggered by a prop. We should explore some alternatives, like only turning this on if the camera is actually in a 2d space. We have to decide which plane this applies to as well - I assume we want it to apply to the XY plane.
* Are there alternatives we can consider? For example - if the issue is that we're running into slow speeds when panning/zooming, we could disable hitmap during those times. If we're running into slow speeds when the camera and all elements in the scene are static, we could cache the entire hitmap framebuffer and do hit tests on the cached framebuffer instead of painting, or we could read all the pixels into memory using some bitmap thing and read off of the saved bitmap in memory. @brianc might have more ideas

smallfx:
@davidswinegar Great points!
1) This was my main concern for sure. I don't really think this code belongs with the rendering stuff, though that is where `readHitmap` was to begin with, so I just sort of followed suit... I am not really sure where else I can place this stuff such that I can get the "same" scene data within Worldview (and maybe the answer is to not put it in Worldview).

2) Right, my initial thinking was that doing all the collisions in "one place" rather than having a function passed in with the data is that it eventually allows for some kind of sparse grid / "spatial hashing" collision check culling against the whole data set. I was going to implement this, though I felt like I needed to get some feedback on this hacky version before making any decisions that would take longer to implement.

3) Enabling this when the camera is in 2D mode without needing an additional prop would be ideal! I guess there must already be some sort of prop that says the camera is orthogonal, pointed down along the scene's Z axis?

4) I might be missing something, though we need hover state while objects are moving (being dragged) and also while they are stationary. It seems like disabling hit detection on mouse move wouldn't enable us to have that.
Also, the framebuffer cache would be sort of a large cache in some cases, right? For the case when there are, for example, 100 objects (so far this is not the case for our application, but) you would need to cache 100 framebuffers and call `regl.readPixel` for each of them on mouse move (when we need hover for the new coordinates)... Unfortunately it seems like `regl.readPixel` was the slowest part.
Perhaps disabling stacked collision detection is the key to this... though we also have a lot of cases where we need event propagation if i'm not mistaken.

PR #336, opened 2020-01-30
Add static transform (/tf_static) to 3DViz Panel

lwohlhart:
## Summary
The visualisation of the TF tree currently doesn't include static transformations published at the /tf_static topic.
Since many robot descriptions (and also environment models) are built using this concept it is necessary to include these transformations.

Therefore I simply added the messages received on /tf_static to the transform consumer in the ThreeDimensionalViz panel.

## Test plan

The change can be tested and verified replaying the example.bag file from packages/webviz-core/public/fixtures/
The 'turtle1' frame actually has a 'carrot' frame statically attach to it at 1 meter in x direction.

Prior to the change the carrot frame doesn't show up at all. 
Now the frame is available an correctly linked to its parent.

## Versioning impact

Most probably this can be handled as a patch update. Or maybe a minor if your versioning policy requires a minor for a new 'feature'.

Thank you for the nice tool!


packages/webviz-core/src/panels/ThreeDimensionalViz/withTransforms.js
```diff
@@ -44,6 +44,16 @@ function withTransforms<Props: *>(ChildComponent: React.ComponentType<Props>) {
           }
         }
       }
+      const tfs_static = frame[TRANSFORM_STATIC_TOPIC];
+      if (tfs_static) {
+        for (const msg of tfs_static) {
+          for (const tf of msg.message.transforms) {
+            if (tf.child_frame_id !== getGlobalHooks().perPanelHooks().ThreeDimensionalViz.skipTranformFrame) {
```
janpaul123: We probably don't need this line, since we only need this for `/tf` for our internal version. Better to keep this complication out of this new code.

lwohlhart: I see, I'll get rid of it in a moment
Anyways if I look at it correctly there's a typo in the other line  ^^ (skipTranformFrame is missing an 's' I guess) so this should be undefined either way

janpaul123:
@lwohlhart ah I totally forgot about this. CI seems to be failing because it's a PR to a fork, but that's OK. I'll pull this into our internal repo and then I'll soon do a sync to push this out to this open source repo again. Thanks!

janpaul123:
I included these changes in #396. Thanks again!

lwohlhart:
yay thx !
😃

PR #340, opened 2020-02-07
Update jest config

davidswinegar:
Add this line, which is required to get internal CI to pass so that
we can add this worker to the open-source version.

### DIFF
--- a/jest/jest.config.js
+++ b/jest/jest.config.js
@@ -25,6 +25,7 @@ module.exports = {
   ],
   setupTestFrameworkScriptFile: "<rootDir>/packages/webviz-core/src/test/setupTestFramework.js",
   moduleNameMapper: {
+    "worker-loader!./ChartJSWorker": "<rootDir>/webviz-core/src/test/MockWorker.js",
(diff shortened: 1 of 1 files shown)


jest/jest.config.js
```diff
@@ -25,6 +25,7 @@ module.exports = {
   ],
   setupTestFrameworkScriptFile: "<rootDir>/packages/webviz-core/src/test/setupTestFramework.js",
   moduleNameMapper: {
+    "worker-loader!./ChartJSWorker": "<rootDir>/webviz-core/src/test/MockWorker.js",
```
janpaul123: Should we do this by default for `worker-loader!*`?

jtbandes: Why is this path different from the others around it? (missing `/packages/`)

janpaul123: Ah yeah, didn't notice that; that seems wrong! 😮

davidswinegar: Whoops, didn't notice these comments before merging. Fixed with https://github.com/cruise-automation/webviz/pull/343

PR #342, opened 2020-02-10
Fix worker loader

troygibb:
## Summary

Revisiting this old issue that I couldn't figure out a few months ago. After loading up https://webviz.io/app/?demo, I saw this error:

Screen Shot 2020-02-10 at 11 12 43 AM

Turns out our worker-loader was scooping up `*.worker.js` files in node_modules, and was doing a double-require (hence the syntax error shown above).

Solution is to just exclude `node_modules` from the worker-loader. Pretty straightforward, and mimics our private webpack.config.js.

## Test plan

Tested with `npm run docs`, and should pass CI.

## Versioning impact

Only effects webviz-core -- no version bump needed.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -87,6 +87,7 @@ module.exports = {
       },
       {
         test: /\.worker\.js$/,
+        exclude: /node_modules/,
         use: {
           loader: "worker-loader",
           options: { name: "[name].[ext]?[hash]" },


troygibb:
> Thanks! Any good way we can test this? Probably not since we don't have integration tests yet in this repo?

Existing screenshots should be a pretty good proxy for an integration test in this case. Still quite flakey tho :(

PR #349, opened 2020-03-06
<GLText> improvements for text rendering

hhsaez:
## Summary

**Note: This PR is no longer valid. It has been replaced by #360**

This PR improves Worldview's existing GLText command, adding new functionalities for enhanced font rendering and scale invariance. 

### Higher resolution font
GLText now uses a higher resolution font atlas, which increases the resolution of the SDF texture, avoiding visual artifacts when zooming in/out the text:

**Before:**
<image src="https://user-images.githubusercontent.com/1727802/76123182-d3c2e400-5fac-11ea-9788-47758d12f919.png" height="300px" />

**Now:**
<image src="https://user-images.githubusercontent.com/1727802/76123180-d32a4d80-5fac-11ea-89c6-bfc634bf5809.png" height="300px" />

### Scale Invariance
Keeps the text at the same size (relative to the screen resolution) regardless of the zoom level in the 3D world. It produces a final result similar as when using the <Text> command (which renders text using HTML elements on top of the 3D video).

**Note:** scaleInvariant does not work when the billboard option is false for a marker.

Usage: 
`<GLText scaleInvariantFontSize="40">{markers}</GLText>`

The scaleInvariantFontSize option allows you to change the font size for the text when scale invariant is active. 

Because of how the SDF approach works, visual artifacts may appear when the text size is too small. In such cases, I'm skipping smoothing the distance value from the SDF texture which produce better results. This is handled by fragment shader.

I also updated the documentation with the new *scaleInvariantFontSize* option

## Test plan
Added several stories to showcase the new options and compare their output with the default implementation:

* *high resolution text*: Renders GLText at a higher zoom level to showcase the higher resolution font. 
* *scaleInvariant*: Shows how GLText renders text using the scaleInvariantFontSize. Three tests are provided: 
  * *scaleInvariant*: Font size is 10px. Does not use smooth interpolation.
  * *scaleInvariant-20*: Font size is 20px. Does not use smooth interpolation.
  * *scaleInvariant-40*: Font size is 40px. Since this is a bigger font, it will use smooth interpolation in the fragment shader.
* *autoBackgroundColor scaleInvariant*: Same as *autoBackgroundColor* story, but using the scaleInvariant option
* *changing text*: Updated version that avoid unpredictable results and allows you to change the text manually.

## Versioning impact
Minor

### DIFF
--- a/docs/src/4.15.GLText.mdx
+++ b/docs/src/4.15.GLText.mdx
@@ -7,3 +7,9 @@ import { GLText } from "./jsx/allLiveEditors";
 The prop types are identical to `<Text />`, except that additionally each Text object supports an optional `billboard?: boolean` property. If true (the default), text will always face the camera; if false, text may be rotated in 3D according to `pose.orientation`.
 
 <GLText />
+
+## Scale Invariance
+
+`<GLText />` also supports rendering text at a constant scale regardless of the zoom level by using the optional `scaleInvariantFontSize?: number` property. If set (default is `undefined`), the text will be rendered always at the same size, resembling the behavior of the `<Text />` command. 
+
+Please note that `scaleInvariantFontSize` only works if the `billboard` property is set to true.
--- a/docs/src/4.15.GLText.mdx
+++ b/docs/src/4.15.GLText.mdx
@@ -1,4 +1,4 @@
-import { GLText } from "./jsx/allLiveEditors";
+import { GLText, GLTextScaleInvariant } from "./jsx/allLiveEditors";
 
 # Text
 
@@ -13,3 +13,5 @@ The prop types are identical to `<Text />`, except that additionally each Text o
 `<GLText />` also supports rendering text at a constant scale regardless of the zoom level by using the optional `scaleInvariantFontSize?: number` property. If set (default is `undefined`), the text will be rendered always at the same size, resembling the behavior of the `<Text />` command. 
 
 Please note that `scaleInvariantFontSize` only works if the `billboard` property is set to true.
+
+<GLTextScaleInvariant />
--- a/packages/regl-worldview/src/commands/GLText.js
+++ b/packages/regl-worldview/src/commands/GLText.js
@@ -49,15 +49,15 @@ type Props = {
 
 type FontAtlas = {|
   textureData: Uint8Array,
-  textureWidth: number,
-  textureHeight: number,
-  charInfo: {
-    [char: string]: {|
-      x: number,
+    textureWidth: number,
+      textureHeight: number,
+        charInfo: {
+  [char: string]: {|
+    x: number,
       y: number,
-      width: number,
+        width: number,
     |},
-  },
+},
 |};
 
 // Font size used in rendering the atlas. This is independent of the `scale` of the rendered text.
@@ -461,6 +461,12 @@ export default function GLText(props: Props) {
   const context = useContext(WorldviewReactContext);
   const { dimension } = context;
   const scaleInvariantFontSize = props.scaleInvariantFontSize || DEFAULT_SCALE_INVARIANT_SIZE;
+
+  // Compute the actual size for the text object in NDC coordinates (from -1 to 1)
+  // In order to make sure the text is always shown at the same size regardless of
+  // the canvas dimensions (as Text does), we need to scale it based on both the desired 
+  // font size and the current worldview resolution. Otherwise, the text will be 
+  // displayed at different sizes depending on the worlview canvas dimension.
   const scaleInvariantSize = scaleInvariantFontSize / dimension.height;
 
   const [command] = useState(() => makeTextCommand());
--- a/packages/regl-worldview/src/commands/GLText.js
+++ b/packages/regl-worldview/src/commands/GLText.js
@@ -43,6 +43,8 @@ type Props = {
   children: $ReadOnlyArray<TextMarkerProps>,
   autoBackgroundColor?: boolean,
   hiresFont?: boolean,
+  debugSDF?: boolean,
+  scaleInvariant?: boolean,
 };
 
 type FontAtlas = {|
@@ -129,6 +131,7 @@ const vert = `
   uniform mat4 projection, view, billboardRotation;
   uniform float fontSize;
   uniform vec2 atlasSize;
+  uniform bool scaleInvariant;
 
   // per-vertex attributes
   attribute vec2 texCoord;
@@ -165,16 +168,29 @@ const vert = `
     return v + (2.0 * cross(q.xyz, temp));
   }
 
+  vec4 computeVertexPosition(vec3 markerPos) {
+    vec3 pos;
+    if (billboard == 1.0) {
+      pos = (billboardRotation * vec4(markerPos, 1.0)).xyz + posePosition;
+    } else {
+      pos = rotate(markerPos, poseOrientation) + posePosition;
+    }
+    return projection * view * vec4(pos, 1.0);
+  }
+
   void main () {
     vec2 srcSize = vec2(srcWidth, fontSize);
     vec3 markerSpacePos = scale * vec3((destOffset + position * srcSize + alignmentOffset) / fontSize, 0);
-    vec3 pos;
-    if (billboard == 1.0) {
-      pos = (billboardRotation * vec4(markerSpacePos, 1)).xyz + posePosition;
-    } else {
-      pos = rotate(markerSpacePos, poseOrientation) + posePosition;
+    gl_Position = computeVertexPosition(markerSpacePos);
+
+    if (scaleInvariant) {
+      float screenScale = 0.015;
+      float w = gl_Position.w;
+      w *= screenScale;
+      markerSpacePos *= w;
+      gl_Position = computeVertexPosition(markerSpacePos);
     }
-    gl_Position = projection * view * vec4(pos, 1);
+
     vTexCoord = (srcOffset + texCoord * srcSize) / atlasSize;
     vEnableBackground = enableBackground;
     vForegroundColor = foregroundColor;
@@ -190,6 +206,7 @@ const frag = `
   uniform mat4 projection;
   uniform sampler2D atlas;
   uniform float cutoff;
+  uniform bool debugSDF;
 
   varying vec2 vTexCoord;
   varying float vEnableBackground;
@@ -214,7 +231,9 @@ const frag = `
       gl_FragColor.a *= step(1.0 - cutoff, dist);
     }
 
-    // gl_FragColor = vec4( dist, dist, dist, 1.0 );
+    if (debugSDF) {
+      gl_FragColor = vec4( dist, dist, dist, 1.0 );
+    }
 
     if (gl_FragColor.a == 0.) {
       discard;
@@ -240,6 +259,8 @@ function makeTextCommand() {
         atlasSize: () => [atlasTexture.width, atlasTexture.height],
         fontSize: command.fontSize,
         cutoff: CUTOFF,
+        debugSDF: command.debugSDF,
+        scaleInvariant: command.scaleInvariant,
       },
       instances: regl.prop("instances"),
       count: 4,
@@ -427,5 +448,7 @@ export default function GLText(props: Props) {
   // so just attach it to the command object for now.
   command.autoBackgroundColor = props.autoBackgroundColor;
   command.fontSize = fontSize;
+  command.debugSDF = props.debugSDF === true;
+  command.scaleInvariant = props.scaleInvariant === true;
   return <Command reglCommand={command} {...props} />;
 }
--- a/packages/regl-worldview/src/commands/GLText.js
+++ b/packages/regl-worldview/src/commands/GLText.js
@@ -42,23 +42,25 @@ type Props = {
   ...CommonCommandProps,
   children: $ReadOnlyArray<TextMarkerProps>,
   autoBackgroundColor?: boolean,
+  hiresFont?: boolean,
 };
 
 type FontAtlas = {|
   textureData: Uint8Array,
-  textureWidth: number,
-  textureHeight: number,
-  charInfo: {
-    [char: string]: {|
-      x: number,
+    textureWidth: number,
+      textureHeight: number,
+        charInfo: {
+  [char: string]: {|
+    x: number,
       y: number,
-      width: number,
+        width: number,
     |},
-  },
+},
 |};
 
 // Font size used in rendering the atlas. This is independent of the `scale` of the rendered text.
-const FONT_SIZE = 40;
+const DEFAULT_FONT_SIZE = 40;
+const HIRES_FONT_SIZE = 160;
 const MAX_ATLAS_WIDTH = 512;
 const SDF_RADIUS = 8;
 const CUTOFF = 0.25;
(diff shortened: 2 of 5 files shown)


packages/regl-worldview/src/commands/GLText.js
```diff
@@ -320,11 +358,15 @@ function makeTextCommand() {
           marker.colors?.[1] || (command.autoBackgroundColor && isColorDark(fgColor) ? BG_COLOR_LIGHT : BG_COLOR_DARK);
         const hlColor = marker?.highlightColor || { r: 1, b: 0, g: 1, a: 1 };
 
+        if (!marker.billboard && command.scaleInvariant) {
+          console.warn("Scale invariant option is only supported for billboard markers");
```
janpaul123: I'd be more aggressive and just throw an error, so that we're really sure people don't do this:

```suggestion
          throw new Error("Scale invariant option is only supported for billboard markers");
```

packages/regl-worldview/src/stories/GLText.stories.js
```diff
@@ -91,6 +170,30 @@ storiesOf("Worldview/GLText", module)
...
+      const [text, setText] = useState("Hello\nWorldview");
+      useLayoutEffect(() => {
+        let i = 0;
+        const id = setInterval(() => {
+          setText(`New text! ${++i}`);
+          if (inScreenshotTests()) {
+            clearInterval(id);
+          }
+        }, 100);
+        return () => clearInterval(id);
```
janpaul123: We automatically take screenshots of these stories in CI, after 200ms, which we use to catch regressions; see https://github.com/cruise-automation/webviz/pull/349/files#diff-c846e28069eda40e4003933c8306d5ffR47. So this would lead to a flakey screenshot test.

It might be better to just add a button or text field or so to change the text. It would still be nice to do one text change automatically though, so that the screenshot test also captures that changing text works properly.

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -416,9 +458,17 @@ function makeTextCommand() {
 }
 
 export default function GLText(props: Props) {
+  const context = useContext(WorldviewReactContext);
+  const { dimension } = context;
+  const scaleInvariantFontSize = props.scaleInvariantFontSize || DEFAULT_SCALE_INVARIANT_SIZE;
+  const scaleInvariantSize = scaleInvariantFontSize / dimension.height;
```
janpaul123: Why exactly is the `scaleInvariantSize` dependent on the height of Worldview? That's not totally obvious to me. Might be worth a comment perhaps?

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -42,6 +42,9 @@ type Props = {
   ...CommonCommandProps,
   children: $ReadOnlyArray<TextMarkerProps>,
   autoBackgroundColor?: boolean,
+  hiresFont?: boolean,
```
janpaul123:
```suggestion
  highResolutionFont?: boolean,
```
perhaps? We don't have to pay github by the character. 😉

janpaul123:
This is looking really great! I'd like also a review from @troygibb or @jtbandes or @brianc who know this code better.

Some more thoughts:
- Would there be a downside to just always do the high resolution text rendering? In my opinion, the fewer properties the better, so I'd only like to add a property to configure high resolution on or off is there is truly a compelling reason to not use it. Otherwise, let's just always use high resolution text, and everyone will benefit from it!
- When we're happy with the API, be sure to also add some docs here: https://github.com/cruise-automation/webviz/blob/master/docs/src/jsx/commands/GLText.js That will show up on this page: https://webviz.io/worldview/#/docs/commands/gltext You should be able to run that locally (see top-level readme).

packages/regl-worldview/src/commands/GLText.js
hhsaez: @janpaul123 I agree about having a more aggressive approach here. While the code is handling this case and just ignoring scale invariance for non billboards, a user will only be aware of this when looking at the debugger's console. If we thrown an error instead, the problem is much more visible and easier to fix.
What do you think @troygibb?

packages/regl-worldview/src/stories/GLText.stories.js
hhsaez: @janpaul123 I copy this story from the one above (see "changing text") and just added the "hires" option. I'm not sure I'm understanding what should be the correct approach. Should I update the behavior for "changing text" story as well?

packages/regl-worldview/src/commands/GLText.js
hhsaez: @janpaul123 I agree, it's not obvious. I just added this comment to the code that explains it:

`Compute the actual size for the text object in NDC coordinates (from -1 to 1)
In order to make sure the text is always shown at the same size regardless of the canvas dimensions (as Text does), we need to scale it based on both the desired font size and the current worldview resolution. Otherwise, the text will be displayed at different sizes depending on the worlview canvas dimension.`

packages/regl-worldview/src/commands/GLText.js
hhsaez: @janpaul123 Changed it to highResolutionFont in a recent commit :)

packages/regl-worldview/src/commands/GLText.js
troygibb: sounds good to me!

MatthewSteel: Probably doesn't matter, but is it reasonable to enforce this strucutrally? Maybe signal `billboard` with a nullable object containing billboard-relevant options like scale invariance?

janpaul123: Ah that's a great idea, e.g. ```js
type Props = {|
  …other stuff…
  billboard: boolean | {| scaleInvariantFontSize?: number |}
|};
```

Because it probably never makes sense to use scale invariance without billboards.

File: packages/regl-worldview/src/stories/GLText.stories.js (hunk 2 above)
Comment (janpaul123) [reply packages/regl-worldview/src/stories/GLText.stories.js]: Ah yes, that one is bad too, would be good to fix that one as well.

Btw, you might have noticed that CI is a bit broken in that it often times out, but we have plans on fixing this soon-ish (we're considering using a different CI system so we're waiting on that decision).

File: packages/regl-worldview/src/commands/GLText.js [hunk 5]
```diff
@@ -416,9 +458,23 @@ function makeTextCommand() {
...
+  // displayed at different sizes depending on the worlview canvas dimension.
+  const scaleInvariantSize = scaleInvariantFontSize / dimension.height;
+
   const [command] = useState(() => makeTextCommand());
   // HACK: Worldview doesn't provide an easy way to pass a command-level prop into the regl commands,
   // so just attach it to the command object for now.
   command.autoBackgroundColor = props.autoBackgroundColor;
+  command.fontSize = props.highResolutionFont ?? props.scaleInvariant ? HIGH_RESOLUTION_FONT_SIZE : DEFAULT_FONT_SIZE;
+  command.scaleInvariant = props.scaleInvariant === true;
+  command.scaleInvariantSize = scaleInvariantSize;
```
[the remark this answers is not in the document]
janpaul123: Yeah exactly. We can combine this suggestion and @MatthewSteel's suggestion above to get a pretty compact API: https://github.com/cruise-automation/webviz/pull/349#discussion_r390029252

packages/regl-worldview/src/commands/GLText.js
hhsaez: We decided not to add new options to markers and remove the warning message instead.

packages/regl-worldview/src/stories/GLText.stories.js
hhsaez: @janpaul123 @troygibb I updated the changingText story to make it more predictable. It changes the text once when initializing. There's also a button to change the text manually when using storyboard

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -416,9 +458,23 @@ function makeTextCommand() {
...
+  // font size and the current worldview resolution. Otherwise, the text will be
+  // displayed at different sizes depending on the worlview canvas dimension.
+  const scaleInvariantSize = scaleInvariantFontSize / dimension.height;
+
   const [command] = useState(() => makeTextCommand());
   // HACK: Worldview doesn't provide an easy way to pass a command-level prop into the regl commands,
   // so just attach it to the command object for now.
   command.autoBackgroundColor = props.autoBackgroundColor;
+  command.fontSize = props.highResolutionFont ?? props.scaleInvariant ? HIGH_RESOLUTION_FONT_SIZE : DEFAULT_FONT_SIZE;
+  command.scaleInvariant = props.scaleInvariant === true;
```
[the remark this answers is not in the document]
hhsaez: I updated the code to use the shorthand :)

packages/regl-worldview/src/commands/GLText.js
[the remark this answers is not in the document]
hhsaez: I updated the code to have only the scaleInvariantFontSize property. FontSize is now a constant (using the high resolution value) and the scaleInvariant flag is set depending of the value of scaleInvariantFontSize

hhsaez:
@janpaul123 @troygibb @MatthewSteel Thanks for your comments. I updated my code based on what we discussed so far. Please take a look at it when you have the chance.

I also updated the documentation for GLText to include the new scaleInvariantFontSize option.

docs/src/4.15.GLText.mdx
```diff
@@ -7,3 +7,9 @@ import { GLText } from "./jsx/allLiveEditors";
 The prop types are identical to `<Text />`, except that additionally each Text object supports an optional `billboard?: boolean` property. If true (the default), text will always face the camera; if false, text may be rotated in 3D according to `pose.orientation`.
 
 <GLText />
+
+## Scale Invariance
+
+`<GLText />` also supports rendering text at a constant scale regardless of the zoom level by using the optional `scaleInvariantFontSize?: number` property. If set (default is `undefined`), the text will be rendered always at the same size, resembling the behavior of the `<Text />` command.
```
jtbandes: Sorry if this has been asked before; I haven't read through all the prior comments. Did you consider reusing the `scale.z` for this (as Text does; http://wiki.ros.org/rviz/DisplayTypes/Marker#View-Oriented_Text_.28TEXT_VIEW_FACING.3D9.29_.5B1.1.2B-.5D), rather than using a separate prop for `scaleInvariantFontSize`?

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -188,28 +206,41 @@ const frag = `
...
+
   void main() {
     float dist = texture2D(atlas, vTexCoord).a;
 
     // fwidth(dist) is used to provide some anti-aliasing. However it's currently only used
     // when the solid background is enabled, because the alpha blending and
     // depth test don't work together nicely for partially-transparent pixels.
     float edgeStep = smoothstep(1.0 - cutoff - fwidth(dist), 1.0 - cutoff, dist);
+
+    if (scaleInvariant && vBillboard == 1.0 && scaleInvariantSize < 0.03) {
```
jtbandes: Just curious if you have run into any precision issues with `vBillboard == 1.0`? I recall that while working on this originally I had some strange experiences with the `varying` interpolation producing values that were very close but not exactly 1.0, so just want to confirm that you aren't experiencing this.

docs/src/4.15.GLText.mdx
jtbandes: Hmm, additional thought — it seems like this design doesn't allow different text markers to be different sizes (while all being scale-invariant), is that correct?

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -416,9 +449,21 @@ function makeTextCommand() {
...
 export default function GLText(props: Props) {
+  const context = useContext(WorldviewReactContext);
+  const { dimension } = context;
+
   const [command] = useState(() => makeTextCommand());
   // HACK: Worldview doesn't provide an easy way to pass a command-level prop into the regl commands,
   // so just attach it to the command object for now.
   command.autoBackgroundColor = props.autoBackgroundColor;
+
+  command.scaleInvariant = !!props.scaleInvariantFontSize;
```
jtbandes: nit: what about `scaleInvariantFontSize={0}`? (Not that this would be desirable anyway)

docs/src/4.15.GLText.mdx
hhsaez: @jtbandes it does support different sizes based on the `scale` property for markers, but the size remains constant regardless of the zoom level.

packages/regl-worldview/src/commands/GLText.js
hhsaez: I'm not seeing precision issues. I prefer to use boolean values for flags (or avoid flags completely), but I didn't want to introduce more changes than needed.

jtbandes: Yes, I think I ran into an issue where regl / WebGL 1 don't support `uniform bool` in fragment shaders (or something like that)

packages/regl-worldview/src/commands/GLText.js
hhsaez: A value of 0 (zero) will disable scale invariance and a negative one will flip the Y coordinate. Like you said, I'm not sure if that's a valid use case, but the size effects are visible. Should I constraint the sizes to positive values only?

docs/src/4.15.GLText.mdx
jtbandes: Oh I see, so the scaleInvariantFontSize is multiplied by the scale? That makes sense, although I think the documentation isn't clear about this, maybe you can update to clarify more

packages/regl-worldview/src/commands/GLText.js
jtbandes: Or just use
```suggestion
  command.scaleInvariant = props.scaleInvariantFontSize != null;
```
here?

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -416,9 +449,21 @@ function makeTextCommand() {
...
+  // Compute the actual size for the text object in NDC coordinates (from -1 to 1)
+  // In order to make sure the text is always shown at the same size regardless of
+  // the canvas dimensions (as Text does), we need to scale it based on both the desired
+  // font size and the current worldview resolution. Otherwise, the text will be
+  // displayed at different sizes depending on the worlview canvas dimension.
+  command.scaleInvariantSize = (props.scaleInvariantFontSize ?? 0) / dimension.height;
```
davidswinegar: Can we add a test for changing the dimensions of the canvas after it's been rendered once to make sure that this still works?

docs/src/4.15.GLText.mdx
janpaul123: I think this would be greatly clarified by an example, e.g. in addition to the example above. E.g. using a couple of different text markers with different sizes, and some with `billboard: true` and some with `billboard: false`.

packages/regl-worldview/src/commands/GLText.js
hhsaez: @jtbandes I made the change you suggested. Now it handles the scaleInvariantFontSize properly.

docs/src/4.15.GLText.mdx
hhsaez: I included a live example GLText's doc that shows how to use scaleInvariantFontSize. It's not as pretty as the matrix one, but it shows how scale invariance behaves compared to a regular GLText object. I also included a Grid that hopefully helps visualize this behavior.

packages/regl-worldview/src/commands/GLText.js
hhsaez: Done :)

hhsaez:
I'm closing this PR in favor of #360

PR #354, opened 2020-03-10

janpaul123:
Changelog:
- Add "Download" layout button.
- Fixed the Plot panel not showing the delete icon when the panel is small.
- Fixed the Diagnostics Detail panel not linking to the proper value when clicking on a chart icon.
- Actually support reading unsorted bags
- Fixed Node Playground types with `sec` and `nsec` fields not being identified as of type "time".
- Add animated Webviz logo overlay until data is loaded. 
- Fixed an issue in the "Faster 3D Text" experimental feature where numeric text markers were not rendering.
- Fixed an issue in the Image panel where rogue markers were appearing.
- Fixed a bug where the image panel broke when you dragged it to a new position in the layout.
- Fix a bug where new image panels crash.
- Fixed a major issue in the Topic Groups feature where the layout config was getting cleared.
- Fixed State Transitions input field getting obscured by the panel toolbar.
- Added ability to set a custom diff topic in Raw Messages panel; other UI and usability improvements.
- Added button to the Diagnostics Detail panel for opening a State Transitions panel for the top-level message.
- Use [ctrl + f] (or [cmd + f] on Mac) with the "Faster 3D Text" experimental feature, to quickly navigate search results in the 3D Panel.
- Rosout panel now supports viewing data from /webviz_bag_2.
- Added ID keys to the summary labels of all diffed rows (e.g. `{} track_id: 5`) in the Raw Messages panel.
- Fixed filters not working properly in the Raw Messages panel.
- Sped up scrubbing back and forth when using a local bag.
- Prioritized diff topics with the same datatype as the base topic, in the Raw Messages panel's list of diff topic autocomplete options.
- Added config option in Raw Messages panel to show both the changed and unchanged fields of a message being diffed, not just the changed fields.
- Fixed autocomplete options from trailing off-screen in Raw Messages panel.
- Bumped minimum required Chrome version to 79.
- General performance improvements.
- Added ability to auto-sync all 3D panels from any 3D panel's camera settings.
- Added ability to diff consecutive messages for a given topic in the Raw Messages panel.
- Added ability to set custom topic path as x-axis for Plot panel.
- Fixed crashing topic settings for car and tracking topics when using topic tree.
- Fixed bug when overrideColor for marker settings is not set.
- Fixed bug in Raw Messages panel where topic paths longer than the base topic name were not displaying messages correctly.
- Hide `/webviz/**` topics that don't have any valid input topics from showing up in various topic selections.
- Fixed app to show an error instead of freezing when rendering large diagnostics messages.
- Local bag caching has been moved to the "Experimental Features" menu, and is disabled by default. This should result in shorter startup times, and (paradoxically) faster load times, as the browser's local caching APIs have shown to have unreliable performance. We only recommend turning this experimental feature back on when streaming data over a slow network. We're looking to make improvements to this feature in the future so we can turn it on by default again.
- Fixed a Raw Messages panel bug in which mousing off the autocomplete menu onto an adjacent Diagnostics Detail panel closes the menu.
- Fixed a bug in Node Playground where input topics were still being subscribed to even if they were commented out.
- Add `unlimitedMemoryCache` experimental feature to load everything in memory without regard for caching limits. Useful for when you have a machine with lots of memory and you're explicitly OK with Webviz taking up a lot of it. Use at your own risk; this might crash your browser!


janpaul123:
FYI @MatthewSteel this might conflict with some of your current work so best to rebase against this!

packages/webviz-core/package.json
```diff
@@ -10,10 +10,10 @@
     "@simonwep/pickr": "1.4.7",
     "@trust/webcrypto": "0.9.2",
     "antd": "3.26.5",
-    "chart.js": "2.7.3",
+    "chart.js": "davidswinegar/Chart.js#ab170d249db7cae3313d9e2515ef35283c9bb5e8",
     "chartjs-plugin-annotation": "0.5.7",
-    "chartjs-plugin-datalabels": "0.4.0",
-    "chartjs-plugin-zoom": "0.6.6",
+    "chartjs-plugin-datalabels": "0.7.0",
```
davidswinegar: looking at this and realizing this is out of date - we should build a check to make sure this is synchronized internally.

PR #356, opened 2020-03-10
MemoryCacheDataProvider prefetches more data for "large" files

MatthewSteel:
##  Summary

While there is space left in the cache and there are no read requests currently in flight, the `MemoryCacheDataProvider` should intelligently fill up its cache.

This is currently already done for files shorter than 35 seconds long, but
 - We'd rather just have one threshold (size) rather than two (size and length), and
 - The technique currently used for short files ("request the whole thing") is not appropriate for files that won't fit in the cache.

The first couple of commits reorder some of the logic in the class to make the call graph simpler, replacing some mutual recursion with explicit iteration. Most of the changes are in the final commit.

## Test plan

A test has been added to demonstrate some new behaviour -- loading data *before* a read-request when there's enough space in the cache. Other behaviour is currently untested, including
* Fetching data close to the right of the cursor preferentially,
* Stopping once the cache is full.

It looks ok when verifying manually with large and small files (for small files the behaviour seems unchanged), but I could use some guidance on adding appropriate unit tests.

In manual testing I saw a couple of errors (one tab-crash and one in-product red bubble saying something about the provider) but I haven't been able to reproduce them. I'll keep trying to reproduce them, but code review might help -- I suspect the problem is in the fourth commit, if anywhere. I haven't seen it in a long time, though, so maybe it doesn't exist?

## Versioning impact

None?

### DIFF
--- a/packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
+++ b/packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
@@ -280,6 +280,10 @@ export default class MemoryCacheDataProvider implements DataProvider {
     }
 
     // Then see if we need to set a new connection based on the new connection and read requests state.
+    this._maybeRunNewConnections();
+  }
+
+  _maybeRunNewConnections() {
     const newConnection = getNewConnection({
       currentRemainingRange: this._currentConnection ? this._currentConnection.remainingBlockRange : undefined,
       readRequestRange: this._readRequests[0] ? this._readRequests[0].blockRange : undefined,
--- a/packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
+++ b/packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
@@ -221,9 +221,7 @@ export default class MemoryCacheDataProvider implements DataProvider {
     return this._preloadTopics;
   }
 
-  // Gets called any time our "connection", read requests, or topics change.
-  _updateState() {
-    // First, see if there are any read requests that we can resolve now.
+  _resolveFinishedReadRequests() {
     this._readRequests = this._readRequests.filter(({ timeRange, blockRange, topics, resolve }) => {
       if (topics.length === 0) {
         resolve([]);
@@ -269,6 +267,12 @@ export default class MemoryCacheDataProvider implements DataProvider {
 
       return false;
     });
+  }
+
+  // Gets called any time our "connection", read requests, or topics change.
+  _updateState() {
+    // First, see if there are any read requests that we can resolve now.
+    this._resolveFinishedReadRequests();
 
     if (this._currentConnection && !isEqual(this._currentConnection.topics, this._getCurrentTopics())) {
       // If we have a different set of topics, stop the current "connection", and refresh everything.
--- a/packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
+++ b/packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
@@ -436,8 +436,9 @@ export default class MemoryCacheDataProvider implements DataProvider {
         );
       }
 
-      // Now `this._recentBlockRanges` and `this._blocks` have been updated, so we can purge the
-      // cache and report progress.
+      // Now `this._recentBlockRanges` and `this._blocks` have been updated, so we can resolve
+      // requests, purge the cache and report progress.
+      this._resolveFinishedReadRequests();
       this._purgeOldBlocks();
       this._updateProgress();
 
@@ -456,11 +457,9 @@ export default class MemoryCacheDataProvider implements DataProvider {
         ...this._currentConnection,
         remainingBlockRange: { start: currentBlockIndex + 1, end: blockRange.end },
       };
-      this._resolveFinishedReadRequests();
     }
     // Connection successfully completed.
     delete this._currentConnection;
-    this._resolveFinishedReadRequests();
     return true;
   }
 
--- a/packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
+++ b/packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
@@ -7,7 +7,7 @@
 //  You may not use this file except in compliance with the License.
 
 import { simplify } from "intervals-fn";
-import { isEqual, uniq } from "lodash";
+import { isEqual, sum, uniq } from "lodash";
 import { TimeUtil, type Time } from "rosbag";
 import uuid from "uuid";
 
@@ -21,14 +21,13 @@ import type {
 } from "webviz-core/src/dataProviders/types";
 import filterMap from "webviz-core/src/filterMap";
 import { getNewConnection } from "webviz-core/src/util/getNewConnection";
-import { type Range, mergeNewRangeIntoUnsortedNonOverlappingList } from "webviz-core/src/util/ranges";
+import { type Range, mergeNewRangeIntoUnsortedNonOverlappingList, missingRanges } from "webviz-core/src/util/ranges";
 import reportError from "webviz-core/src/util/reportError";
 import { fromNanoSec, subtractTimes, toNanoSec } from "webviz-core/src/util/time";
 
 // I (JP) mostly just made these numbers up. It might be worth experimenting with different values
 // for these, but it seems to work reasonably well in my tests.
 export const MEM_CACHE_BLOCK_SIZE_NS = 0.1e9; // Messages are laid out in blocks with a fixed number of milliseconds.
-const MINIMUM_CACHE_SIZE_NS = 35e9; // Number of nanoseconds that we'll always keep in the cache.
 const READ_AHEAD_NS = 3e9; // Number of nanoseconds to read ahead from the last `getMessages` call.
 const READ_AHEAD_BLOCKS = Math.ceil(READ_AHEAD_NS / MEM_CACHE_BLOCK_SIZE_NS);
 const CACHE_SIZE_BYTES = 2.5e9; // Number of bytes that we aim to keep in the cache.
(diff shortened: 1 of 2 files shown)


packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
```diff
@@ -267,24 +268,33 @@ export default class MemoryCacheDataProvider implements DataProvider {
...
           `MemoryCacheDataProvider connection ${this._currentConnection ? this._currentConnection.id : ""}`,
           err ? err.message : "<unknown error>",
           "app"
         );
       });
+      if (!connectionSuccess) {
+        // Connection interrupted, or otherwise unsuccessful.
+        break;
+      }
+      // See if there are any more read requests we should field.
```
janpaul123: Maybe it would be even clearer to call `_resolveFinishedReadRequests` here instead of before the return statements below?

janpaul123: (Ah you still need it at the end of the while loop, but maybe the one [here](https://github.com/cruise-automation/webviz/pull/356/commits/d9b3bbbbb44634ee38f844759afd1d9b94dba36f#diff-ff549a830b60d70e1ed51ce42222356fR386) would be clearer when moved? Not sure.)

MatthewSteel: I think that makes sense for the first one, but not for the second. We know that `_setConnection` will return before `this._readRequests[0]` is done, but inside the `while` loop we might also complete *other* read requests.

What do you think about moving it just underneath the call to `_updateProgess` and `_purgeOldBlocks`?

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
```diff
@@ -129,6 +134,9 @@ export default class MemoryCacheDataProvider implements DataProvider {
   // Total length of the data in nanoseconds. Used to compute progress with.
   _totalNs: number;
 
+  // Block index of the last read request. Used as a heuristic for a good place to prefetch data.
+  _cursorPosition: number;
```
janpaul123: I think this is a duplicate of `_lastResolvedCallbackEnd`? Though I like `_cursorPosition` better as a name. 😄

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
janpaul123: Oh yeah that looks like a great place!

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
MatthewSteel: Done, thanks.

janpaul123: Oh no I guess it's subtly different; you set `_cursorPosition` to the start of the range when adding a new read request, and we set `_lastResolvedCallbackEnd` to the end of a range when resolving a read request. Still, feels like we're using them for similar things…

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
```diff
@@ -106,6 +94,23 @@ export function getBlocksToKeep({
   return { blockIndexesToKeep, newRecentRanges: recentBlockRanges };
 }
 
+// Get the best place to start prefetching a block, given the uncached ranges and the cursor position.
+// In order of preference, we would like to prefetch:
+// - The leftmost uncached block to the right of the cursor, or
+// - The leftmost uncached block to the left of the cursor, if one does not exist to the right.
+//
+// Exported for tests.
+export function getPrefetchStartPoint(uncachedRanges: Range[], cursorPosition: Number): Number {
```
janpaul123:
```suggestion
export function getPrefetchStartPoint(uncachedRanges: Range[], cursorPosition: number): number {
```

Flow/JS sadly has both `number` and `Number` but we typically use the lowercase one which explicitly denotes the primitive (e.g. `(5 instanceof Number) === false` 😭).

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
janpaul123: In fact then you can even put it before `_purgeOldBlocks`. Would that solve the bug you found?

MatthewSteel: Done. I put it above those calls instead of below, because there's a *tiny* chance we might purge something that could be used to resolve a request. (This sadly doesn't fix the other bug.)

janpaul123: Haha looks like we're getting some nice race conditions in our comments. 😉

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
```diff
@@ -268,19 +273,46 @@ export default class MemoryCacheDataProvider implements DataProvider {
...
+
+  _getPrefetchRange() {
+    const bounds = { start: 0, end: this._blocks.length };
+    const uncachedRanges = missingRanges(bounds, this._getDownloadedBlockRanges());
+    if (!uncachedRanges.length) {
+      return; // We have loaded the whole file.
+    }
+
+    const prefetchStart = getPrefetchStartPoint(uncachedRanges, this._lastResolvedCallbackEnd || 0);
+    return { start: prefetchStart, end: prefetchStart + 1 };
```
janpaul123: Instead of `prefetchStart + 1`, should we set this to the start of the next downloaded range? It's a bit minor, but we'd get some small benefits from it, e.g. when the next read request comes in and it's about to get downloaded (e.g. it's only a few blocks "away"), we won't be terminating the connection (which can be expensive) per the `continueDownloadingThreshold`.

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
MatthewSteel: Yeah, they're different, but from playing around with it it works fine. I tried some other things and they had problems, but `_lastResolvedCallbackEnd` does the right thing.

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
MatthewSteel: Done. Fixed another couple of flow problems, let me know if the fixes aren't great.

Also: I rebased and force-pushed those fixes, but I guess I shouldn't have, right? Should I be waiting until some time later to make the history pretty?

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
```diff
@@ -393,6 +425,8 @@ export default class MemoryCacheDataProvider implements DataProvider {
       };
       this._resolveFinishedReadRequests();
     }
+    // Implicit return impossible, but satisfy flow:
```
janpaul123: In that case you can also throw an error here instead. I think that should also satisfy Flow, and would actually alert us if this assumption was violated.

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
MatthewSteel: The start of the next downloaded range could be far away, so I guess we need a way to cut off that connection if we hit our memory budget if we do that.

Also, it would cause blocks in `_recentBlockRanges` to merge. Is that problematic? Would we instead want to aim on block earlier?

janpaul123: Ah yeah I just realised the memory budget issue. In that case I _think_ that the current approach is a bit clearer. Might be worth a comment somewhere?

Re blocks in `_recentBlockRanges` merging; that will also happen with the current approach right, when we fill in all the blocks in between two ranges? I think that's okay, and even expected.

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
MatthewSteel: Haha, oh dear -- flow has a problem with the early-return, and eslint has a problem with unreachable code :-). I've made the success case break out of the loop to try to satisfy both. Will rebase it onto the relevant earlier commit later.

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
MatthewSteel: Yeah, we want it to happen if they *actually* merge, but if they're a million miles apart I think it's a bad outcome -- we'll never actually join them up until/unless we play the whole intervening interval, but we make the later block "seem recently used" right away.

The "one block short" hack isn't great either -- if it's preloading a big chunk of data and the user seeks somewhere else, there's a big invisible range inside `_recentBlockRanges` that could interfere with things, I think.

We could request a larger range and signal to `_setConnection` that this is a low-priority connection that can be stopped as soon as we start evicting things. It's another hack, though.

I've added a comment explaining why the code does what it does.

janpaul123:
This is looking really good! Looking carefully through the code I can't spot anything that could've caused the errors you mentioned. It's also very possible that they existed before this change, so unless we can reproduce them, I'd be fine with it.

I do think it would be good to now add the tests you mentioned.

As for version bumps, since this is a change to webviz-core which we haven't packaged up currently, so don't worry about it.

janpaul123:
Also be sure to rebase against #354 because it will definitely conflict! 😮

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
janpaul123: I think that's okay. Usually when we merge we squash it all into one commit. And it was very useful for me to read through it commit-by-commit!

MatthewSteel:
Some more unit tests added, rebased into the "uniform prefetching" commit. (That commit is unfortunately getting kinda big...)

The PR is now based on top of the commit inside #354. Let me know if I've done that wrong, or if I need to do something else like pointing this PR at that one.

Seeeems to still work.

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
MatthewSteel: > Usually when we merge we squash it all into one commit

Anyone reading the git-blame to understand why the code is the way it is will have even less context than you do now. If individual commits add clarity to reviewers, I think it's extra important they be preserved for maintainers.

janpaul123: Yeah, I tend to agree with you, but in practice you can look up the original PR in Github (which is linked through the PR number) and get the context that way. It is unfortunate though that the context is not captured in git itself though. If you feel strongly maybe we can discuss your proposal at our next retrospective meeting?

janpaul123:
@MatthewSteel can you rebase against master again since I've merged #354 and the diff is now all messed up. :)

PR #357, opened 2020-03-10
Make screenshot tests more verbose

janpaul123:
And also only do one at a time.

This makes it so that screenshots are less likely to cause a timeout in CircleCI, and also gives us more information to debug if a screenshot hangs the system.

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,8 +134,9 @@
     "docs-deploy": "cp -r docs/public __temp_deploy__ && cp --remove-destination packages/webviz-core/public/index.html __temp_deploy__/app/index.html && echo 'webviz.io' > __temp_deploy__/CNAME && gh-pages -d __temp_deploy__ && rm -rf __temp_deploy__",
     "publish": "lerna run clean && lerna run build && lerna publish",
     "screenshot": "NODE_ENV=development storybook-chrome-screenshot -c stories --parallel 8 --inject-files stories/injected.js --browser-timeout 1200000",
-    "screenshot-ci": "npm run screenshot -- --parallel 1 --debug --puppeteer-launch-config '{\"executablePath\": \"/usr/bin/google-chrome\", \"headless\": false, \"args\":[\"--use-gl=swiftshader\", \"--no-sandbox\",\"--disable-setuid-sandbox\", \"--disable-dev-shm-usage\", \"--headless\", \"--mute-audio\", \"--user-agent=PuppeteerTestingChrome/80.\"]}'",
-    "screenshot-debug": "npm run screenshot -- --puppeteer-launch-config '{\"headless\": false, \"--args\":[\"--user-agent=PuppeteerTestingChrome/80.\"]}'",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -134,8 +134,8 @@
     "docs-deploy": "cp -r docs/public __temp_deploy__ && cp --remove-destination packages/webviz-core/public/index.html __temp_deploy__/app/index.html && echo 'webviz.io' > __temp_deploy__/CNAME && gh-pages -d __temp_deploy__ && rm -rf __temp_deploy__",
     "publish": "lerna run clean && lerna run build && lerna publish",
     "screenshot": "NODE_ENV=development storybook-chrome-screenshot -c stories --parallel 8 --inject-files stories/injected.js --browser-timeout 1200000",
-    "screenshot-ci": "npm run screenshot -- --puppeteer-launch-config '{\"executablePath\": \"/usr/bin/google-chrome\", \"headless\": false, \"args\":[\"--use-gl=swiftshader\", \"--no-sandbox\",\"--disable-setuid-sandbox\", \"--disable-dev-shm-usage\", \"--headless\", \"--mute-audio\", \"--user-agent=PuppeteerTestingChrome/72.\"]}'",
-    "screenshot-debug": "npm run screenshot -- --parallel 1 --debug --puppeteer-launch-config '{\"headless\": false, \"--args\":[\"--user-agent=PuppeteerTestingChrome/72.\"]}'",
```
MatthewSteel: So there are two changes here, right?

1. `screenshot-ci` logs more, and does one at a time, and

2. `screenshot-debug` logs *less*, and does more than one at a time?

A couple of questions from reading the change and the PR message:

* Why is it less likely to time out in CI if CI just does one at a time now?

* Why remove debug logging from `screenshot-debug`?

janpaul123: Oh shoot I misread, and thought that `screenshot-debug`, called `screenshot-ci` and so that those flags would be redundant. I'll restore `screenshot-debug` to its original state.

> Why is it less likely to time out in CI if CI just does one at a time now?

I don't think it is, but I was suspecting that the browser might hang sometimes, in which case it'd be easier to pinpoint which test caused that if we do one at the time. I'm not sure though, so maybe we should for now only make it more verbose and see if that helps enough. I'll do that.

janpaul123: Oh that said @hhsaez said "I ran screenshot-debug from your branch and it now finishes in under 5 minutes in my machine (previously it took 16 min) :)" so maybe I'll keep around my modified `screenshot-debug` line as well and give them both more explicit names.

hhsaez: I think the difference is whether or not debug mode is on. If enabled, screenshot-debug takes a longer time than without it (which make sense, I think). Unfortunately, I'm not familiar enough with this process to understand if debug mode is really needed or nor.

janpaul123: Yeah, and the parallelism probably. I split it out to an `npm run screenshot-local` and an `npm run screenshot-local-debug`. 👍

janpaul123: Hm, in CI [this job](https://circleci.com/gh/cruise-automation/webviz/2039) with normal parallelism seems to hang. And also it doesn't look like it's much faster anyway. So I'm going to only do 1 screenshot at a time in CI after all.

PR #360, opened 2020-03-11
<GLText> improvements for text rendering

hhsaez:
## Summary

**Note: I'm creating a new PR as a workaround for the current issues with CI. For details and previous discussions, please check #349**

This PR improves Worldview's existing GLText command, adding new functionalities for enhanced font rendering and scale invariance. 

### Higher resolution font
GLText now uses a higher resolution font atlas, which increases the resolution of the SDF texture, avoiding visual artifacts when zooming in/out the text:

**Before:**
<image src="https://user-images.githubusercontent.com/1727802/76123182-d3c2e400-5fac-11ea-9788-47758d12f919.png" height="300px" />

**Now:**
<image src="https://user-images.githubusercontent.com/1727802/76123180-d32a4d80-5fac-11ea-89c6-bfc634bf5809.png" height="300px" />

By default, <GLText /> will use the best possible resolution but it's also possible to customize this setting by using the *resolution* property:

`<GLText resolution="40">{markers}</GLText>`

### Scale Invariance
Keeps the text at the same size (relative to the screen resolution) regardless of the zoom level in the 3D world. It produces a final result similar as when using the <Text> command (which renders text using HTML elements on top of the 3D video).

**Note:** scaleInvariant does not work when the billboard option is false for a marker.

Usage: 
`<GLText scaleInvariantFontSize="40">{markers}</GLText>`

The scaleInvariantFontSize option allows you to change the font size for the text when scale invariant is active. 

Because of how the SDF approach works, visual artifacts may appear when the text size is too small. In such cases, I'm skipping smoothing the distance value from the SDF texture which produce better results. This is handled by fragment shader.

I also updated the documentation with the new *scaleInvariantFontSize* option

### Alphabet

GLFont will generate a new texture atlas whenever new characters appear in markers. If those characters are known in advance, the optional `alphabet` property can be use as an optimization, reducing the number of times the atlas needs to be generated.

Usage: 
`<GLText alphabet={['a', 'b', 'c', 'd', 'e']}>{markers}</GLText>`

## Test plan
Added several stories to showcase the new options and compare their output with the default implementation:

* *resolution - XX*: Renders GLText at a higher zoom level to showcase font at different resolutions. 
  * *resolution - default*: Default resolution (160)
  * *resolution - 80*: Resolution set to 80
  * *resolution - 40*: Resolution set to 40 (min value)
* *scaleInvariant*: Shows how GLText renders text using the scaleInvariantFontSize. Three tests are provided: 
  * *scaleInvariant*: Font size is 10px. Does not use smooth interpolation.
  * *scaleInvariant-20*: Font size is 20px. Does not use smooth interpolation.
  * *scaleInvariant-40*: Font size is 40px. Since this is a bigger font, it will use smooth interpolation in the fragment shader.
* *scaleInvariant resize*: Makes sure scaleInvariant works correctly when the canvas' dimensions are changed after rendering.
* *alphabet*: Shows how to use the alphabet property to lower the amount of times the texture atlas needs to be re-generated due to new characters.
* *autoBackgroundColor scaleInvariant*: Same as *autoBackgroundColor* story, but using the scaleInvariant option
* *changing text*: Updated version that avoid unpredictable results and allows you to change the text manually.

## Versioning impact
Minor


janpaul123:
@hhsaez this looks great overall, but the Worldview-docs-GLText.png screenshot seems to be broken (see report above). The other screenshot changes look good (everything seems to have gotten crisper!).

docs/src/jsx/commands/GLText.js
```diff
@@ -171,7 +171,7 @@ function Example() {
           phi: oldState.phi,
         }))
       }>
-      <GLText>{matrix.toMarkers()}</GLText>
+      <GLText alphabet={ALPHABET}>{matrix.toMarkers()}</GLText>
```
troygibb: where is `ALPHABET` defined?

hhsaez: It's in that same file, line24:
https://github.com/cruise-automation/webviz/blob/0d445c5e315296af47fb9376068d2fe7ac999f0f/docs/src/jsx/commands/GLText.js#L24

hhsaez:
Updated documentation with the new optional `resolution` and `alphabet` options

docs/src/4.15.GLText.mdx
```diff
@@ -7,3 +7,29 @@ import { GLText } from "./jsx/allLiveEditors";
...
+
+The `resolution?: number` optional property allows to override this behavior by providing a custom character resolution value. If your application does not need to render text that is too close the camera, it is recommended to reduce the font resolution to ensure a better performance.
+
+Default value is `resolution={160}`. Minimum value is `resolution={40}`. 
+
+## Alphabet
+
+`<GLText />` generates a new texture atlas on the fly whenever a merker's text contains at least one character that was not rendered before. This is a complex operation and may impact on the performance of your application. 
+
+For that reason, is recommended to use the optional `alphabet?: string[]` property to provide a list of characters that are allowed as part of any marker's text.
```
troygibb:
```suggestion
In order to avoid texture generation when you have an expected subset of characters, you can use the optional `alphabet?: string[]`. 
```

docs/src/4.15.GLText.mdx
```diff
@@ -7,3 +7,29 @@ import { GLText } from "./jsx/allLiveEditors";
...
+
+Default value is `resolution={160}`. Minimum value is `resolution={40}`. 
+
+## Alphabet
+
+`<GLText />` generates a new texture atlas on the fly whenever a merker's text contains at least one character that was not rendered before. This is a complex operation and may impact on the performance of your application. 
+
+For that reason, is recommended to use the optional `alphabet?: string[]` property to provide a list of characters that are allowed as part of any marker's text. 
+
+For example, if markers are known to have only digits, you can use the `alphabet` property as follows to reduce the number of times the texture atlas need to be re-generated from (potentially) 10 times to only once.
```
troygibb:
```suggestion
For example, if markers are known to have only digits, you can use the `alphabet` property as follows to reduce the number of times the texture atlas needs to be re-generated from a worst case scenario of 10 times to only once.
```

packages/webviz-core/package-lock.json
```diff
@@ -50,11 +50,6 @@
         }
       }
     },
-    "@cruise-automation/hooks": {
-      "version": "0.0.1",
-      "resolved": "https://registry.npmjs.org/@cruise-automation/hooks/-/hooks-0.0.1.tgz",
-      "integrity": "sha512-sUwYShqRJAyg2XS7+uc9ZwdNkRArLdyjPTF6Ebl1GphXHX5QYU14aMt3SO3QtVg0fd/KSWM4FinRosxZDq5uZg=="
-    },
```
troygibb: FYI these changes should not be committed to source control, since you haven't changed anything in `package.json`.

hhsaez: I just reverted the changes to package-lock.json in a new commit

PR #363, opened 2020-03-12
GLText selection

hhsaez:
## Summary

This PR implements selection for GLText command. 

As for other commands (i.e. Cubes, Cones, etc), we need to set the getChildrenForHitmap property in <Command> to a valid function. But in the case of GLText, that's not enough.

In contrast with other commands that uses fromGeometry() or withPose() utility functions (i.e. Cubes), GLText has a custom shader and properties which prevent the correct rendering of the selectable objects to the hitmap. One of those problems is that the color is not set correctly (when hitmap is on, Color is a vec4 instead of a Color object) and nothing gets rendered. 

In addition, we don't want to render the text as is into the hitmap because that will lead to a bad user experience. If so, users will have to click exactly in those pixels where the glyphs are, preveting selection if a user clicks inside the letter "O", for example. For that reason, whenever we need to render the hitmap for GLText, we're rendering the polygons instead of the glyphs (see comment in fragment shader).

I also added integration tests, described below. 

## Test plan

I added several stories handling special cases for text selection:

* Single GLText object with default properties
* Single GLText with combinations of billboard, background and scale invariance
* GLText and background object, with stacked objects on and off.
* GLText just containing the letter "O", to prove that selection works when using polygons instead of glyphs (with and without background objects).

## Versioning impact

Minor


packages/regl-worldview/src/commands/GLText.js
```diff
@@ -354,7 +363,19 @@ function makeTextCommand(alphabet?: string[]) {
...
 
-        const fgColor = marker.colors?.[0] || marker.color || BG_COLOR_LIGHT;
+        const hitmapColor = (markerColor) => {
+          return {
+            r: markerColor?.[0] ?? 0,
+            g: markerColor?.[1] ?? 0,
+            b: markerColor?.[2] ?? 0,
+            a: markerColor?.[3] ?? 1,
+          };
+        };
```
troygibb: I'd lift this function definition to be standalone somewhere else in the file, since it'll get re-instantiated for every marker on every re-render.

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -260,42 +266,45 @@ function makeTextCommand(alphabet?: string[]) {
...
-        highlightColor: (ctx, props) => ({ buffer: props.highlightColor, divisor: 1 }),
-        enableBackground: (ctx, props) => ({ buffer: props.enableBackground, divisor: 1 }),
-        enableHighlight: (ctx, props) => ({ buffer: props.enableHighlight, divisor: 1 }),
-        posePosition: (ctx, props) => ({ buffer: props.posePosition, divisor: 1 }),
-        poseOrientation: (ctx, props) => ({ buffer: props.poseOrientation, divisor: 1 }),
-      },
-    });
-
-    return (props: $ReadOnlyArray<TextMarkerProps>) => {
+    const makeDrawText = function(forHitmap: boolean) {
```
troygibb:
```suggestion
    const makeDrawText = (forHitmap: boolean) => {
```

troygibb:
Also the title still says `[WIP]`, I'm assuming that is not the case anymore?

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -142,7 +142,7 @@ const vert = `
   attribute vec2 srcOffset;
   attribute float srcWidth;
   attribute vec2 destOffset;
-
+
```
janpaul123: #nit unnecessary whitespace change. I have my editor set up to remove trailing whitespaces from each line on save; I'd recommend this to everyone for consistency. 😄

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -354,7 +363,19 @@ function makeTextCommand(alphabet?: string[]) {
         let y = 0;
         let markerInstances = 0;
 
-        const fgColor = marker.colors?.[0] || marker.color || BG_COLOR_LIGHT;
+        const hitmapColor = (markerColor) => {
+          return {
+            r: markerColor?.[0] ?? 0,
```
janpaul123: For hitmaps, is it every possible that `markerColor` is not set? That shouldn't happen right? Should we just throw an error in that case instead of defaulting to black? (Or maybe I just don't understand how our hitmaps work?)

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -354,7 +363,19 @@ function makeTextCommand(alphabet?: string[]) {
         let y = 0;
         let markerInstances = 0;
 
-        const fgColor = marker.colors?.[0] || marker.color || BG_COLOR_LIGHT;
+        const hitmapColor = (markerColor) => {
```
janpaul123: Would be good to have a type annotation for the argument here.

packages/regl-worldview/src/stories/assertionStories/GLText.stories.js
```diff
@@ -0,0 +1,145 @@
...
+      },
+      scale: { x: 1, y: 1, z: 1 },
+      color,
+      colors: background ? [color, { r: 1, g: 1, b: 0, a: 1 }] : undefined,
+      billboard,
+    },
+  ];
+}
+
+function createAssertionTest(markers, scaleInvariant) {
```
janpaul123:
```suggestion
function createAssertionTest({ markers, scaleInvariant }: {| markers: TextMarker[], scaleInvariant: boolean |}) {
```

I found the tests below a bit confusing to read, since they sometimes have a `, true`, but it always means something different (sometimes it means `scaleInvariant`,  sometimes it means `stackedObjectsEnabled`). It would be a lot clearer if we pass in an object since you can then easily see the name of the thing that gets enabled.

Also would be good to add some more type information so Flow can help us better.

packages/regl-worldview/src/stories/assertionStories/GLText.stories.js
```diff
@@ -0,0 +1,145 @@
...
+
+function textMarkers({
+  text,
+  billboard,
+  background = false,
+}: {
+  text: string,
+  billboard?: ?boolean,
+  background?: ?boolean,
+}) {
```
janpaul123:
```suggestion
}): TextMarker[] {
```

packages/regl-worldview/src/stories/assertionStories/GLText.stories.js
```diff
@@ -0,0 +1,145 @@
...
+import Cubes from "../../commands/Cubes";
+import GLText from "../../commands/GLText";
+import { clickAtOrigin, WorldviewWrapper } from "../worldviewAssertionUtils";
+import { assertionTest } from "stories/assertionTestUtils";
+
+function textMarkers({
+  text,
+  billboard,
+  background = false,
+}: {
```
janpaul123:
```suggestion
}: {|
```

We try to use [exact object types](https://flow.org/en/docs/types/objects/#toc-exact-object-types) as much as possible since then Flow will complain if you accidentally pass in unknown fields (useful for spelling mistakes, etc).

packages/regl-worldview/src/commands/GLText.js
hhsaez: @janpaul123 right, it shouldn't happen. Here's my rationale for this fix:

The function that calculates the color for hitmap always returns a valid instance (although it's a Vec4[] instead of Color). I might be wrong, but in my tests the color is never black, since that means nothing is selected. So, in the event that we don't get a valid color, we can assume that the object is not selectable, right? If so, using black should be ok.

packages/regl-worldview/src/stories/assertionStories/GLText.stories.js
hhsaez: @janpaul123 There are two functions for creating tests:
* `createAssertionTest` accepts an optional `scaleInvariant` argument. 
* `createAssertionTestWithBackgroundObjects` accepts an optional `stackedObjectsEnabled` argument and scale invariant is not used.
But I'll add your suggestion anyway since it's more clear that way

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -211,6 +211,7 @@ const frag = `
   uniform float cutoff;
   uniform bool scaleInvariant;
   uniform float scaleInvariantSize;
+  uniform bool forHitmap;
```
vidaaudrey: We use `isHitmap` in other files, would be nice to keep consistent

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -354,7 +374,10 @@ function makeTextCommand(alphabet?: string[]) {
         let y = 0;
         let markerInstances = 0;
 
-        const fgColor = marker.colors?.[0] || marker.color || BG_COLOR_LIGHT;
+        // If we need to render text for hitmap framebuffer, we only render the polygons using
+        // the foreground color (which needs to be converted to RGBA since it's a vec4).
+        // See comment on fragment shader above
+        const fgColor = forHitmap ? hitmapColor(marker.color) : marker.colors?.[0] || marker.color || BG_COLOR_LIGHT;
```
vidaaudrey: Since we use the color to locate objectId, would be good to be more strict about fallbacks. 

When `hitmap` is enabled, the `color` or `colors` should always be present.  
 since `GLText` doesn't have any `colors` as input props, we don't need to fall back to `marker.colors?.[0]`? https://github.com/cruise-automation/webviz/blob/master/packages/regl-worldview/src/utils/getChildrenForHitmapDefaults.js#L75-L104

If `color` is not present, we could fallback to `[0, 0, 0, 1]` which is the default canvas color, similar to GLTFScene: https://github.com/cruise-automation/webviz/blob/master/packages/regl-worldview/src/commands/GLTFScene.js#L203

packages/regl-worldview/src/stories/assertionStories/GLText.stories.js
janpaul123: Yeah I understood how the code works, but it took me a while to realise it.

packages/regl-worldview/src/commands/GLText.js
hhsaez: Thanks JP! I already setup vscode to remove trailing whitespaces from now on

packages/regl-worldview/src/stories/assertionStories/GLText.stories.js
hhsaez: Right. I pushed an updated version of createAssertionTest() that uses a single object  as parameter. As a side effect of that, there's now only one function to setup the test that handles both scale invariance and stacked objects. Thanks!

packages/regl-worldview/src/stories/assertionStories/GLText.stories.js
hhsaez: Great! I updated my tests to use exact objects.

packages/regl-worldview/src/commands/GLText.js
hhsaez: Yes, I agree. We should keep the code as consistent as possible. I already updated my code to use isHitmap. Thanks!

packages/regl-worldview/src/commands/GLText.js
hhsaez: @vidaaudrey Sorry, I'm not sure I understand. Should I fallback to `marker.colors?.[0]` if `marker.color` is undefined when rendering for the hitmap? And only then fallback to black if both `marker.color` and `marker.colors?.[0]` are both undefined? 

Also, I'm already falling back to black in `hitmapColor()` if marker.color is not valid in order to avoid selection. Is that correct?

packages/regl-worldview/src/stories/assertionStories/GLText.stories.js
hhsaez: Done!

packages/regl-worldview/src/commands/GLText.js
vidaaudrey: Hi sorry, i didn't see your fallback inlined `: markerColor?.[0] ?? 0,`.
nit: would be nice to write it like `const fgColor = isHitmap ? (marker.color || [0, 0, 0, 1]) : ...` so it's a more explicit.  Separately, could also fix the polygon to support vec4 as color input if it doesn't already.

janpaul123: On that note maybe we already have a utility helper somewhere to convert between object/array representations?

vidaaudrey: yeah, quite a few utils here: https://github.com/cruise-automation/webviz/blob/master/packages/regl-worldview/src/utils/commandUtils.js#L55

hhsaez: @vidaaudrey @janpaul123 Check out my latest commit. I modified TextMarker to use `Color | Vec4` for both `color` and `colors` properties, in the same way other commands are doing it. That let me use the existing `vec4ToRGBA` utils but I had to add a new one called `toColor` for color conversion in order to fix flow errors because of the union type. Please take a look and let me know what you think.

packages/regl-worldview/src/commands/GLText.js
hhsaez: While this code is outdate, I am use explicit types now. Let me know what you think.

packages/regl-worldview/src/commands/GLText.js
janpaul123: Yeah looks good to me!

PR #365, opened 2020-03-13
Evict cached blocks from end of ranges sometimes

MatthewSteel:
## Summary

Fix existing bug: Seeking to a soon-to-be-evicted block during playback
results in an infinite loop in which a block is requested, immediately
evicted, and then requested again etc.

This change fixes the bug by trying to not evict blocks in those
circumstances: When there are open read requests, we avoid evicting from
the edge of blocks closest to the first ("highest priority") request.

If there are no open read requests, perhaps because we've just resolved
the last one, try not to evict blocks near the last resolved request too
-- odds are fair we'll request another block nearby soon.

Also adds the small benefit that when there are two "close" ranges with a
gap in between, and the user is playing at the end of the first (earlier)
region, as the buffer of the first region extends, the cached blocks of the
second region are evicted _from the end_, and the two regions eventually
meet.

## Test plan

I've added a couple of unit tests to cover explicitly passing locations to the right- and left- edges of a block range, and existing unit tests cover the (unusual) case where we _don't_ pass a hint to `getBlocksToKeep`.

## Versioning impact

Just a bugfix. `getBlocksToKeep` is "exported for tests", I don't think anyone else depends on its signature.

### DIFF
--- a/packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
+++ b/packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
@@ -49,6 +49,7 @@ export function getBlocksToKeep({
   recentBlockRanges,
   blockSizesInBytes,
   maxCacheSizeInBytes,
+  badEvictionLocation,
 }: {|
   // The most recently requested block ranges, ordered from most recent to least recent.
   recentBlockRanges: Range[],
@@ -58,6 +59,8 @@ export function getBlocksToKeep({
   blockSizesInBytes: (?number)[],
   // The maximum cache size in bytes.
   maxCacheSizeInBytes: number,
+  // A block index to avoid evicting blocks from near.
+  badEvictionLocation: ?number,
 |}): { blockIndexesToKeep: Set<number>, newRecentRanges: Range[] } {
   let cacheSizeInBytes = 0;
   const blockIndexesToKeep = new Set<number>();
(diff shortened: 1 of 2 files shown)


packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
```diff
@@ -66,7 +69,8 @@ export function getBlocksToKeep({
     const blockRange = recentBlockRanges[blockRangeIndex];
 
     // Work backwards from the end of the range, since those blocks are most relevant to keep.
```
janpaul123: Comment is out of date now?

MatthewSteel: I guess it depends on how you interpret "the end" :thinking:

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
```diff
@@ -84,14 +89,35 @@ export function getBlocksToKeep({
         return {
           blockIndexesToKeep,
           // Adjust the oldest `newRecentRanges`.
-          newRecentRanges: [...recentBlockRanges.slice(0, blockRangeIndex), { start: blockIndex, end: blockRange.end }],
+          newRecentRanges: [
+            ...recentBlockRanges.slice(0, blockRangeIndex),
+            { start: Math.min(blockIndex, startIndex), end: Math.max(blockIndex, startIndex) + 1 },
```
janpaul123: This looks correct, but it took me a few minutes to wrap my head around it. Maybe clearer would be something like `increment > 0 ? { … } : { … }`?

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
```diff
@@ -84,14 +89,35 @@ export function getBlocksToKeep({
...
         };
       }
     }
   }
   return { blockIndexesToKeep, newRecentRanges: recentBlockRanges };
 }
 
+// Helper to identify which end of a block range is most appropriate to evict when there is an open
+// read request.
+function getEvictionIterationParameters(
```
janpaul123: #nit Maybe `getBlocksToKeepDirection` or something?

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
```diff
@@ -486,11 +512,19 @@ export default class MemoryCacheDataProvider implements DataProvider {
       return;
     }
 
+    // If we have open read requests, we really don't want to evict blocks in or near the first one.
+    // If we don't have open read requests, try not to evict blocks near the last one we fetched
+    // (if any), because our next request will likely be nearby.
```
janpaul123: Very clear! 👍

packages/webviz-core/src/dataProviders/MemoryCacheDataProvider.js
MatthewSteel: You're right, that's much better.

PR #368, opened 2020-03-16
Stop NodeDataWorker in UserNodePlayer on close

MatthewSteel:
## Summary

We're currently leaking a 50MB `NodeDataWorker` every time we open a
file. (Most of the weight is probably typescript?) We should explicitly
terminate the worker instead.

Alternatively, we could reuse these -- either by keeping references to
old unused workers, or by making them static and never terminating them
(assuming they're stateless.) To inform a discussion of this point, I've
measured the time to instantiate a `NodeDataWorker` in an `Rpc` and
send it a dummy message. It takes between 1.5 and 1.6 seconds on my
personal laptop.

A design question, answerable later: Should `Rpc` have a `terminate()`
method? I imagine this sort of thing is pretty common.

Also something to fix later: On application startup, we also leak a
650KB roslib "workersocket" every second. Not much, but it's over 2 gigs
per hour so it could occasionally cause users pain.

## Test plan

I've loaded the app a few times, imported a few bags and made sure the web-workers have been cleaned up. If you think this could use some unit testing I'd definitely appreciate some ideas on how to do that.

## Versioning impact

None.

### DIFF
--- a/packages/webviz-core/src/players/UserNodePlayer/index.js
+++ b/packages/webviz-core/src/players/UserNodePlayer/index.js
@@ -39,7 +39,7 @@ import {
 import type { UserNodeLog } from "webviz-core/src/players/UserNodePlayer/types";
 import type { UserNodes } from "webviz-core/src/types/panels";
 import type { RosDatatypes } from "webviz-core/src/types/RosDatatypes";
-import Rpc from "webviz-core/src/util/Rpc";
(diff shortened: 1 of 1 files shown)


troygibb:
Oh and around unit testing, I don't see any great unit testing opportunities here. Seems fine the way it is.

PR #372, opened 2020-03-18
Fix glyph size when rendering text

hhsaez:
## Summary

There's a bug in `<GLText />` resulting in some characters being cropped when rendering:

Screen Shot 2020-03-18 at 11 31 43 AM

The problem is that we're not taking into account the small margin (named `BUFFER` in the code) for each glyph that is used when computing the SDF for the font.

This is the result after my fix:

Screen Shot 2020-03-18 at 11 31 04 AM

I added a comment and a test as well 

## Test plan

Added a story to make sure each character is not cropped. It will help if we change the font in the future (I noticed a comment about that).

## Versioning impact

This should be a patch. No new functionality was added.

### DIFF
--- a/packages/regl-worldview/src/stories/GLText.stories.js
+++ b/packages/regl-worldview/src/stories/GLText.stories.js
@@ -89,6 +89,21 @@ storiesOf("Worldview/GLText", module)
       </Container>
     );
   })
+  .add("glyph ascent and descent", () => {
+    const markers = textMarkers({ text: "BDFGHJKLPQTY\nbdfghjklpqty", billboard: true });
+    const target = markers[9].pose.position;
+    return (
+      <Container
+        cameraState={{
+          target: [target.x, target.y + 2, target.z],
+          perspective: true,
+          distance: 8,
+        }}>
+        <GLText>{markers}</GLText>
+        <Axes />
+      </Container>
+    );
+  })
   .add("scaleInvariant", () => {
     const markers = textMarkers({ text: "Hello\nWorldview", billboard: true });
     return (
--- a/packages/regl-worldview/src/commands/GLText.js
+++ b/packages/regl-worldview/src/commands/GLText.js
@@ -272,7 +272,7 @@ function makeTextCommand(alphabet?: string[]) {
         // For the font size, make sure there's enough room to handle
         // both ascent and descent for each glyph. Otherwise, some characters
         // might get cropped when rendering
-        fontSize: command.resolution + BUFFER,
+        fontSize: command.resolution,
         cutoff: CUTOFF,
         scaleInvariant: command.scaleInvariant,
         scaleInvariantSize: command.scaleInvariantSize,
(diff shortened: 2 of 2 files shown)


packages/regl-worldview/src/stories/GLText.stories.js
```diff
@@ -89,6 +89,21 @@ storiesOf("Worldview/GLText", module)
       </Container>
     );
   })
+  .add("glyph ascent and descent", () => {
+    const markers = textMarkers({ text: "BDFGHJKLPQTY\nbdfghjklpqty", billboard: true });
+    const target = markers[9].pose.position;
+    return (
+      <Container
+        cameraState={{
```
vidaaudrey: nit: `cameraState` usually require `onCameraStateChange` together
```suggestion
        defaultCameraState={{
```

jtbandes:
> we're not taking into account the small margin (named BUFFER in the code) for each glyph

Isn't this what the `srcOffset` is doing? Is the descender actually going outside the BUFFER area? It seems like your change has made all the text slightly smaller and I'm not sure if that was intentional (probably doesn't matter since the size is kind of arbitrary anyway).

packages/regl-worldview/src/stories/GLText.stories.js
hhsaez: @vidaaudrey Please correct me if I'm wrong, but since I'm using the `<Container />` class I don't need to specify `onCameraStateChange`, right?

hhsaez:
@jtbandes you're right. It makes more sense to use `srcOffset` to handle this fix and prevents the scaling issue. Then, the solution is to duplicate the BUFFER offset for the y-coordinate to make sure the descent is not cropped.

hhsaez:
> One quick nit: When either the background color/highlighting is enabled, I noticed that the added buffer is disproportionate to what's on top of the text. Not necessary to address here, but maybe add a ticket to tackle later?

@troygibb You are right, the background is not balanced. That is because is now leaving more margin for those characters that have a descent, like 'g':

Screen Shot 2020-03-19 at 11 50 10 AM

I think we should open a separated issue for solving this and probably consult with our designer about how it should look. What do you think?

jtbandes:
> duplicate the BUFFER offset for the y-coordinate to make sure the descent is not cropped

Not sure if this is right either. Here is my mental model:



So, the difference between info.x/info.y and the srcOffset should be only 1x buffer, unless this picture is wrong.

I guess the question is, where does the descender actually fall within this box?

hhsaez:
> So, the difference between info.x/info.y and the srcOffset should be only 1x buffer, unless this picture is wrong.
> 
> I guess the question is, where does the descender actually fall within this box?

@jtbandes Yes, the image is mostly right, but there's something missing. Buffer is used by TinySDF for the empty space between characters. It doesn't really has to do with baselines and descenders. You can make it as big as you want, but the descenders are still cropped.

In our case, we're always a small amount of pixels off from the correct vertical center. I though it was `BUFFER` (10 units), but it's not. `BUFFER` works ok for some font resolutions, but not for all of them. So, after some experimentation, I came up with a value that works on every test scenario and it's dependent on the font resolution.

BTW, I'm not 100% sure, but this might be related:
https://github.com/mapbox/tiny-sdf/issues/11

PR #375, opened 2020-03-20
Fix scale invariance for both perspective and orthographic projections

hhsaez:
## Summary

This PR fixes several issues for `<GLText />` when scale invariance is enabled:

### Orthographic projection

Scale invariance is not working correctly when using orthographic projection (`perspective = false` in camera state) because we're using the w-component to compute the text scale. But that component is always 1 in orthographic mode. So, in that case we need a different approach. Check the comments in shaders for more information.

### Pixel size

Rendered text's size was not matching the pixel size specified in scaleInvariantFontSize. Now it does, although the calculations are different for orthographic and perspective projections.

### Marker's scale

After some discussion with the team, we decided to ignore the scale property of individual markers when scale invariance is enabled, since that resulted in text being too big or too small. It is taken into account when not using scale invariance, of course.

### Use Regl Context instead of WorldviewReactContext

I'm using Regl Context values for things like viewport dimensions and pixel sizes, which seem to be a more correct approach than using WorldviewReactContext. I also added a couple of properties for the camera.

## Test plan

Added new tests for orthographic projection and ignoring individual markers' scale when scale invariance is enabled.

## Versioning impact

Patch. No new functionality has been added.

### DIFF
--- a/packages/regl-worldview/src/camera/camera.js
+++ b/packages/regl-worldview/src/camera/camera.js
@@ -78,6 +78,14 @@ export default (regl: any) => {
         billboardRotation(context, props) {
           return selectors.billboardRotation(this.cameraState);
         },
+
+        isPerspective(context, props) {
+          return this.cameraState.perspective;
+        },
+
+        fovy(context, props) {
+          return this.cameraState.fovy;
+        },
       },
 
       // adds view and projection as uniforms to every command
--- a/packages/regl-worldview/src/commands/GLText.js
+++ b/packages/regl-worldview/src/commands/GLText.js
@@ -134,7 +134,6 @@ const vert = `
   uniform float scaleInvariantSize;
   uniform float viewportHeight;
   uniform float viewportWidth;
-  uniform float pixelRatio;
   uniform bool isPerspective;
   uniform float cameraFovY;
 
@@ -207,17 +206,17 @@ const vert = `
         gl_Position = computeVertexPosition(markerSpacePos);
         scaleInvariantFactor *= gl_Position.w;
         // We also need to take into account the camera's half vertical FOV
-        scaleInvariantFactor *= 0.5 * cameraFovY;
+        scaleInvariantFactor *= cameraFovY;
       } else {
         // Compute inverse aspect ratio
         float invAspect = viewportHeight / viewportWidth;
         // When using orthographic projection, the scaling factor is obtain from
         // the camera projection itself.
         // We also need applied the inverse aspect ratio
-        scaleInvariantFactor *= invAspect / length(projection[0].xyz);
+        scaleInvariantFactor *= 2.0 * invAspect / length(projection[0].xyz);
       }
-      // Apply scale invariant factor and pixel ratio to get the correct height in pixels
-      markerSpacePos *= pixelRatio * scaleInvariantFactor;
+      // Apply scale invariant factor
+      markerSpacePos *= scaleInvariantFactor;
     }
 
     // Compute final vertex position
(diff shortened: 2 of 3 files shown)


packages/regl-worldview/src/camera/camera.js
```diff
@@ -78,6 +78,14 @@ export default (regl: any) => {
...
           return selectors.billboardRotation(this.cameraState);
         },
+
+        isPerspective(context, props) {
+          return this.cameraState.perspective;
+        },
+
+        fovy(context, props) {
+          return this.cameraState.fovy;
+        },
```
troygibb: Just thinking out loud... presumably we could get these values from the `projection` value above, since that is derived from `cameraState`, is that correct? Not that we would want to do it that way.

It's curious to me that we haven't had to pass down these values before. I guess we haven't had to manipulate marker props so directly before (`scale` in this case).

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -228,12 +258,12 @@ const frag = `
     // depth test don't work together nicely for partially-transparent pixels.
     float edgeStep = smoothstep(1.0 - cutoff - fwidth(dist), 1.0 - cutoff, dist);
 
-    if (scaleInvariant && vBillboard == 1.0 && scaleInvariantSize < 0.03) {
-      // If scale invariant is enabled and scaleInvariantSize is "too small", do not interpolate 
-      // the raw distance value since at such small scale, the SDF approach causes some 
+    if (scaleInvariant && vBillboard == 1.0 && scaleInvariantSize <= 20.0) {
```
troygibb: this is a pretty big bump from 0.03-- why the change?

hhsaez:
> Is that a regression?

Yes, sorry. I was using pixel ratio which is not the correct way and that's why there were so many differences in the tests. I fixed that in my latest commit.

packages/regl-worldview/src/camera/camera.js
hhsaez: Indeed, I haven't seen those properties being used anywhere else. GLText does a lot of things differently than other commands like Cubes or Lines. 
While `fovy` can be derived from the perspective projection matrix, it does requires a bit math and I prefer not to make shaders more complex than what they are right now. `isPerspective` cannot be derived, though.

packages/regl-worldview/src/commands/GLText.js
hhsaez: It's higher because this new approach does not normalize `scaleInvariantSize` by dividing it by the screen height before sending it to the shaders. The end end result is the same as before, to avoid using SDF if the character size is too small (less than 20 pixels).

packages/regl-worldview/src/stories/GLText.stories.js
```diff
@@ -18,13 +18,22 @@ function textMarkers({
...
   background?: ?boolean,
+  randomScale?: ?boolean,
 }) {
   const radius = 10;
   const count = 10;
+  const scale = (i: number) => {
+    if (!randomScale) {
+      return { x: 1, y: 1, z: 1 };
+    }
+    // It's not really random. Otherwise, screenshot tests won't work correctly
```
janpaul123: You could also use the `rng` helper function which uses a fixed seed in screenshot tests 😄 https://github.com/cruise-automation/webviz/blob/94302209e46a6787216ad9aea2ed1d8f9cc69d1c/packages/regl-worldview/src/stories/util.js#L24

hhsaez: That's great! Thanks!

PR #382, opened 2020-03-26
Fix overlapped markers when using GLText

hhsaez:
## Summary

When rendering text using `<GLText />`, if two or more markers share the same position they will be displayed one on top of the other, which makes it harder to read.

The following changes avoid the overlap by rendering new markers in different lines (provided they share the same position, of course). 

## Test plan

Added new stories for testing this behavior in particular.

## Versioning impact

Patch


hhsaez:
> Just double-checking in case I misread something here, but this still works with hitmaps right?

Yes. Markers are rendered to the hitmap in their new position, so it's still possible to select them.

PR #383, opened 2020-03-26
fixes debounced paint issues

josh-ribakoff-cruise:
## Summary

Changes the debounced paint to ~use a home made throttle~  ~use a RAF game loop~ use RAF, so drawing can be sync'd with request animation frame loop instead of firing 10ms into a frame with a timer event, boosts perf by around 20% in certain use cases. Avoids a bug where consumers calling `onDirty` too rapidly can prolong painting indefinitely.

Avoids back pressure when consumers try to schedule paint operations more frequently than the browser can handle.

## Test plan

npm link in RA app

## Versioning impact

minor


packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -227,7 +227,22 @@ export class WorldviewContext {
...
+   * Be careful to not use debounce which can delay paint indefinitely.
+   * Also be careful not to use something like Lodash throttle (which uses debounce)
+   * which as of 3/26/2020 appears to be erroneously using setTimeout instead of
+   * requestAnimationFrame. We want the latter so that paint calls can be sync'd
+   * with the browser's animation loop for maximum performance.
+   */
+  _throttledPaint = () => {
+    const now = performance.now();
+    if (now - this._lastPaintCall < 10) {
+      console.warn("Worldview paint has been throttled");
```
janpaul123: Might be good to not log this warning since it might get a bit spammy for heavy applications, or maybe do it conditionally using `process.env.NODE_ENV !== "production"` or something. Not sure how we do that in other places.

janpaul123:
I don't know much about this code so I tagged some others who do.

packages/regl-worldview/src/WorldviewContext.js
josh-ribakoff-cruise: I had considered this, but I leaned towards logging a warning so a consumer can know they are doing something wrong, which is arguably better than not knowing.

As a consumer I'd rather get a warning than have my app mysteriously not update (due to the debounce).

Maybe we don't need it, but then consumers could just cargo cult their usage of the paint call, calling it all over the place when it isn't needed. I guess that would be fine now that it uses throttle instead of debounce, but feels less ergonomic than with the warning, which was inspired by a similar warning in one of our internal repos. Ideally we'd surface to the consumer's developers that they are doing something wrong (calling paint out of sync), which causes a perf issue.

packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -80,6 +79,7 @@ function compile<T>(regl: any, cmd: RawCommand<T>): CompiledReglCommand<T> {
 // draw calls, hitmap calls, and raycasting.
 
 export class WorldviewContext {
+  _lastPaintCall: number;
```
vidaaudrey: nit: 
```suggestion
  _lastPaintCallTime: number = performace.now();
```

packages/regl-worldview/src/WorldviewContext.js
vidaaudrey: +1.

janpaul123: Ahh I see, yeah I don't know this code well enough to realise that that was actually going to be incorrect behaviour. In that case it might be good to include some actionable information in the warning message to let the developer know what they should do to fix the issue.

packages/regl-worldview/src/WorldviewContext.js
josh-ribakoff-cruise: That could throttle the first call to paint, I think. Best to leave it off, I think

josh-ribakoff-cruise:
@hhsaez 

> is there a reason we're using 10ms
I don't know, but previously the code had a 10ms debounce. I assume the reason 10ms was chosen is because frames can be a bit faster than 16ms sometimes.

> these changes avoid invoking paint() too frequently
No, my changes fix a bug with existing logic that was trying to do that.

Longer explanation:
The problem with the debounce that was here before is that if you called it more frequently than every 10ms, it could never paint, or paint very infrequently because you tried to call it too frequently, which is a bit sadistic if intentionally designed that way.

This PR does not change the frequency anything gets called (or more accurately does not change the frequency at which you can call paint).

This PR just ensures that if you accidentally call paint more frequently than every 10ms, it will still paint (and warn you) as opposed to potentially never painting, or painting very infrequently.

Also, for apps optimizing for latency, it eliminates a 10ms delay that does not need to be there.

More anecdotally, when the 10ms debounce timer would fire close to an animation frame but not in sync with the animation frame, it would cause frames to take 2x longer to render. I'm seeing around a 20% FPS gain with these changes, anecdotally.

josh-ribakoff-cruise:
I'm realizing I may be unclear on the recommended usage pattern.

Is it recommended to call paint() only in a requestAnimationFrame loop, from the top level of the app, so it is only called once every 16ms?

Or is it recommended for deeply nested components to each call paint() directly, so paint() is called multiple times per animation frame?

Or, is it recommended to call paint() outside of a requestAnimationLoop callback? If so, with my changes that type of paint could be delayed potentially indefinitely until something else happens to call paint, so it would just trade one problem for another.

It may be best to first define the expected use cases we need to support, before we decide whether to warn, and if we need to add a "trailing edge" to support consumers who call outside of the animation frame context.

For my use case, the changes in this PR are a net win, but I don't know how others may be using the library.

packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -227,7 +232,16 @@ export class WorldviewContext {
     this.counters.render = Date.now() - start;
   }
 
-  _debouncedPaint = debounce(this.paint, 10);
+  tick = () => {
+    cancelAnimationFrame(this._frame);
```
jtbandes: Since the raf callback doesn't capture any local state, why do you need to cancel the pending frame if there is one already?

packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -227,7 +232,16 @@ export class WorldviewContext {
...
+    this._frame = window.requestAnimationFrame(() => {
+      if (this._dirty) {
+        this.paint();
+        this._dirty = false;
+      }
+      this.tick();
```
jtbandes: Was this recursive call intentional? A user of Worldview whose data never updates shouldn't have to pay the cost of re-rendering at 60fps.

jtbandes:
> Is it recommended to call paint() only in a requestAnimationFrame loop, from the top level of the app 

Historically, I don't think calling `onDirty()` (assuming that is what you mean by `paint()`?) in a loop was ever a recommended usage pattern. I believe we added it for things like map tiles and GLTFScene, which load _static_ external resources, so they need to re-render _once_ when the data loads. The usage pattern we always used was a top-down update, where the parent of Worldview would re-render whenever the data needed to change.

e.g. instead of

```js
function Child1() {
  const externalData = useMyExternalFrequentlyUpdatingState1()
  const worldviewCtx = useRef()
  useLayoutEffect(() => worldviewCtx.current?.onDirty(), [externalData])
  return <WorldviewCtx>{(ctx)=>{
    worldviewCtx.current=ctx
    return <MyPrimitive>{externalData}</MyPrimitive>
  }}</WorldviewCtx>
}
function Parent() {
  return <Worldview>
    <Child1 />
    <Child2 />
  </Worldview>
}
```
, which is pretty awkward, you'd have

```js
function Parent() {
  const externalData = useMyExternalFrequentlyUpdatingState()
  return <Worldview>
    <Child1>{externalData.child1Data}</Child1>
    <Child2>{externalData.child2Data}</Child2>
  </Worldview>
}
```

, which Just Works™.

That said, I think it's reasonable that some apps would prefer a design where the state is held internally in child components; the side effect is that now we have to actually define/optimize for this usage pattern. So thanks for doing the hard work of thinking through it carefully.

> if we need to add a "trailing edge" to support consumers who call outside of the animation frame context

I think a trailing edge would always be necessary. The use case for `onDirty` was a way of telling Worldview that its children need to be redrawn because they changed internally. If there's no trailing edge on the throttle, then some changes might not ever get rendered.

packages/regl-worldview/src/WorldviewContext.js
josh-ribakoff-cruise: Yes it's intentional. They won't pay the cost, this will "tick" but not "paint" if the "dirty" flag has not been set since the last "paint".

 I added `console.log("tick")` and `console.log("paint")` to verify that if I call `onDirty` at 10fps, the loop "ticks" at 60fps, but paint only gets called at 10fps.

The cost of checking the boolean `dirty` flag at 60fps should be nominal

packages/regl-worldview/src/WorldviewContext.js
josh-ribakoff-cruise: Defensive programming in case `tick()` gets called a second time, it won't have two loops going

packages/regl-worldview/src/WorldviewContext.js
jtbandes: I still think it's a waste of CPU that we should avoid

packages/regl-worldview/src/WorldviewContext.js
jtbandes: I just mean `if (!this._frame) this._frame = ...`

josh-ribakoff-cruise: Clearing a pending async operation is idempotent, and also when the async callback runs or is canceled, that binding will still be truthy... I chose to just always clear the pending async operation before setting one, vs adding extra state to track if there is a pending async operation, if that's what you're asking.

packages/regl-worldview/src/WorldviewContext.js
josh-ribakoff-cruise: Hmm, I'm surprised to hear that you feel that way, it's only taking 9 μs every 16ms, that's 9 millionths of a second of CPU only, and the overall changes improve performance for apps that use `onDirty`... What would you suggest instead?

josh-ribakoff-cruise:
> Imagine 5 map tiles finish loading in quick succession; you might only render the first one if all the other calls get throttled.

That's a non issue with the latest change on this PR that use a RAF loop.

> I believe we added it for things like map tiles and GLTFScene, which load static external resources, so they need to re-render once when the data loads.

Interesting, thanks for that context. As an aside, I do think that in a scenario with lots of static resources loading over time, the painting should happen in an animation frame event, not a timer event for the "best" performance...

packages/regl-worldview/src/WorldviewContext.js
jtbandes: I guess I'm just not sure if it might be slightly more efficient to keep the existing request rather than making a new request, if you're calling this frequently. probably doesn't matter much.

packages/regl-worldview/src/WorldviewContext.js
jtbandes: I think it just makes me sad when web apps do things on recurring timers or tight loops and it slows down my browser 😛 Keep in mind that some apps have several Worldviews on a single page.

Doesn't it still work if you just delete this line? If someone calls onDirty, that will lead to a paint call on the next available animation frame. If they don't, nothing else happens.

Might also be worth looking at debouncePromise: https://github.com/cruise-automation/webviz/blob/master/packages/webviz-core/src/util/debouncePromise.js

packages/regl-worldview/src/WorldviewContext.js
josh-ribakoff-cruise: Yeah, in practice we should never get into that state, it's kind of like the trolley problem, if it actually happens, something much worse is probably wrong, but it's good to have a defense in case.

packages/regl-worldview/src/WorldviewContext.js
josh-ribakoff-cruise: I agree, but keep in mind that's why I'm using RAF, to ensure that consumers calling onDirty more frequently than the browser can handle does not cause any performance degradation.

Right now someone can call this every 12ms, and it can take 50ms each time to run, which will segfault the browser due to memory leaks / back pressure

This change will eliminate the possibility of the type of performance issue you're describing.

If we delete the recursion, this will then run 1x only. That means only the first map tile loading would cause a paint, and subsequent map tile loading would not cause a paint. We need this loop to keep checking if there's something new to paint, I'm pretty sure. Also like I'm saying since it's a RAF, it should adaptively be called less often in a CPU constrained browser, so I'm not personally concerned.

Re: debounced promise, this is a long lived loop so if you're looking for a more elegant abstraction I would recommend looking into RxJS with a RAF scheduler, since observables are like long lived promises that emit 0 to N times over time, unlike a promise which just emits once. RxJS has several built in throttle/debounce timers, but I'm contending we specifically don't want a debounce or throttle.

A throttle would bake in assumptions about how CPU constrained the browser is, or how much work is being done. It would still allow the method to be called more frequently than the browser can handle it, and crash the browser or degrade performance.

A debounce would mean if it gets called more frequently than the debounce interval, it just never paints.

The RAF loop avoids both problems, and ensures we never force feed the browser more work than it can handle, because we won't schedule another frame until the previous frame completes.

I can empathize with you here though, and could envision a middle-ground where we have `onDirty` sometimes paint synchronously, and sometimes schedule a RAF callback, depending on the delta between when it was last called vs when it's currently being called. It would a lot of complexity to the code, and would make little difference, in my opinion, but I'd be happy to make the change if you feel strongly.

jtbandes: Disclaimer: I am not really an owner of this library anymore, so I think you should feel free to abandon this conversation and take someone else's opinion if you're blocked. I'm just speculating as a bystander. I find the conversation interesting but I don't want to slow you down too much :)

> If we delete the recursion, this will then run 1x only. That means only the first map tile loading would cause a paint, and subsequent map tile loading would not cause a paint. 

I don't think this is right in general. Suppose one map tile loads, and calls onDirty. If the second one loads before the animation frame, great, its onDirty will do nothing (or replace the current frame request, but that's effectively nothing) and the next animation frame will repaint both of them (effectively using the browser's animation timer as a way of batching the two updates). On the other hand, if the second one loads **after** the first animation frame, its onDirty call will just schedule another frame. That's fine too, since the paint calls won't actually happen faster than the browser can render them.

Since paint is a synchronous call, I'm ignoring the case of "the second one loads **during** the first animation frame". On the other hand, the [debouncePromise code](https://github.com/cruise-automation/webviz/blob/master/packages/webviz-core/src/util/debouncePromise.js) I wrote was meant to handle the case of an asynchronous bit of work, and that's why it has a second flag for `callPending`. With async work (paint), if a second call comes in while the first one is being serviced (painted), you need to give the second one another opportunity to run, but only after the first one has finished (since you don't want to be painting 2 things at once).

I admit `debouncePromise` was probably not the best name (I think I just forgot / didn't know the terminology of debounce vs throttle). I agree with your points about "classic" throttle and debounce both making assumptions about performance limits or leaving it open to perf degradations. My `debouncePromise` doesn't have a timer or animation frame, though, it exists only for the purpose of batching requests while ensuring there is always a trailing-edge call. I also realize I should've added more comments to it 😓

You might be able to tweak your current implementation pretty easily, since the Worldview paint() is synchronous. On the other hand you might be able to use debouncePromise easily too, like this:

```js
function rafPromise(): Promise<void> {
  return new Promise((resolve) => window.requestAnimationFrame(resolve));
}
debouncedPaint = debouncePromise(async () => {
  await rafPromise();
  paint();
});
```

josh-ribakoff-cruise: I like all the feedback I can get :) So no worries.

> If the second one loads after the first animation frame, its onDirty call will just schedule another frame

I think that's the part you may be missing here. `onDirty()` just toggles a boolean now, with my latest changes.  Then separately the RAF loop which is always running checks the `dirty` flag and conditionally paints. `onDirty()` isn't scheduling any work directly, just setting that flag for the RAF loop to read on it's next tick. Does that change your feedback?

josh-ribakoff-cruise:
Anyone have any idea what's going on w/ the CI failures?

```
[reg-suit] debug reg-simple-keygen-plugin is inialized with: 
{
  "expectedKey": "a3fc95267f9eb68530cbb0da27a1eb912032ee71",
  "actualKey": "036e269d298bb2429182eaca444ed44322d31da2"
}
[reg-suit] debug reg-publish-s3-plugin is inialized with: 
{
  "bucketName": "open-source-webviz-ui"
}
[reg-suit] debug reg-notify-github-plugin is inialized with: 
{
  "clientId": "MzQzsDQzNjc30y9PTSrLrNI3MzQxNDbUTy4qzSxO1U0sLcnPTSzJzM8DAA==",
  "setCommitStatus": false
}
[reg-suit] debug userDirs: 
{
  "actualDir": "/home/circleci/project/__screenshots__"
}
[reg-suit] debug workingDirs: 
{
  "base": "/home/circleci/project/.reg",
  "actualDir": "/home/circleci/project/.reg/actual",
  "expectedDir": "/home/circleci/project/.reg/expected",
  "diffDir": "/home/circleci/project/.reg/diff"
}
[reg-suit] info Detected the previous snapshot key: 'a3fc95267f9eb68530cbb0da27a1eb912032ee71'
{ CredentialsError: Missing credentials in config
    at IncomingMessage.<anonymous> (/home/circleci/project/node_modules/aws-sdk/lib/util.js:865:34)
    at IncomingMessage.emit (events.js:194:15)
    at IncomingMessage.EventEmitter.emit (domain.js:441:20)
    at endReadableNT (_stream_readable.js:1125:12)
    at process._tickCallback (internal/process/next_tick.js:63:19)
  message: 'Missing credentials in config',
  retryable: false,
  time: 2020-03-31T18:50:47.068Z,
  code: 'CredentialsError',
  originalError:
   { message: 'Could not load credentials from any providers',
     retryable: false,
     time: 2020-03-31T18:50:47.067Z,
     code: 'CredentialsError' } }
```

packages/regl-worldview/src/WorldviewContext.js
jtbandes: I see, yeah, I guess I'm arguing for going back to something like this:

```js
onDirty = () => {
  if (!this._frame) this._frame = requestAnimationFrame(this.paint);
}
componentDidUpdate() {
  if (!this._frame) this._frame = requestAnimationFrame(this.paint);
}
paint = () => {
  /*webgl rendering here*/
  this._frame = null;
}
```

Do you think this implementation would still be problematic for your use case? It seems like the best of both words, where the actual paints will never be delayed longer than necessary, and you can call onDirty as often as you like.

josh-ribakoff-cruise: I think that's solid feedback, I updated the PR and I can see in both the perf tab & through manual testing the perf is even better now. I of course had to modify your psuedo code to wrap this.paint in an arrow function so RAF doesn't set the context for `this` to the window object. Thanks for your feedback!

packages/regl-worldview/src/Worldview.js
```diff
@@ -153,24 +153,7 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
...
-    // queue up a paint operation on the next frame, if we haven't already
-    if (!this._tick) {
-      this._tick = requestAnimationFrame(() => {
-        this._tick = undefined;
-        try {
-          worldviewContext.paint();
-        } catch (error) {
-          // Regl automatically tries to reconnect when losing the canvas 3d context.
-          // We should log this error, but it's not important to throw it.
-          if (error.message === "(regl) context lost") {
```
jtbandes, line 166: I think you might need to keep this somewhere, maybe @davidswinegar can advise more but I think this error was causing lots of problems for us in sentry.

danyim: +1 to keeping this filter as we experienced the same spam on our team's error reporter.

janpaul123:
Any updates on this? Happy to merge once @jtbandes is happy with the code!

packages/regl-worldview/src/Worldview.js
josh-ribakoff-cruise: I'm not an expert on WebGL but this is a code smell from an error/flow control perspective. Generally when a program has an "error" and happily continues, that to me suggests something is wrong.  Is it possible we're not discerning between the "expected" error, and an actual error here? What if the reconnection code in regl actually failed, would it just throw the same error?

I'll add this logic back, in the `WorldviewContext`, for now... but can't we also just mute errors in Sentry? Or alternatively, would it make sense to send a PR to regl to not throw the error if there was a successful reconnect?

janpaul123: > would it make sense to send a PR to regl to not throw the error if there was a successful reconnect

I remember us doing that, or maybe we only talked about it? I think it is a mistake in regl that should ideally be fixed there. @jtbandes @brianc do you remember if we ended up making a PR to regl?

jtbandes: Or @troygibb? https://github.com/cruise-automation/webviz/pull/265

brianc: I _asked them_ about making a PR for a [separate issue](https://github.com/regl-project/regl/issues/533) (not having an unstoppable `console.error` if you double-dispose of a texture).  I never talked w/ them about a context lost error.  That's a bit alarming to just swallow it entirely as sometimes if the context is lost...it's gone for good and yer app crashes.  Maybe regl throws the same error somewhere else if context is lost?

josh-ribakoff-cruise: Thanks for the info. Yeah agreed, but at least this decision is decoupled from my PR now.

josh-ribakoff-cruise:
@janpaul123 any objections to merging?

janpaul123:
@josh-ribakoff-cruise In CI it looks like Flow is unhappy. Looks like a pretty minor issue. But also can you push this as a branch directly to this repo, so that CI will finish properly and we'll see if any screenshots have changed? (That's a major part of our testing strategy.) I invited your account `josh-ribakoff-cruise` for write access to this repo, which should allow you to push a branch (once you accept that invite). Let me know if you got that invite or if I should send it again.

joshribakoff:
I'm closing this one & re-opening a new PR from my other account so we can look into the CI details. https://github.com/cruise-automation/webviz/pull/403

This PR has also become hard to follow & there's a lot of outdated information about my changes, so the new PR should be a better description of the changes.

PR #384, opened 2020-03-27
<GLText> Disable depth testing when scale invariance is enabled 

hhsaez:
## Summary

These changes help `<GLText />` markers to be rendered on top of other objects in the 3D world. 

Before:
Screen Shot 2020-03-27 at 10 40 27 AM

After:
Screen Shot 2020-03-27 at 10 40 30 AM

In order to render the text on top of other objects, depth testing needs to be disabled. 

In addition, `<GLText />` should be drawn last, which is something application developers must do themselves. I added an entry in the documentation about this.

## Test plan

Added three new stories to show how `<GLText />` is rendered when other objects are overlapping and whether or not scale invariance is enabled.

## Versioning impact

Patch

### DIFF
--- a/docs/src/4.15.GLText.mdx
+++ b/docs/src/4.15.GLText.mdx
@@ -10,15 +10,19 @@ The prop types are identical to `<Text />`, except that additionally each Text o
 
 ## Scale Invariance
 
-`<GLText />` also supports rendering text at a constant scale regardless of the zoom level by using the optional `scaleInvariantFontSize?: number` property. If set (default is `undefined`), the text will be rendered always at the same size, resembling the behavior of the `<Text />` command. 
+`<GLText />` also supports rendering text at a constant scale regardless of the zoom level by using the optional `scaleInvariantFontSize?: number` property. If set (default is `undefined`), the text will be rendered always at the same size, resembling the behavior of the `<Text />` command.
 
 Please note that `scaleInvariantFontSize` only works if the `billboard` property is set to true.
 
 <GLTextScaleInvariant />
 
+### Rendering text on top of other objects
+
+If scale invariance is enabled, you should make sure `<GLText />` is the last command being rendered in your world representation. That will make all text markers to be drawn on top of other visible objects.
+
 ## Resolution
 
-By default, `<GLText />` uses a high resolution font texture atlas for rendering text, which may have an impact on the performance of your application. 
+By default, `<GLText />` uses a high resolution font texture atlas for rendering text, which may have an impact on the performance of your application.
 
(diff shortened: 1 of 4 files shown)


packages/regl-worldview/src/commands/GLText.js
```diff
@@ -297,7 +297,11 @@ function makeTextCommand(alphabet?: string[]) {
     const atlasTexture = regl.texture();
     const makeDrawText = (isHitmap: boolean) => {
       return regl({
-        depth: defaultDepth,
+        // When using scale invariance, we want the text to be drawn on top
+        // of other elements. This is achieved by disabling depth testing
+        // In addition, make sure the <GLText /> command is the last one
+        // being rendered.
+        depth: command.scaleInvariant ? disabledDepth : defaultDepth,
```
vidaaudrey: Curious if you have considered using `layerIndex`?

hhsaez: Disabling depth testing is required regardless of the rendering order. Then, `layerIndex` should help for the second part of my comment, where we need to make sure `<GLText />` is the last command to be rendered, but I preferred not to enforce it on the command itself. What do you think?

docs/src/4.15.GLText.mdx
```diff
@@ -10,25 +10,29 @@ The prop types are identical to `<Text />`, except that additionally each Text o
...
-`<GLText />` also supports rendering text at a constant scale regardless of the zoom level by using the optional `scaleInvariantFontSize?: number` property. If set (default is `undefined`), the text will be rendered always at the same size, resembling the behavior of the `<Text />` command. 
+`<GLText />` also supports rendering text at a constant scale regardless of the zoom level by using the optional `scaleInvariantFontSize?: number` property. If set (default is `undefined`), the text will be rendered always at the same size, resembling the behavior of the `<Text />` command.
 
 Please note that `scaleInvariantFontSize` only works if the `billboard` property is set to true.
 
 <GLTextScaleInvariant />
 
+### Rendering text on top of other objects
+
+If scale invariance is enabled, you should make sure `<GLText />` is the last command being rendered in your world representation. That will make all text markers to be drawn on top of other visible objects.
```
troygibb:
```suggestion
If scale invariance is enabled, you should make sure `<GLText />` is the last command being rendered in your world representation. That will draw these text markers on top of other visible objects.
```

packages/regl-worldview/src/commands/GLText.js
vidaaudrey: 👍 Makes sense

PR #378, opened 2020-03-25
Update packages/webviz-core from internal repo

janpaul123:
Changelog:
- Added support for streaming a second bag using the `remote-bag-url-2` query parameter in the URL. #374
- Fixed crashing on bags with topics that don't have a message definition. #373
- Various performance improvements.
- Improved caching of deserialized messages should result in fewer browser crashes.
- Added `unlimitedMemoryCache` experimental feature to load everything in memory without regard for caching limits. Useful for when you have a machine with lots of memory and you're explicitly OK with Webviz taking up a lot of it. Use at your own risk; this might crash your browser!
- Included an "Add Topics" option for adding markers to the Image panel.
- Made performance improvements to the Plot and State Transitions panels. If you notice any issues, these performance improvements can be disabled in the Experimental Features menu, under "Use a web worker to render the Plot panel".
- In the Image panel, we again filter available marker topics by the namespace of the currently selected camera.
- Added support for publishing messages over the Websocket connection using a Publish panel. #323

PR #388, opened 2020-03-30
Use webpack aliasing instead of npm aliasing

davidswinegar:
I've had a lot of trouble getting this aliasing to update across yarn
and package-lock for the hotloader version of react-dom. Instead
load the package directly and alias it in webpack.

Test plan: CI


janpaul123:
Great, it looks like this would also fix my screenshots in #378! Just need to do a `npm run lint:fix`, and then make sure that we do the same in our internal repo so that stays in sync.

davidswinegar:
@janpaul123 I actually think I tracked down the root cause of why NPM aliasing was broken, and I think I'm going to fix that instead.

PR #362, opened 2020-03-12
Update package versions from internal

davidswinegar:
Update these package versions, which were out of date with our
internal repo.

PR #389, opened 2020-03-31
Update docker CI image

davidswinegar:
This is a fix for a long-winded attempt to get packages synced between
internal and open source versions. We use NPM aliasing internally, but
it doesn't seem to work in open-source - see #362, which keeps failing
the build in CI even though the build works locally. It turns out that the
version of NPM in CI (6.4.2) doesn't support package aliasing while
my local version of NPM does - support was only added in 6.9.0 
(https://npm.community/t/release-npm-6-9-0/5911).

The fix is simple: update the dockerfile version that we use to
be able to use package aliasing in CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,8 @@ jobs:
       - run: echo $REG_SUIT_EXPECTED_KEY
 
       - run:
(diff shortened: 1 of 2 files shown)


davidswinegar:
This should fail CI until I actually push the dockerfile - going to wait for a thumbs up before doing so.

janpaul123:
Oh maybe those screenshots are due to very minor font changes in the new docker image?

PR #403, opened 2020-04-17
Changing onDirty & paint to be more performance friendly

joshribakoff:
This is a continuation of https://github.com/cruise-automation/webviz/pull/383

## Summary

As a consumer of Worldview, I was debugging performance. I found something I think is a performance anti-pattern & should be changed. I see we use a Lodash debounce on our `onDirty` which calls paint, I think this should change.

Issue # 1 - There were frames that would take so long to render that their durations would sometimes exceed a whole second (FPS dropping below 1 FPS). 

Screen Shot 2020-03-25 at 4 33 02 PM

Issue # 2 - I also noticed there was a lot of work happening on the main thread, which seemed animation related, but was not running in an animation frame event, but instead was happening in a timer event. This would mean we delay painting & miss the animation frame events, which would extend painting until the next frame, delaying the frame duration. It would also paint sometimes multiple times per animation event.

Screen Shot 2020-03-25 at 4 26 42 PM

Issue # 3 - I also noticed, if you apply CPU throttling or give Worldview a complex enough scene such that painting takes longer than 10ms, Worldview will try to paint every 10ms & creates back pressure, which is where the main thread is being force fed work to do faster than it can do it - as a result real time updates can lag behind, and the app can feel sluggish.

This PR gets rid of the debounced paint method bound to the `onDirty` callback. It instead makes the `onDirty` callback schedule an animation frame to paint if there is not already one scheduled.

If you call `onDirty` very rapidly like in issue # 1 - it will paint on every animation frame event the browser fires, painting will now no longer be delayed up to indefinitely.

If you call `onDirty` less often than the RAF events fire, you should still see benefits because the browser is better able to synchronize the frame rate, and work isn't being done after the animation event, delaying the frame/painting, so issue # 2 should be solved.

Issue # 3 should also be solved because if the painting takes longer, the browser will not fire any animation frame events until it is ready for more work, so we will no longer force feed the main thread causing back pressure.

This also solves some other issues by consolidating the multiple places we control painting within Worldview, to prevent other scenarios where Worldview can paint multiple times per frame.

## Test plan

_How was this change tested & verified?_

## Versioning impact

minor

### DIFF
--- a/packages/regl-worldview/src/WorldviewContext.js
+++ b/packages/regl-worldview/src/WorldviewContext.js
@@ -212,6 +212,8 @@ export class WorldviewContext {
     try {
       this._paint();
     } catch (error) {
+      // Regl automatically tries to reconnect when losing the canvas 3d context.
+      // We should log this error, but it's not important to throw it.
       if (error.message === "(regl) context lost") {
         console.warn(error);
       } else {
--- a/packages/regl-worldview/src/WorldviewContext.js
+++ b/packages/regl-worldview/src/WorldviewContext.js
@@ -82,7 +82,7 @@ export class WorldviewContext {
   _commands: Set<RawCommand<any>> = new Set();
   _compiled: Map<Function, CompiledReglCommand<any>> = new Map();
   _drawCalls: Map<React.Component<any>, DrawInput> = new Map();
-  _frame: AnimationFrameID | null;
+  _frame: ?AnimationFrameID;
   _paintCalls: Map<PaintFn, PaintFn> = new Map();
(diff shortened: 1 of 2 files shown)


packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -207,6 +209,18 @@ export class WorldviewContext {
   };
 
   paint() {
+    try {
+      this._paint();
+    } catch (error) {
+      if (error.message === "(regl) context lost") {
```
troygibb, line 215: any chance you could persist the comment added in `Worldview.js` so we keep the context?

jtbandes:
Thanks for pushing this through! Nice API improvement.

PR #413, opened 2020-04-30
Disable hitmap on mouse down and/or after dragging

hhsaez:
## Summary

Rendering the hitmap can be a very expensive operation (i.e. big point clouds) and right now we're generating it on both `onMouseDown` and `onMouseUp` in some cases. This leads to a delay whenever the user interacts with the 3D view:

1. User clicks LMB. We receive an `onMouseDown` event and the hitmap is rendered
    1. User cannot drag the view until the hitmap is rendered. This can take several seconds if the dataset is big enough
1. User drags the view by moving the mouse and holding LMB. Hitmap is not rendered unless `hitmapOnMouseMove` is `true`.
    1. User can drag the view normally without delay, provided `hitmapOnMouseMove` is not `true`. 
1. User releases LMB. We receive an `onMouseUp` event and the hitmap is rendered again.
    1. User cannot perform another action until the hitmap is rendered again. This also can take several seconds.

In this PR I'm extending Worldview to allow customizing which mouse events should be ignored when generating the hitmap. For that reason, I'm introducing a new property: `disableHitmapForEvents`, which is an array of the mouse events that should not trigger a hitmap generation.

By default, we don't want to render the hitmap if the user moves the mouse. 

In addition, if we also need to avoid rendering the hitmap when dragging the mouse, we will use the property as follows:

```
<Worldview disableHitmapForEvents={["onMouseDown", "onMouseMove", "onMouseUp"]} />
```
The above code will only render the hitmap whenever we receive an `onClick` event, improving performance and user experience in general.

## Test plan

* Existing tests and stories should pass
* Added new stories
* Manually, when using a big point cloud. Just drag/zoom/rotate the view and it will be more fluid now. 

## Versioning impact
* Minor: It does not break backward compatibility


packages/regl-worldview/src/Worldview.js
```diff
@@ -38,6 +38,10 @@ export type BaseProps = {|
   backgroundColor?: Vec4,
   // rendering the hitmap on mouse move is expensive, so disable it by default
   hitmapOnMouseMove?: boolean,
+  // we avoid rendering the hitmap on mouse down by default
+  hitmapOnMouseDown?: boolean,
+  // we avoid rendering the hitmap after dragging by default
+  hitmapAfterMouseDrag?: boolean,
```
davidswinegar, line 44: can we instead have a `disableHitmapForEvents: MouseEventEnum[]` prop, and default it to `["mouseMove"]`, but overridable using `hitmapOnMouseMove`? And then treat `hitmapOnMouseMove` as legacy and log a warning when it is used, and mark it as deprecated in the docs? We'll also have to add these props to the docs.

As currently written this will change the default hitmap behavior for anyone upgrading the library, so I'm a bit leery of adding it like this. @troygibb what do you think?

hhsaez: Yes, I think that would be great. I also prefer not to change the current behavior.

hhsaez: @davidswinegar Check out my latest changes. I added the `disableHitmapForEvents` prop, as well as new stories and docs. 
If `hitmapOnMouseMove` is provided, it will override `disableHitmapForEvents` on mouse move.

docs/src/3.1.Worldview.mdx
```diff
@@ -11,6 +11,7 @@
 | `shiftKeys`           | `boolean`                                       | true                   | enable/disable shift modifier for keyboard controls                                                     |
 | `backgroundColor`     | `Vec4`                                          | `[0,0,0,1]`            | change the scene background color                                                                       |
 | `hitmapOnMouseMove`   | `boolean`                                       |                        | if enabled, the clicked object will be returned from the mouse event handler. Disabled by default for performance reasons. |
+| `disableHitmapForEvents`   | `[MouseEventEnum]`                         |                        | disable rendering the hitmap for some mouse events for performance reasons. Only "onMouseMove" is disabled by default. |
```
janpaul123, line 14:
```suggestion
| `disableHitmapForEvents`   | `[MouseEventEnum]`                         | ["onMouseMove"]    | disable rendering the hitmap for some mouse events for performance reasons. Only "onMouseMove" is disabled by default. |
```

docs/src/3.1.Worldview.mdx
```diff
@@ -11,6 +11,7 @@
 | `shiftKeys`           | `boolean`                                       | true                   | enable/disable shift modifier for keyboard controls                                                     |
 | `backgroundColor`     | `Vec4`                                          | `[0,0,0,1]`            | change the scene background color                                                                       |
 | `hitmapOnMouseMove`   | `boolean`                                       |                        | if enabled, the clicked object will be returned from the mouse event handler. Disabled by default for performance reasons. |
```
janpaul123, line 13: This one seems now redundant with `disableHitmapForEvents`?

janpaul123: Ah I see you marked it deprecated below — should we just remove it from the docs then?

packages/regl-worldview/src/Worldview.js
```diff
@@ -201,8 +220,10 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
       return;
     }
 
-    // rendering the hitmap on mouse move is expensive, so disable it by default
-    if (mouseEventName === "onMouseMove" && !this.props.hitmapOnMouseMove) {
+    // Rendering the hitmap is expensive, so we disable it for some events
+    // Rendering the hitmap on mouse move is disabled by default.
+    const { hitmapOnMouseMove, disableHitmapForEvents = ["onMouseMove"] } = this.props;
```
janpaul123, line 225:
```suggestion
    const { disableHitmapForEvents = hitmapOnMouseMove ? [] : ["onMouseMove"] } = this.props;
```

or so? I find the clause below a bit hard to reason about, so it might be better to show exactly how `hitmapOnMouseMove` gets migrated to `disableHitmapForEvents`.

packages/regl-worldview/src/Worldview.js
janpaul123: Hm, this new model doesn't seem to do `hitmapAfterMouseDrag` any more, or does it? We still do render a hitmap on mouse up after a drag now, even though for Webviz we don't really need to any more?

hhsaez: That's right. I just updated the PR description.

`hitmapAfterMouseDrag` is no longer needed, so I removed it in my latest commit. My initial goal was to render a hitmap only if the user has clicked (that is, mouseDown then mouseUp without moving), which is exactly what the `onClick` event does.

Still, if we need to support rendering after a mouse drag, we just do:
```
disableHitmapForEvents = ["onMouseDown", "onMouseMove"];
```

Does that make sense?

docs/src/3.1.Worldview.mdx
hhsaez: I prefer to leave it for now, since the property is deprecated and not really removed. We should remove it on a future major release, though.

packages/regl-worldview/src/Worldview.js
hhsaez: I'm not sure about that change. 
My understanding is that `hitmapOnMouseMove` should override only what `disableHitmapForEvents` does for `onMouseMove`. In other words, if `disableHitmapForEvents` specifies that both `onMouseDown` and `onMouseMove` events should be ignored, using `hitmapOnMouseMove` leads to Worldview ignoring only `onMouseDown` for hitmap rendering.
Based on your suggested change, by using `hitmapOnMouseMove` we will be triggering the hitmap on a `onMouseDown` event too since we're clearing `disableHitmapForEvents`. 
Does this makes sense?

janpaul123: We're deprecating `hitmapOnMouseMove` though right? Since it will be completely superseded by `disableHitmapForEvents`? So I think we can assume that if a user has `disableHitmapForEvents` set, that they won't have `hitmapOnMouseMove` set. In fact, maybe we should throw an error if both are set, to avoid confusion.

docs/src/3.1.Worldview.mdx
janpaul123: Then at least let's update the docs to make very clear that this is deprecated and that moving forward people should use `disableHitmapForEvents` instead. Maybe move it to the bottom of the table, too?

packages/regl-worldview/src/Worldview.js
janpaul123: Ah cool, I didn't realise we already have `onClick`; that makes a ton of sense then. 👍

packages/regl-worldview/src/Worldview.js
```diff
@@ -111,6 +125,18 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
       }
     }
 
+    if (hitmapOnMouseMove) {
+      if (disableHitmapForEvents) {
+        console.error(
```
janpaul123, line 130:
```suggestion
        throw new Error(
```

perhaps even?

PR #415, opened 2020-05-01
Don't ignore keyboard events when the target is a Button or a Select

hassoncs:
Previously we'd ignore events if their target was a Button or a Select, but this breaks keyboard shortcuts once a Button has been focused. In order to keep keyboard shortcuts working all the time, we now allow events targeted at Button and Select elements.

## Test plan

Manually verified

## Versioning impact

It does impact the behavior so maybe minor/patch

### DIFF
--- a/packages/@cruise-automation/button/package.json
+++ b/packages/@cruise-automation/button/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@cruise-automation/button",
-  "version": "0.0.7",
+  "version": "1.0.7",
   "main": "lib/index.js",
   "description": "Cruise button",
   "license": "Apache-2.0",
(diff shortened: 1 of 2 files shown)


hassoncs:
We could keep it backwards compatible, but even coming up with a reasonable prop name makes this functionality seem a bit weird– `ignoreEventsOnNonTextEntryElements`. My vote would be to not add a prop, but I'm willing to do it if we think it would be useful.

brianc:
> If we do this PR I do think we should do a major bump because it's backwards incompatible.

Yeah my thinking is if a major version bump isn't too hard should respect semver & change the behavior without worrying about keeping it backwards compatible.  I don't think this package is widely used internally.  Furthermore, if someone wants to disable key listener behavior on a button, they can p easily wrap it w/ their own event handler to stop propagation.

janpaul123:
Sounds good, let's just do a major version bump then. Thanks both!

PR #416, opened 2020-05-05
Fix gltext recreating draw command every tick

davidswinegar:
When looking at performance tools in Webviz, I could see that the
`makeDrawText` command was being recompiled every tick. Fix this
command so that we never recompile it, saving ~8-15ms per tick (!).

Test plan: this should have no screenshot test changes.


janpaul123:
Whoa wild! We should really set up some automated performance tests so we can prevent stuff like this from regressing…!

davidswinegar:
Looks like there's some issue with the depth. Going to dig in tomorrow.

PR #424, opened 2020-05-11
Remove `event.stopPropagation` from react keylistener

troygibb:
## Summary

The line in our key listener `stopPropagation` was causing an issue with the 3D panel. Since the Worldview implementation of `CameraListener` depends on propagated events, when we added a custom `ctrl + f` handler, it halted the Worldview `f` keydown event (tilt) from propagating to the Worldview way of handling keydown events (Worldview does not add event listeners).

It makes sense to me that the onus of propagating or not propagating events should be on the custom key handler itself, not on the library. I did a sanity check on some other react key handler packages to see if they stopped event propagation, and it seems like we're an outlier here.

Obviously I'm a bit worried about any unintended breakage here of other consumers, I figured I could manually go through all of our instances of it to double check. The other option here was to simply add a `stopPropagation` prop so we could limit the effect of this change to just our 3D panel usage of it, however I thought this was cleaner. 

## Test plan

Verified locally.

## Versioning impact

Minor update.

### DIFF
--- a/packages/react-key-listener/package-lock.json
+++ b/packages/react-key-listener/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "react-key-listener",
-  "version": "0.1.0",
+  "version": "1.1.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/packages/react-key-listener/package-lock.json
+++ b/packages/react-key-listener/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "react-key-listener",
-  "version": "0.0.7",
+  "version": "0.1.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


janpaul123:
Lol looks like we have to change this tiny library a lot lately (see @hassoncs' recent PR as well). I think it's time to just inline this tiny library into our Webviz repo so we can change it more easily. It's hardly worth having it as a library at all, it's so small and simple, so after inlining let's maybe just remove it as a library from this repo?

troygibb:
> Lol looks like we have to change this tiny library a lot lately (see @hassoncs' recent PR as well). I think it's time to just inline this tiny library into our Webviz repo so we can change it more easily. It's hardly worth having it as a library at all, it's so small and simple, so after inlining let's maybe just remove it as a library from this repo?

Yea that sounds good! I'll make the change in Webviz now.

Thoughts on the change itself tho?

Issue #406, opened 2020-04-21
Run webviz from source build without launching docs page


ben-greenberg:
I've built webviz from source and have been using it with great success. The only kink in my system is that currently when I want to open up the webviz application in a browser, I have to run ```npm run docs``` which launches the docs app in addition to running the webviz application.

Is there another way to run just the application without launching the docs page?

janpaul123:
You could host it online somewhere? We host webviz.io/app using Github Pages.

castaway2000:
@janpaul123 I think what they are saying is that they would like to run the webviz App as landing page. Is there an easy way to implement the webviz-core package in a standard html page or modify the existing React application to point the landing page to be the webviz app? 

I was also looking into doing this using the webviz-core package as a static website however it seems the webviz-core package is not able to source /dist/webvizCoreBundle.js and is not in the repo from what I can tell, How is this bundle sourced?

if the bundle was included it would make for an easier solution to the problem by simply running the package as a standalone app.

janpaul123:
You can see here how we publish our [`gh-pages`](https://github.com/cruise-automation/webviz/tree/gh-pages) branch: https://github.com/cruise-automation/webviz/blob/6623f080b2bee11144df1bb70c2ac93db51a37fa/package.json#L135

You can also statically host the gh-pages branch directly yourself. It's what gets hosted on webviz.io. You can see that it has an `app/index.html` that contains the actual application.

janpaul123:
Maybe we should pull that out into its own command so everyone can easily generate a static build? Happy to take PRs for that!

kevinrev26:
What if I wanted to use the individual components of the WebViz Core?

How should I import them?

janpaul123:
We don't currently support that. Those are considered internal API and might change heavily on every release. But if you really want to I guess you can just `import` them like any other component; e.g. `import useDataSourceInfo from "webviz-core/src/PanelAPI/useDataSourceInfo";`

janpaul123:
Btw I set up a Slack workspace to make it easier for us to chat; join us using the link here: #461

Issue #267, opened 2019-11-11
automatic github action to build webviz


stefan-enway:
you need to be in github action beta (will be public November 13th; to days)
```name: Node CI
on: [push]

jobs:
  build:

    runs-on: ubuntu-latest

    strategy:
      matrix:
        node-version: [10.x]

    steps:
    - uses: actions/checkout@v1
    - name: Use Node.js ${{ matrix.node-version }}
      uses: actions/setup-node@v1
      with:
        node-version: ${{ matrix.node-version }}
    - name: npm bootstrap & build
      run: |
        npm run bootstrap
        npm run build --if-present
        #npm run docs (runs webserver)
      env:
        CI: true
    - uses: elgohr/Publish-Docker-Github-Action@master
      with:
        name: cruise-automation/webviz/webwiz:latest
        registry: docker.pkg.github.com
        dockerfile: Dockerfile
        username: <your-automation-username-here->
        password: ${{ secrets.GITHUB_TOKEN }}
```
Dockerfile
```
# Copyright (c) 2018-present, GM Cruise LLC
#
# This source code is licensed under the Apache License, Version 2.0,
# found in the LICENSE file in the root directory of this source tree.
# You may not use this file except in compliance with the License.

# This container is published at https://hub.docker.com/r/cruise/webviz-ci.

FROM node:10.15.3-slim

# Install some general dependencies for stuff below and for CircleCI;
# https://circleci.com/docs/2.0/custom-images/#required-tools-for-primary-containers
RUN apt-get update && apt-get install -yq libgconf-2-4 wget git ssh --no-install-recommends

# Install Google Chrome for Puppeteer.
# https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-in-docker
RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
    && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
    && apt-get update \
    && apt-get install -y google-chrome-stable fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst ttf-freefont --no-install-recommends
ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true

ENV WEBVIZ_IN_DOCKER=true

[...]

Issue #120, opened 2019-04-03
Electron packaged core Webviz as stand-alone app


edufford:
Not sure if this is useful for others, but I thought it might be handy to be able to package the core Webviz ROS bag analyzer app into a stand-alone Electron application.

I made a ["webviz-electron" branch on my fork](https://github.com/edufford/webviz/tree/webviz-electron) with mods to do this, but it has only been tested for Mac and Ubuntu Linux so far.  Electron-builder should be able to package for Windows too, but I'm not sure if it works.

Thanks to the Cruise team for open-sourcing this interesting project.

janpaul123:
Cool stuff! What’s your use case for an electron app exactly?

edufford:
I'm new to Node/React and Electron so mainly I was just curious to learn about how the projects are structured, built, and deployed/packaged.

From previous work in automotive development though, a few potential use cases for offline stand-alone data analysis that I was thinking about are:

1. On-road testing in development vehicles using laptops without online access
2. Lab PCs that might be isolated from the internet for data/source code confidentiality
3. Keeping a local stable build version to use vs the latest deployed online release

It's definitely nice to have a central server-based, continuously updated distribution option for this kind of app, but maybe an offline packaged option could also fill some use case gaps since Electron makes it fairly easy to add on this capability for Node/React projects.

janpaul123:
Ah that makes sense! Since Webviz is just a static page I'd recommend downloading the [gh-pages branch](https://github.com/cruise-automation/webviz/tree/gh-pages) (or going to the website and using the "Save as" feature of most browsers) and then opening webviz-core.html in a browser.

That said, we might revisit using an Electron app in the future, since it would allow us to use some more powerful local APIs. In that case we'd definitely use your contribution, so thanks! Definitely a good learning experience though I bet. :) Closing for now.

edufford:
Ah, I see.  Thanks for the feedback.

Actually, before using Electron, I had tried just loading the webviz-core.html file from the gh-pages branch pack but it fails to load a bag file when dropped onto the page (Chrome, Firefox, and Safari, Mac and Linux), so I figured there was something dynamic that needed a web server to load.

I looked into it some more, and it seems there are a couple issues with just loading a local file version:

1. [By default, Chrome doesn't allow loading web workers from local files](https://stackoverflow.com/a/23206866).  There is a possible workaround by launching Chrome from the command line with the **--allow-file-access-from-files** flag, but then this results in a file path error:
[...]

PR #426, opened 2020-05-11
Add option to build just Webviz

janpaul123:
Plus a Dockerfile that just installs Webviz.

There have been multiple requests for something like this, especially
from people who want to use Webviz in situations with limited internet
connectivity. See e.g. #120, #267, and #406.

I’ve also renamed the existing Dockerfile.

I’ll follow up with another PR that updates the README to explain how to
use the new Docker image once I’ve figured out how to set up automated
Docker builds.

Test plan: tested this locally. There’s some risk of this getting out of
sync since we don’t run this in CI, but hopefully with the automated
Docker build which I’ll set up later we’ll at least get some coverage on
this flow.

### DIFF
--- a/Dockerfile-static-webviz
+++ b/Dockerfile-static-webviz
@@ -1,0 +1,36 @@
+# Copyright (c) 2020-present, GM Cruise LLC
+#
+# This source code is licensed under the Apache License, Version 2.0,
+# found in the LICENSE file in the root directory of this source tree.
+# You may not use this file except in compliance with the License.
+
+# This is a static build of just the Webviz application.
+# This container is published at https://hub.docker.com/r/cruise/webviz.
+
+FROM node:10.16-slim
+
+# Install some general dependencies for stuff below and for CircleCI;
+# https://circleci.com/docs/2.0/custom-images/#required-tools-for-primary-containers
+RUN apt-get update && apt-get install -yq gnupg libgconf-2-4 wget git ssh --no-install-recommends
+
+# Install dumb-init which helps with proper signal handling.
+RUN apt-get install -y dumb-init
+ENTRYPOINT ["/usr/bin/dumb-init", "--"]
+
+# Copy in the files
+COPY . .
+
+# Delete .gitignore files, so we don't accidentally pollute our image.
+RUN git clean -Xf
+
+# Bumped up from the default old_space size (512mb) as it was being exceeded during builds.
+ENV NODE_OPTIONS="--max_old_space_size=4096"
+
+# Build the static webviz.
+RUN npm run install-ci
+RUN npm run build
+RUN npm run build-static-webviz
+
+# Start the server.
+EXPOSE 8080
+CMD ["npm", "run", "serve-static-webviz"]
--- a/README.md
+++ b/README.md
@@ -1,8 +1,6 @@
 # [Webviz](https://webviz.io/) [![CircleCI](https://circleci.com/gh/cruise-automation/webviz.svg?style=svg)](https://circleci.com/gh/cruise-automation/webviz)
 
-**Drag and drop your own bag files into [Webviz](https://webviz.io/app/) to explore your robotics data, or connect to a live robot or simulation using the [rosbridge_server](http://wiki.ros.org/rosbridge_suite/Tutorials/RunningRosbridge).**
-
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -131,9 +131,10 @@
...
     "docs": "DEV_SERVER=true webpack-dev-server",
-    "docs-deploy": "cp -r docs/public __temp_deploy__ && cp --remove-destination packages/webviz-core/public/index.html __temp_deploy__/app/index.html && echo 'webviz.io' > __temp_deploy__/CNAME && gh-pages -d __temp_deploy__ && rm -rf __temp_deploy__",
+    "docs-deploy": "cp -r docs/public __temp_deploy__ && rm __temp_deploy__/app/index.html && cp packages/webviz-core/public/index.html __temp_deploy__/app/index.html && echo 'webviz.io' > __temp_deploy__/CNAME && gh-pages -d __temp_deploy__ && rm -rf __temp_deploy__",
     "publish": "lerna run clean && lerna run build && lerna publish",
-    "serve-dist": "cd docs/public && npx http-server .",
+    "build-static-webviz": "rm -rf __static_webviz__ && cp -r docs/public/app __static_webviz__ && rm __static_webviz__/index.html && cp packages/webviz-core/public/index.html __static_webviz__/index.html && sed -i -- 's/\\/dist\\/webvizCoreBundle.js/webvizCoreBundle.js/' __static_webviz__/index.html && NODE_ENV=production STATIC_WEBVIZ=true webpack",
+    "serve-static-webviz": "node_modules/http-server/bin/http-server __static_webviz__",
```
troygibb, line 137: 👌 We could even do this with the python one-liner we've reccomended before, e.g. `python -m SimpleHTTPServer 8000`.

README.md
```diff
@@ -15,6 +13,19 @@
...
+
+We recommend using the [hosted version of Webviz](https://webviz.io/app/?demo), which can connect to your [rosbridge_server](http://wiki.ros.org/rosbridge_suite/Tutorials/RunningRosbridge) or stream in bag data from your S3/GCS bucket. This way you'll always use the latest version of Webviz.
+
+However, sometimes the hosted version is inconvenient, when streaming data from robots on the field where there is poor internet connectivity. For this it is useful to use a static build of webviz.
+
+```sh
+npm run bootstrap # install dependencies
+npm run build # build all packages
+npm run build-static-webviz # generate static build in __static_webviz__
+npm run serve-static-webviz # serve static build on localhost:8080
```
troygibb, line 26: could we just reduce this to one command that chains them all together? Especially since so many folks are coming from a non-nodejs background, I feel like it would be a lot less distracting to make this a one-command app.

Dockerfile-static-webviz
```diff
@@ -0,0 +1,36 @@
...
+ENTRYPOINT ["/usr/bin/dumb-init", "--"]
+
+# Copy in the files
+COPY . .
+
+# Delete .gitignore files, so we don't accidentally pollute our image.
+RUN git clean -Xf
+
+# Bumped up from the default old_space size (512mb) as it was being exceeded during builds.
+ENV NODE_OPTIONS="--max_old_space_size=4096"
```
troygibb, line 27: could this variable be reset below (line 33) so that the actual server instance doesn't run with the expanded size?

package.json
vidaaudrey: Python  server may not work for other OS

Dockerfile-static-webviz
janpaul123: I believe that this is just a maximum, and it won't allocate all this space if not necessary (by default it's set to 2GB). I confirmed by running the docker image locally and seeing that it doesn't increase memory usage significantly.

README.md
janpaul123: Yeah, I think in effect that's what the Docker image is going to be. Once I figure out automated builds, I'll add a section here that just has one command for running the Docker image, and then these `npm` commands are the more manual version if you want to get in the weeds of it yourself (in which case it's useful to know what the different underlying commands do).

package.json
janpaul123: Yeah I figured it's nice to avoid other languages in our dependencies where we can. Definitely a great trick. Notice also the line that this replaces is a nice one we can recommend to people who are more familiar with JS: `npx http-server .` 😄 I just removed that here because it installs/updates http-server every time you run it.

PR #429, opened 2020-05-12
Fix build memory usage

janpaul123:
Because even though we bumped the limit in the Dockerfile, the Docker
Hub still only gives us an instance with 2GB RAM (see
https://success.docker.com/article/what-are-the-current-resource-limits-placed-on-automated-builds).

The latest version of the Terser plugin has reduced memory usage. I also updated Webpack itself for good measure!

Test plan: Couldn’t actually test if this fixes the automated build though, because
I ran into this issue: https://forums.docker.com/t/422-unprocessable-entity-when-configuring-the-automated-build-in-docker-hub/70527/2
when trying to add a different branch for autobuilding. :( But it works
for me locally without the limit increase in the Dockerfile, so that’s
encouraging.


package.json
```diff
@@ -105,12 +105,12 @@
...
     "style-loader": "^0.23.1",
     "styled-components": "4.4.1",
-    "terser-webpack-plugin": "^1.1.0",
+    "terser-webpack-plugin": "3.0.1",
     "text-encoding": "^0.7.0",
     "unist-util-visit": "^1.4.0",
     "url-loader": "^1.1.2",
     "url-search-params": "^1.1.0",
-    "webpack-cli": "^3.3.3",
+    "webpack-cli": "3.3.11",
```
vidaaudrey, line 113: Is it intentional to lock the webpack-cli versions? By not locking it we probably can catch more issues related to dependency versions.

janpaul123: In our internal version we lock all versions so I figured I'd do it here too. Shouldn't matter too much though because we also have package-lock.json? How would we catch more issues related to dependency versions?

PR #435, opened 2020-05-13
Attempt to fix "invalid height" regl errors

davidswinegar:
This error occurs when a texture is beyond the max texture size. It
only shows up on certain linux machines right now, and started in
the past month. It's possible that this error is because of some
error in the messages we're seeing, but a more likely reason is the
launch of the gltext feature.

We are creating an atlas in GLText with a limited width, which means
that the height direction grows much larger than the width. It's
possible (not reproduced) that this causes the height to exceed
GL_MAX_TEXTURE_SIZE on some machines. Instead set the width of the
atlas to the actual GL_MAX_TEXTURE_SIZE, so that we give our
characters more horizontal space and hope that we don't exceed the
height limits.

Test plan: hope

### DIFF
--- a/packages/regl-worldview/src/commands/GLText.js
+++ b/packages/regl-worldview/src/commands/GLText.js
@@ -57,7 +57,6 @@ type CharacterLocations = {
 // Font size used in rendering the atlas. This is independent of the `scale` of the rendered text.
 const MIN_RESOLUTION = 40;
 const DEFAULT_RESOLUTION = 160;
-const MAX_ATLAS_WIDTH = 512;
 const SDF_RADIUS = 8;
 const CUTOFF = 0.25;
 const BUFFER = 10;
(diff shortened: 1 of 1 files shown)


janpaul123:
Seems good to me but I'd like @hhsaez' input. 👍

davidswinegar:
@troygibb hernan sent me: http://webglstats.com/webgl/parameter/MAX_TEXTURE_SIZE which says that everyone has at least min 2048 texture size, and almost everyone has at least 4096. I tested with our layout and we get a ~3900 height for the texture normally, but it's possible someone either has a 2048 limit on their computer (esp if they're accessing it in some VM maybe?) or that somehow a few extra characters are added to our atlas and we get above a 4096 size. Switching the width to 2048 drops the height to 900 so I feel pretty good about that solution.

troygibb:
Whoa sweet graph! I still think it'd be worth following up with the person who's experiencing most of the errors though. 

Seems like a safe enough change in any case.

PR #438, opened 2020-05-19
Avoid duplicating position buffer during line rendering

hhsaez:
## Summary

We're currently duplicating position data when rendering lines in order to fix an issue on some specific hardware (see #24). 

In this PR I'm implementing an alternative approach, which is to use 4-component vectors for positions. Based on previous experience, we might be causing alignment issues when using `vec3` and different stride/offset values for positions and that could be the reason behind the original problem. 

## Test plan

* All existing tests and stores must pass
* Added new stories for line strip rendering
* I need help in order to test this implementation since I don't have access to the specific hardware. Even when this implementation doesn't duplicate data, this solution won't be good enough if the original problem persists. See #24 for more details and a simple way to reproduce the problem.

## Versioning impact
* Patch

### DIFF
--- a/packages/regl-worldview/src/commands/Lines.js
+++ b/packages/regl-worldview/src/commands/Lines.js
@@ -68,7 +68,8 @@ edge cases.)
 */
 
 const FLOAT_BYTES = Float32Array.BYTES_PER_ELEMENT;
-const POINT_BYTES = 3 * FLOAT_BYTES;
+const POINT_COMPONENTS = 4;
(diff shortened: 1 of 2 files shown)


jtbandes:
If it works, looks great!

hassoncs:
@hhsaez, I'm working on cleaning up old branches/PRs. Mind I close this?

hhsaez:
@hassoncs  Let's close it.

PR #453, opened 2020-06-03
Storybook v5 and Storycap

hassoncs:
Upgrade Storybook from v4 to v5
Upgrade our screenshot setup from `storybook-chrome-screenshot` to `storycap`

Most of the changes are just renaming. The main changes are in the `config.js` for storybook and `package.json` that starts up `storycap`.

🔥🔥🔥 This speeds up the build by x3! From ~1h build times to 20m!
image


packages/webviz-core/src/panels/Plot/index.stories.js
```diff
@@ -151,7 +151,6 @@ const fixture = {
   activeData: {
     startTime: { sec: 0, nsec: 202050 },
     endTime: { sec: 24, nsec: 999997069 },
-    currentTime: { sec: 0, nsec: 750000000 },
```
janpaul123, line 154: Why this change?

stories/config.js
```diff
@@ -1,34 +1,44 @@
...
-addDecorator(initScreenshot());
-setScreenshotOptions({
-  // if this function is present on the page, we wait for the promise it returns to resolve before taking a screenshot.
-  waitFor: "waitFor",
-  delay: 100, // Small delay for rerenders that some components do.
-  viewport: {
-    width: 1001,
-    height: 745,
+addDecorator((storyFn) => {
+  // TODO: [EP-26051] This can be removed when modals remove themselves
```
janpaul123, line 28: We usually avoid internal JIRA tags in the open source repo.

hassoncs: Ah, good catch. I'll remove it here but leave it in the closed source repo 👍

packages/webviz-core/src/panels/Plot/index.stories.js
hassoncs: I think this accidentally slipped in, will revert

hassoncs:
@vidaaudrey, sadly there's no naming pattern I could use that gets the before/after image paths to match :( Storycap has a `--flat` flag, but it still returns different paths

PR #454, opened 2020-06-04
Updates to various package-lock.json

rk-yen:
## Summary

When doing a `npm run bootstrap` the files added to the PR got changed. To keep the repository up-to-date, we should commit the updates to the lock files.

## Test plan

After committing the change, `npm run bootstrap` was run again. And, it was verified that no files were in the `modified` state.

## Versioning impact

This change has no impact on versioning.


janpaul123:
Sweet! Currently the way our CI is set up it always fails for forks so we don't leak our secrets, so I pulled in your branch here: #455. I'll merge it shortly when it's green. Thank you!

rk-yen:
> Sweet! Currently the way our CI is set up it always fails for forks so we don't leak our secrets, so I pulled in your branch here: #455. I'll merge it shortly when it's green. Thank you!

Thanks for taking the effort.
For future PRs, would you recommend me creating a PR from a branch instead of a 'fork' ?

janpaul123:
I don't think that's currently possible with our access controls so a fork works just fine! 👍

PR #464, opened 2020-06-10
Remove react-key-listener

troygibb:
## Summary

Some basic cleanup since we moved the key listener to just be a regular component [here](https://github.com/cruise-automation/webviz/blob/remove-react-key-listener/packages/webviz-core/src/components/KeyListener.js#L1).

## Test plan

Pass CI.

## Versioning impact

None! Unless we want to remove it from npm @janpaul123


jtbandes:
Does that mean new projects shouldn't depend on this anymore? Why?

janpaul123:
It's so small you can just write it yourself or copy/paste it.

PR #477, opened 2020-07-21
Allow custom depth and blend states in lines

hhsaez:
## Summary

This PR allows the `<Lines/>` command to use custom `depth` and `blend` settings for advanced rendering techniques. This way, we can control not only opacity for lines, but also how they should be blended with other objects in the scene. For example, we can split lines into different groups, rendering each of them with different colors.

Memoization is required when rendering lines since we want to avoid creating multiple `render` functions with the same settings.

## Test plan

* Existing tests must pass
* Added a story that uses a custom blend setting 

## Versioning impact

`Patch`: This is a very small change


packages/regl-worldview/src/commands/Lines.js
```diff
@@ -456,11 +464,12 @@ const lines = (regl: any) => {
   }
 
   // Disable depth for debug rendering (so lines stay visible)
-  const render = (debug, commands) => {
+  const render = (props, commands) => {
+    const { debug, depth = defaultReglDepth, blend = defaultReglBlend } = props;
```
davidswinegar, line 468: can you add these props to flow types as well?

hhsaez:
@davidswinegar I added flow types for `BlendState` and `DepthState` in my latest commit, but I'm only using them in the `render` function because I didn't want to make major changes in the API. We should start using these new types now whenever we do other changes.

@hassoncs I included both states in the docs, as optional params. I don't think we should add specific code examples for them, since I can't think of any simple and meaningful example, since these are used in more complex situations. What do you think?

packages/regl-worldview/src/commands/Lines.js
```diff
@@ -463,14 +464,23 @@ const lines = (regl: any) => {
...
+  // Create a new render function based on rendering settings
+  // Memoization is required in order to prevent creating too many functions
+  // that use the same arguments, potentially leading to memory leaks.
+  const memoizedRender = memoize(
+    (props: { debug?: boolean, depth?: DepthState, blend?: BlendState }) => {
+      const { debug, depth = defaultReglDepth, blend = defaultReglBlend } = props;
+      if (debug) {
+        return regl({ depth: { enable: false } });
+      }
+      return regl({ depth, blend });
```
hassoncs, line 476: We could make this `memoizedRender ` a bit simpler by moving the `debug` flag check out of this function and into `renderLine`. Then we could just memoize on `depth` and `blend` here:

```
const memoizedRender = memoize(
    ({ depth = defaultReglDepth, blend = defaultReglBlend }: { depth?: DepthState, blend?: BlendState }) => regl({ depth, blend }),
    (...args) => JSON.stringify(args)
  );
```

hhsaez: Wouldn't that end up creating more `render` functions for debug mode only?

PR #479, opened 2020-07-22

janpaul123:
Changelog:
- Added a warning when using the "custom" axis setting in the plot panel when the x and y axes do not have the same number of elements.
- Clicking in experimental preloaded plots seeks the playback cursor.
- Added support to customize namespace override colors.
- Fixed 3D panel's lidar point color selector spilling out of container.
- Messages in Node Playground won't have a `datatype` field on them any more, due to internal data structure changes. If you depend on that field, please change your nodes to use `topic` instead.
- Fixed Node Playground bug that prevented playing nodes from imported layouts.
- Fixed bug where selecting an instanced object would remove linked global variables.
- Fixed bug in the 3D panel where the selected object details menu would not show the topic name or global variable link buttons for instanced objects.
- Improved Node Playground error handling for nodes that have badly formatted filenames.
- Added support for plotting values nested inside JSON fields (e.g. `/logs.data.some_field`), either by typing the path into the Plot panel or by clicking the Raw Messages panel's chart icons.
- Enabled preloaded plots to follow playback after a zoom or pan.
- Enforced scrolling in charts to zoom in on the x-axis only, by default.
- Added visible topic count at group level in the 3D panel.
- Fix bug in Node Playground resulting in the error "cannot find name 'json'".
- Fixed messages not loading when subscribing to a new topic when sorting messages by header stamp.
- Added a field to the "Diagnostics - Detail" panel's config to persist collapsed sections.
- Added support to resize the 3D panel topic tree horizontally by dragging the bottom right corner.
- Preloaded plots are now enabled by default.
- Long strings are truncated by default in the RawMessages panel.
- Fixed parsing of arrays in messages' `json` fields.
- Fixed index-based plots to show data for just the current time tick.
- Added support for updating the `seek-to` query parameter whenever playback is paused.
- Added ability to toggle sorting by diagnostic level in the Diagnostics panel.
- Enabled displaying all diagnostics, grouped by name, for a given hardware ID in the "Diagnostics - Detail" panel.
- Fixed bug in Diagnostics - Detail panel where incoming messages were not being collected correctly.
- Notify the user when a Webviz node throws an error.
- Updated seeking in the playback bar to trigger a change in the URL's `seek-to` query param.
- Fixed corners in grouped line strips.


janpaul123:
I manually went through the screenshots. This one looked suspicious:

Screen Shot 2020-07-22 at 11 07 16 AM

But it's actually correct. Previously we would always show the same thing as the little endian screenshot there, which was incorrect, and @vidaaudrey fixed it to interpret the data as big endian for this screenshot. 👍

PR #500, opened 2020-08-31
Disable patching in open source for internal CI

vidaaudrey:
Supporting layout URL patching in open source version causes complexity in managing internal CI right now.  Will add it back later.


packages/webviz-core/src/loadWebviz.js
```diff
@@ -178,7 +178,6 @@ const defaultHooks = {
     const { REMOTE_BAG_URL_2_QUERY_KEY } = require("webviz-core/src/util/globalConstants");
     return [REMOTE_BAG_URL_2_QUERY_KEY];
   },
-  maybeUpdateURLToTrackLayout: () => {},
```
troygibb, line 181: maybe just have this return false? It's a bit weird to use this value below knowing full well it doesn't exist.

vidaaudrey:
Closing in favor of #501

PR #511, opened 2020-09-12
Replace react-container-dimensions with custom code in Dimensions

davidswinegar:
Similar to what is coming to the rest of Webviz, but use this in the
Worldview. Replace the inefficient react-container-dimensions with
custom Dimensions.

Test plan: screenshot tests.


vidaaudrey (approved):
🎉

PR #512, opened 2020-09-14

vidaaudrey:
Changelog:
- Removed antd dependency to reduce bundle size.


esthersweon (approved):
Would be nice to be more specific about layout issue in the PR description (i.e. "panel splitting + removing"?). Thank you for opening this PR!

vidaaudrey:
> Would be nice to be more specific about layout issue in the PR description (i.e. "panel splitting + removing"?). Thanks for opening this PR!

That was fixed in #510. Just realized the layout issue for internal, so I removed it from changelog :)

Issue #491, opened 2020-08-04
Marker Array Points Not Scaled Correctly


jasonimercer:
Points in a Marker Array message are scaled in units of pixels rather than meters as the unit in other elements.

issue:

https://github.com/cruise-automation/webviz/blob/master/packages/regl-worldview/src/commands/Points.js#L32

vs

code that considers the camera:

https://github.com/cruise-automation/webviz/blob/master/packages/regl-worldview/src/commands/Lines.js#L147-L151

rviz and webviz visualizing points in a marker array:
![image (23)](https://user-images.githubusercontent.com/12039018/89295041-633cf380-d62e-11ea-91d0-8312688ddb38.png)

example input with points in a marker array 
[2020-07-28-12-26-37.bag.gz](https://github.com/cruise-automation/webviz/files/5022124/2020-07-28-12-26-37.bag.gz)

troygibb:
Thanks @jasonimercer for reporting! We may have to refactor our `<Points />` command to account for this issue. 

Can you describe your use case a bit more? Why do you need to express points so exactly? Why not use some of our other primitive shapes?

cc @hhsaez

jasonimercer:
We use Marker Arrays for rich diagnostics. Our code publishes lines, polygons, arrows, text, points to give our triage teams an easier view into the state of the robot and the surroundings.

In this marker array we denote remembered obstacles that were point-like as well as obstacles that were identified as moving polygons. We also show demarcations between regions we can observe and will add data to memory, regions we only use to invalidate memory and regions where we rely on memory.

When we are displaying point-like lidar data in our aggregated memory, the points are single pixels using Webviz (which on modern displays and a busy interface, is as good as invisible - see pixel data in screenshot above). We don't need them to be 100% accurate, we just need them to be visible similar to RViz so we can continue to shift the hold-outs towards webviz.

jasonimercer:
Callout added to image above demonstrating issue:

![89295041-633cf380-d62e-11ea-91d0-8312688ddb38](https://user-images.githubusercontent.com/12039018/89418377-f85aed80-d6fd-11ea-9bb0-434a4c9091b3.png)

troygibb:
Thanks Jason for the context -- I just wrote a ticket to track this issue internally.

jasonimercer:
I see code has been merged related to this case. Is it live on the demo site or is there something that needs to be done to trigger it? When I load my example bag file above into the demo site the display appears the same.

[...]

PR #513, opened 2020-09-15
Use world units for point sizes

hhsaez:
By default, <Points /> uses the `scale` property to determine the size of all points. `gl_PointSize` expects values in pixels rather than in units in world space, though.

The changes in this commit introduce a new optional flag to use world space point sizes instead of pixels. In order to accomplish that goal, the vertex shader computes a scaled point size based on the current view transformation and render target size.

Test plan: Added new stories for `<Points />` command, using both screen and world space point sizes with unit cubes for reference.

## Versioning impact

Minor: the new flag is used to ensure backward compatibility

### DIFF
--- a/docs/src/4.8.Points.mdx
+++ b/docs/src/4.8.Points.mdx
@@ -11,6 +11,10 @@ import { Points } from './jsx/allLiveEditors'
 | `children`          | `Point[]` | `[]`           | array of Point markers to render      |
 | `useWorldSpaceSize` | `boolean` | `undefined`    | use world space units for point sizes |
 
+### Using World Space sizes
+
+By default, points are rendered in `view space`, meaninig that their size (in pixels) is constant and not affected by the point's position in the world. In contrast, when the `useWorldSpaceSize` flag is set to `true`, points will be rendered in `world space`, meaning that their size (indicated in world units, i.e. meters) will be calculated based on their distance to the camera.
+
 ### Point
 
 ```js
--- a/docs/src/4.8.Points.mdx
+++ b/docs/src/4.8.Points.mdx
@@ -6,9 +6,10 @@ import { Points } from './jsx/allLiveEditors'
 
 ## Props
 
-| Name       | Type      | Default | Description                      |
-| ---------- | --------- | ------- | -------------------------------- |
-| `children` | `Point[]` | `[]`    | array of Point markers to render |
+| Name                | Type      | Default        | Description                           |
+| ------------------- | --------- | -------------- | ------------------------------------- |
+| `children`          | `Point[]` | `[]`           | array of Point markers to render      |
+| `useWorldSpaceSize` | `boolean` | `undefined`    | use world space units for point sizes |
 
 ### Point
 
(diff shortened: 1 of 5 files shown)


packages/regl-worldview/src/commands/Points.js
```diff
@@ -6,64 +6,100 @@
...
+        // Then, we compute the distance between both points in clip space, dividing
+        // by the w-component to account for distance in perspective projection.
+        float d = length(up.xyz / up.w - gl_Position.xyz / gl_Position.w);
+
+        // Finally, the point size is calculated using the size of the render target
+        // and it's aspect ratio. We multiply it by 0.5 since distance in clip space
+        // is in range [0, 2] (because clip space's range is [-1, 1]) and
+        // we need it to be [0, 1].
+        float invAspect = viewportHeight / viewportWidth;
+        gl_PointSize = pointSize * 0.5 * d * viewportWidth * invAspect;
```
troygibb, line 63: great comments!

docs/src/4.8.Points.mdx
```diff
@@ -6,9 +6,10 @@ import { Points } from './jsx/allLiveEditors'
 
 ## Props
 
-| Name       | Type      | Default | Description                      |
-| ---------- | --------- | ------- | -------------------------------- |
-| `children` | `Point[]` | `[]`    | array of Point markers to render |
+| Name                | Type      | Default        | Description                           |
+| ------------------- | --------- | -------------- | ------------------------------------- |
+| `children`          | `Point[]` | `[]`           | array of Point markers to render      |
+| `useWorldSpaceSize` | `boolean` | `undefined`    | use world space units for point sizes |
```
troygibb, line 12: might be worth adding a bit how this relates to `scale`

troygibb (approved):
Nice work here. 

I guess the part that's left is figuring out how to integrate this with the rest of Webviz, any ideas here? I.e. Jason's [issue here](https://github.com/cruise-automation/webviz/issues/491) is rendering points from `MarkerArray`.

hhsaez:
> I guess the part that's left is figuring out how to integrate this with the rest of Webviz, any ideas here? I.e. Jason's [issue here](https://github.com/cruise-automation/webviz/issues/491) is rendering points from `MarkerArray`.

The optional `useWorldSpaceSize ` flag allows Worldview to remain generic, supporting both render modes (screen and world space), since it doesn't really know about markers. In Webviz, we need to use that flag so point markers will be always rendered in world space. This is a similar to how we handle screen invariance in `<GLText />`, which is always enabled in Webviz.

Does that make sense?

troygibb:
> The optional useWorldSpaceSize flag allows Worldview to remain generic, supporting both render modes (screen and world space), since it doesn't really know about markers. In Webviz, we need to use that flag so point markers will be always rendered in world space. This is a similar to how we handle screen invariance in <GLText />, which is always enabled in Webviz.
> Does that make sense?

Yea totally! I guess my real question here is whether adding this flag on by default in Webviz will be backwards compatible with many of the points we already display. Lidar points seem easy, since we control how they're rendered. However, for example, I don't have a good sense how much folks rely on `MarkerArray` points in production.

PR #514, opened 2020-09-15
Bump node version

davidswinegar:
I am installing a dependency that needs a newer version of node -
lets bump it.


hassoncs (approved):
As long as we're sure everything builds in both repos, I don't see why not. Can we specify this node version in our `package.json` too? Maybe following this answer: https://stackoverflow.com/questions/29349684/how-can-i-specify-the-required-node-js-version-in-package-json

Using the wrong version of node has been a source of user issues in the past, so anything we can do to automatically help would be nice

davidswinegar:
That's a good point - I'll put together a follow up PR once this is in the main repo.

PR #524, opened 2020-09-30
Do not require a scene in GLTF

MatthewSteel:
The `scene` property is optional. If it's not specified, use "the first"
scene.

In a .glb file I opened that didn't have a specified scene, there was
one key in the `scenes` object: `"0"`. In GLTF 2 `scenes` is an array, so
we might be able to get away with just `scenes[0]`, but this seems
slightly more robust.

I don't think we need to be super careful about files without scenes.
Maybe we want some controls for the scene to draw to be specified in the
command if there could be more than one, but let's not worry about that
until we come across it.

Tested by loading a file up in the worldview docs site locally. I don't have
a publicly available .glb example.

### DIFF
--- a/packages/regl-worldview/src/commands/GLTFScene.js
+++ b/packages/regl-worldview/src/commands/GLTFScene.js
@@ -38,6 +38,14 @@ function glConstantToRegl(value: ?number): ?string {
   throw new Error(`unhandled constant value ${JSON.stringify(value)}`);
 }
 
+const getSceneToDraw = ({ json }) => {
+  if (json.scene != null) {
+    return json.scene;
+  }
+  // Draw the first scene if the scene key is missing.
+  return Object.keys(json.scenes)[0];
+};
+
 const drawModel = (regl) => {
   const command = regl({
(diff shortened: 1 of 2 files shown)


packages/regl-worldview/src/commands/GLTFScene.js
```diff
@@ -38,6 +38,14 @@ function glConstantToRegl(value: ?number): ?string {
   throw new Error(`unhandled constant value ${JSON.stringify(value)}`);
 }
 
+const getSceneToDraw = ({ json }) => {
```
davidswinegar, line 41: should this throw an error if we can't find scene or scenes?

davidswinegar (approved):

packages/regl-worldview/src/commands/GLTFScene.js
MatthewSteel: I'll make it do that. Maybe it should just not render anything if there are no scenes, but we can relax it later -- throwing is still "more permissive" than current behaviour.

PR #526, opened 2020-10-19

vidaaudrey:
## Changelog

**Improvements**
- Performance for 3D panel-heavy layouts has been improved.
- Added utf-8 support for strings when the `useBinaryTranslation` experimental feature is turned on.
- Charts with very large numbers of points display sampled data to improve performance.
- Added support for newline characters in the Raw Messages panel.
- Node Playground will now automatically format your code on save. This behavior can be toggled using the panel settings menu.
- Added a switch in the Topic Tree that enables a `Diff Mode` visualization when two sources are loaded in the 3D Panel.
- Added an `other source` option in the Raw Messages panel's diff mode to compare the same topic across different sources.
- Added support for Global Variables in Node Playground scripts.

**Fixes**
 - Fixed websocket crash when the `useBinaryTranslation` feature is enabled.
 - Fixed bug where disabling a topic would cause the Topic Tree to report undefined datatype errors.
 - Fixed size for `<Point />` markers to use meters instead of pixels in 3D Panel.


hassoncs (approved):
🎉 😃

PR #528, opened 2020-10-19
Support changing the color of laserscans (copy)

vidaaudrey:
Copy of #525

### DIFF
--- a/packages/webviz-core/src/panels/ThreeDimensionalViz/LaserScanVert.js
+++ b/packages/webviz-core/src/panels/ThreeDimensionalViz/LaserScanVert.js
@@ -18,11 +18,12 @@ export default `
   uniform float range_min;
   uniform float range_max;
   uniform bool isHitmap;
+  uniform vec4 hitmapColor;
+  uniform vec4 overrideColor;
 
   attribute float index;
   attribute float range;
   attribute float intensity;
-  attribute vec4 hitmapColor;
 
   varying vec4 vColor;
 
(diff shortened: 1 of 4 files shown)


packages/webviz-core/src/panels/ThreeDimensionalViz/LaserScanVert.js
```diff
@@ -38,7 +39,7 @@ export default `
     } else if (isHitmap) {
       vColor = hitmapColor;
     } else {
-      vColor = vec4(0.5, 0.5, 1, 1);
+      vColor = overrideColor;
```
esthersweon, line 42: Do we need to handle case for when `overrideColor` is not defined?

Also, should we just use `color` for topic settings going forward, since we are treating both the same way, and the 2 names have caused issues in the past?

packages/webviz-core/src/panels/ThreeDimensionalViz/commands/LaserScans.js
```diff
@@ -47,6 +50,9 @@ const laserScan = (regl: Regl) =>
       angle_increment: regl.prop("angle_increment"),
       range_min: regl.prop("range_min"),
       range_max: regl.prop("range_max"),
+
+      overrideColor: (context, props) => toRGBA(props.settings?.overrideColor || DEFAULT_FLAT_COLOR),
```
esthersweon, line 54: Ah, never mind my first comment - `overrideColor will always be defined`

esthersweon (approved):

packages/webviz-core/src/panels/ThreeDimensionalViz/LaserScanVert.js
vidaaudrey: I still like `overrideColor ` over `color` since it's more explicit.  Could definitely consolidate into one later though.

PR #529, opened 2020-10-20
Fix marker array scaling with small scale

iwanders:
## Summary

This PR resolves the last issue of #491. The current marker scale calculation is broken for any marker scale less than 1. This was caused by [this line](https://github.com/cruise-automation/webviz/blob/master/packages/regl-worldview/src/commands/Points.js#L89), where the marker scale was being bounded by `regl.limits.pointSizeDims`, which is `[1, 2047]`. No markers could be below one meter in size. Instead this limiting needs to happen after the new calculation that was added in #513. So [here](https://github.com/cruise-automation/webviz/blob/master/packages/regl-worldview/src/commands/Points.js#L67) is where the limiting should be applied. To achieve that I added `uniform` entries to the `Point`, to hold the value for the the point limits, then after we've done the calculation for the point size we can min/max appropriately. This ensures that when zooming out completely points are always shown as at least one pixel (checked this manually).

## Test plan

I first identified and fixed the issue. Then I reverted that change, added two unit tests where the points drawn with `useWorldSpaceSize=true` resulted in different sizes:
before_non_unit_cube_orthographic

before_non_unit_perspective

Then I reapplied my proposed fix (second commit) and confirmed that the sizes were now identical, as one would expect:
after_non_unit_orthographic

after_non_unit_perspective

I tried to grab proper screenshots with `npm run screenshot-local-debug`, but that crashed with an error about an invalid array length.

## Versioning impact

This is a bugfix in the current version of `master`, currently marker arrays with points always show up as a size of 1 meters.

### DIFF
--- a/packages/regl-worldview/src/commands/Points.js
+++ b/packages/regl-worldview/src/commands/Points.js
@@ -21,7 +21,7 @@ type Props = {
 
 const makePointsCommand = (useWorldSpaceSize: boolean) => {
   return (regl: Regl) => {
-    const [min, max] = regl.limits.pointSizeDims;
+    const [minLimitPointSize, maxLimitPointSize] = regl.limits.pointSizeDims;
     return withPose({
       primitive: "points",
       vert: `
@@ -34,6 +34,8 @@ const makePointsCommand = (useWorldSpaceSize: boolean) => {
     uniform bool useWorldSpaceSize;
     uniform float viewportWidth;
     uniform float viewportHeight;
+    uniform float minPointSize;
+    uniform float maxPointSize;
 
     attribute vec3 point;
     attribute vec4 color;
@@ -64,6 +66,9 @@ const makePointsCommand = (useWorldSpaceSize: boolean) => {
       } else {
         gl_PointSize = pointSize;
       }
+
+      // Finally, ensure the calculated point size is within the limits.
+      gl_PointSize = min(maxPointSize, max(minPointSize, gl_PointSize));
     }
     `,
       frag: `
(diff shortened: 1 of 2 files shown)


packages/regl-worldview/src/commands/Points.js
```diff
@@ -34,6 +34,8 @@ const makePointsCommand = (useWorldSpaceSize: boolean) => {
     uniform bool useWorldSpaceSize;
     uniform float viewportWidth;
     uniform float viewportHeight;
+    uniform float minPointSize;
+    uniform float maxPointSize;
```
iwanders, line 38: I personally wasn't a fan of having to pass this through like this. I feel we should have direct access to the `ALIASED_POINT_SIZE_RANGE` parameter from this environment. But some searching didn't result in anything to retrieve `ALIASED_POINT_SIZE_RANGE` without the `getParameter` on a webgl context. So I gave up on that, if anyone knows how to access the limits without having to pass them in explicitly like this let me know.

iwanders:
Pipeline failed on missing credentials again, same message as in my [other PR](https://github.com/cruise-automation/webviz/pull/525#issuecomment-712291683).

vidaaudrey (approved):
LGTM. @hhsaez could you help take a look as well?

packages/regl-worldview/src/commands/Points.js
hhsaez: Unfortunately, that's the only way to do it. We cannot access those values directly from shaders unless we use uniforms. So, you're approach is the correct one.

hhsaez (approved):
This looks great. Thanks for fixing it and adding the tests!

packages/regl-worldview/src/commands/Points.js
iwanders: That surprises me, maybe something to do with the shader being indepedent of the environment it's being run in... Thanks for confirming this approach :+1:

iwanders:
@vidaaudrey , @hhsaez , can one of you duplicate this PR such that the pipeline runs with your permissions to get around the ci permission failure?

iwanders:
Closing this as it has been merged with #530 . Thanks for the help @vidaaudrey  and @hhsaez.

vidaaudrey:
@iwanders  FYI regl-worldview@0.12.2 is published to include your changes here.

PR #533, opened 2020-10-23
Allow custom render states for <Triangles /> and <FilledPolygons />

hhsaez:
## Summary

This PR uses `withRenderStateOverrides` to set custom depth and blend values in both `<Triangles />` and `<FilledPolygons />`. This is required for advanced rendering modes in Webviz, life `Diff Mode`. 

## Test plan

New stories were created for both commands, with and without render state overrides. 

## Versioning impact

Minor. This is a bug fix and it's backward compatible.


vidaaudrey (approved):
LGTM

PR #536, opened 2020-10-27
Worldview: Fix render state memoization to avoid performance issues

hhsaez:
## Summary

The function `withRenderStateOverrides` does not handle memoization correctly, leading to performance issues on Worldview clients. Since `memoizedRender` is created and invoked in the same scope, the memoization does not really happens.

I'm moving the memoization so it is done outside of `renderElement`, which is the correct approach. After this change, performance is back to normal

## Test plan

Cover by existing screenshot tests.

## Versioning impact

Patch.

### DIFF
--- a/packages/regl-worldview/src/utils/withRenderStateOverrides.js
+++ b/packages/regl-worldview/src/utils/withRenderStateOverrides.js
@@ -15,7 +15,7 @@ const withRenderStateOverrides = (command: any) => (regl: any) => {
   // Generate the render command once
   const reglCommand = command(regl);
 
-  // Use memoization to prevent generating too multiple render commands
+  // Use memoization to avoid generating multiple render commands for the same render states
   // for the same render states
   const memoizedRender = memoize(
     (props: { depth: DepthState, blend: BlendState }) => {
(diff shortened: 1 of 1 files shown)


packages/regl-worldview/src/utils/withRenderStateOverrides.js
```diff
@@ -15,21 +15,22 @@ const withRenderStateOverrides = (command: any) => (regl: any) => {
   // Generate the render command once
   const reglCommand = command(regl);
 
+  // Use memoization to prevent generating too multiple render commands
```
hassoncs, line 18:
```suggestion
  // Use memoization to avoid generating multiple render commands for the same render states
```

hassoncs (approved):
Thanks for the quick fix! I tried it out locally and this does the trick 🎉

troygibb:
@hhsaez how can we catch these perf regressions in the future?

hhsaez:
@troygibb That is great question and it's something we've been discussing for a while (and we still are). 

My understanding is that we do have some performance reports running automatically on the internal repository (unfortunately, I don't have access to those results), but the problem with those tests is that they ran **after** the PR has been already merged. On the other hand, it would be great to have some performance test ran for each PR, but that would make the review/merge process much slower since we need to measure several scenarios.

On my end, I knew I had to consider performance issues when implementing these changes and ran the usual tests manually. Yet the data I have is not representative of all scenarios, and I did not notice the performance regression (I had to run two bags at the same time and use dev tools profiling to reproduce the problem myself). I definitely need to find a better dataset for my test.

Any suggestions?

PR #542, opened 2020-11-10
Optimize Dockerfile for smaller container images

### LINKED ISSUE #541: Building the static Docker container fails
When I try to build the static Docker container, it runs `npm run install-ci` which fails:

```
lerna ERR! npm install --global-style exited 1 in 'webviz-core'
lerna ERR! npm install --global-style stderr:
npm WARN tar ENOENT: no such file or directory, open '/packages/webviz-core/node_modules/.staging/hammerjs-9054bb7c/hammer.min.js'
[lots of similar messages]
```

rgov:
## Summary

This change dramatically shrinks the static Webviz Docker image from 1005.44 MB to 74.14 MB, a reduction of 92.63%.

The Dockerfile now uses a multi-stage build to install build-time dependencies into a temporary container, while only the build products are copied into the final container image.

The final container image is based on nginx, a widely-deployed HTTP server and one of the most popular Docker containers. This may remove the need to keep the Node package `http-server` in the project's dependencies, but I'm not aware if it is used elsewhere.

The Dockerfile is now structured to take advantage of the Docker cache when re-building. Code changes that do not alter `package.json` etc. will not require a complete re-installation of dependencies.

The `git clean` stage was replaced with a `.dockerignore` file which prevents the files from being copied into the container in the first place. This file is generated from `.gitignore` and there are instructions provided for re-generating it in the future. This saves a lot of time by not copying `node_modules` and other large files into the Docker build context.

On advice of the official `node` Docker image, I switched from the `slim` variant to the vanilla `node` image. This doesn't have any significant effect. For a smaller final container image, I use the `alpine` variant of nginx.

After this is released, users can run `docker image prune` to reclaim disk space from previous versions of the container image. 

Closes #541 by moving the code from the root directory to the `/app` subdirectory inside the `builder` container.

## Test plan

I built the Docker image on an x86_64 macOS host using Docker Desktop 2.5.0.0 (49427) and briefly visited the built Webviz in Chrome.

I also successfully built it on an arm64 Ubuntu Bionic host.

## Versioning impact

This is not *likely* to affect versioning.

In the unlikely event that someone customized the built `cruise/webviz` image (e.g., used it as a base image for another container), their customizations would probably break. But it is doubtful anyone is customizing the image in this way because the JavaScript has been packed and is virtually immutable.

For backwards compatibility, the nginx configuration is patched to serve on port 8080 by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,6 @@ docker run -p 8080:8080 cruise/webviz
 
 ```sh
 npm run bootstrap # install dependencies
-npm run build # build all packages
 npm run build-static-webviz # generate static build in __static_webviz__
 npm run serve-static-webviz # serve static build on localhost:8080
 ```
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,43 @@
+.dockerignore
+.git
+.gitignore
+Dockerfile*
+
+# The following are defined in .gitignore; see that file for more information.
+#
+# To update this list, append the output of:
+#     sed 's/#.*//g' .gitignore | sed '/^$/d'
+
+logs
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+pids
+*.pid
+*.seed
+*.pid.lock
+lib-cov
+coverage
+.nyc_output
+.grunt
+bower_components
+.lock-wscript
+build/Release
+node_modules/
+jspm_packages/
+typings/
+.npm
+.eslintcache
+.node_repl_history
+*.tgz
+.yarn-integrity
+.env
+.next
+.DS_Store
+.cache/
+__screenshots__/
+dist/
+lib/
+__static_webviz__/
+.reg/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+# Note: When updating this file, please also update .dockerignore
+
 # Logs
 logs
 *.log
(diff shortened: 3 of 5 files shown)


rgov:
As a further improvement I would recommend changing the process from

    npm run build
    npm run build-static-webviz

to just `build-static-webviz`, and change the script to invoke `lerna run build`. This way you avoid running `webpack` twice, which is *very* slow.

README.md
```diff
@@ -40,7 +40,6 @@ docker run -p 8080:8080 cruise/webviz
 
 ```sh
 npm run bootstrap # install dependencies
-npm run build # build all packages
```
vidaaudrey, line 43: I'd leave this because sometimes we'll need to build individual packages for publishing.

vidaaudrey (approved):
LGTM and tested locally. Thanks for the contribution. @MatthewSteel, maybe you could take a look as well?

README.md
rgov: It is specifically in the section "Running the static Webviz application"... If it is needed for other situations, I think those should have separate section of the README.

vidaaudrey: Make sense. I just updated branch. Feel free to merge after the CI passes.

rgov: Thanks @vidaaudrey. Can you please merge? (I don't have permission.)

vidaaudrey:
@rgov  due to some CI issue, the build is not successful even though I tried a few times. We'll fix it in the next couple of weeks. I made a copy of your change and merged in #551.  Will close the PR for now. Thanks for your contribution!

PR #546, opened 2020-11-10
Less buffer copying in <Lines>

MatthewSteel:
Whenever we render a line we copy over the whole buffer, which is as
long as the longest line we've ever seen. This can be very inefficient
if there are a small number of very long lines and a large number of
short lines.

Test plan: No functional changes. Screenshots should stay unchanged.

Bumped minor version.

### DIFF
--- a/packages/regl-worldview/package.json
+++ b/packages/regl-worldview/package.json
@@ -1,6 +1,6 @@
 {
   "name": "regl-worldview",
-  "version": "0.13.3",
+  "version": "0.13.4",
   "description": "A reusable component for rendering 2D and 3D views using regl",
   "license": "Apache-2.0",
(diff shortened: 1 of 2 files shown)


hhsaez (approved):
LGTM!

I believe there was a bug on some specific GPUs that required us to duplicate these buffers in order to render them correctly. But I think you're right, even if we need to send the buffers twice, they can still share the same data, avoiding the extra copy. Nice!

MatthewSteel:
> a bug on some specific GPUs that required us to duplicate these buffers in order to render them correctly

This change doesn't affect how many copies we make. It just means that we don't copy the second half of the buffer in cases when we've only filled up the first half.

PR #549, opened 2020-11-12
Allow adding context attributes in Worldview

davidswinegar:
Allow passing in context attributes in Worldview that will be used
when creating the context from `getContext`.

### DIFF
--- a/packages/regl-worldview/src/Worldview.js
+++ b/packages/regl-worldview/src/Worldview.js
@@ -64,7 +64,7 @@ export type BaseProps = {|
   ...Dimensions,
 
   // Context attributes passed into canvas.getContext.
-  contextAttributes?: ?any,
+  contextAttributes?: ?{ [string]: any },
 |};
 
 type State = {|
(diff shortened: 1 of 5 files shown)


packages/regl-worldview/src/Worldview.js
```diff
@@ -149,6 +152,7 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
         // DEFAULT_CAMERA_STATE is applied if both `cameraState` and `defaultCameraState` are not present
         cameraState: props.cameraState || props.defaultCameraState || DEFAULT_CAMERA_STATE,
         onCameraStateChange: props.onCameraStateChange || undefined,
+        contextAttributes: props.contextAttributes || undefined,
```
esthersweon, line 155: #nit - Is `|| undefined` necessary on lines 154-155?

packages/regl-worldview/src/Worldview.js
```diff
@@ -62,6 +62,9 @@ export type BaseProps = {|
   // Used to scale the canvas resolution and provide a higher image quality
   resolutionScale?: number,
   ...Dimensions,
+
+  // Context attributes passed into canvas.getContext.
+  contextAttributes?: ?any,
```
esthersweon, line 67: Can we make slightly more specific with `?{[string]: any}`?

esthersweon (approved):
Just #nit comments / questions; LGTM.

packages/regl-worldview/src/Worldview.js
esthersweon: Actually, would `|| {}` make more sense for line 155?

Issue #298, opened 2019-12-26
Memory leak when use <GLTFScene>


baoyule-js:
i use an array to store <GLTFScene>,and refresh the array 10 times per second ,and i find the memory usage exploded.i do not know why.

janpaul123:
It would be helpful if you could create a simple webpage that demonstrates the issue; would that be possible?

PatrickQianKun:
same issue，use multiple GLTFScene with glb models will cause regl context lost

haojiaping:
Hi,Can you speak Chinese?   I found a serious one in your project.

vidaaudrey:
Thanks for reporting.  @haojiaping  pls feel free to set up a Google hangout to review/repro the issues together. audrey.li@getcruise.com. Or we could communicate via webviz slack https://join.slack.com/t/webvizchat/shared_invite/zt-hyo9sacz-4jAA5ADjaTaolUBwesMMZQ.

haojiaping:
I have sent an email to you. This memory leak problem occurs every time you open a project. After about 10 minutes, the browser's memory will be full and cause a crash.

hhsaez:
@baoyule-js @haojiaping @vidaaudrey I've created [this branch](https://github.com/cruise-automation/webviz/pull/554) in order to attempt to reproduce this problem, without luck so far. 

The story I created (executed by using `npm run storybook`) creates a scene with multiple `<GLTFScene />` elements and then starts updating them multiple times. I let it run for about an hour and in my case, memory usage is constant (somewhere in between 220MB and 250MB).  

The question I have is whether or not this story is representative of the scene you're trying to execute. Feel free to modify that code as you see fit in order to reproduce the problem.

PR #554, opened 2020-11-23
Add <GLTFScene /> dynamic story

hhsaez:
Add a story using multiple `<GLTFScene />` to attempt to reproduce memory issues found in #298 

This new story loads a GLTF model and updates the scene multiple times. Memory usage seems to be consistent (ranging in between 220MB and 250MB in my case) and doesn't grow up after about an hour of execution.


hassoncs:
Hey @hhsaez, I'm cleaning up old branches – do you still want to get this merged or should we close it?

hhsaez:
@hassoncs I think it's ok to close it. There were no further comments in #298 and this branch was just for testing. We can always recreate it in the future if the original problem happens again. Thanks!

PR #556, opened 2020-12-03
Fix bug with regl-worldview context attributes

davidswinegar:
Fix bug with regl-worldview context attributes, which were not being
passed to children.


vidaaudrey (approved):
Thanks for fixing!

Issue #534, opened 2020-10-23
Error when Publisher component is rendered before rosClient connection hook runs


CebulowyNinja:
Hi,

I found an issue in the _webviz-core_, when the Webviz are connected to the rosbridge via a websocket.
If the Publisher component is rendered before the `rosClient` (roslib.Ros instance) `connection` hook runs at [RosbridgePlayer.js:98](https://github.com/cruise-automation/webviz/blob/master/packages/webviz-core/src/players/RosbridgePlayer.js#L98) an error listed below occured. It's caused by null reference at [RosbridgePlayer.js:331](https://github.com/cruise-automation/webviz/blob/master/packages/webviz-core/src/players/RosbridgePlayer.js#L331).
```
Uncaught TypeError: Cannot read property 'idCounter' of undefined
    at Topic.advertise (roslib.js:3360)
    at Topic.publish (roslib.js:3405)
    at exports.default.publish (RosbridgePlayer.js:349)
    at exports.default.publish (index.js:462)
    at exports.default.publish (OrderedStampPlayer.js:153)
    at Object.publish (index.js:251)
    at Publisher.publish (Publisher.js:50)
    at index.js:180
    at index.js:154
    at Object.ka (react-dom.production.min.js:15)
```

**Test case**
- Run Webviz connected to ros_bridge via websocket.
- Add _Publish_ panel and set any _topic_ (panel settings will be automatically saved).
- Refresh site.
- Try to publish message by _Publish_ button on Publish panel.
- Probably an error occurred at console, but it depends on timing between described code pieces.

jaguardo:
I think I have witnessed this effect without an error, but also without a notification that the publish isn't working (simply no feedback of a published message).  I've witnessed in both in the current webviz Publish panel and in the @jreyesr PR (Teleop panel), periodically the Publish msg doesn't go through.

Workaround, for both publish panels, if the topic is removed/changed to something else (anything, doesn't even have to be a current topic) and then changed back it will start working.

jreyesr:
Yes, this bug is especially obnoxious because it shows no feedback on the UI (but there should be an error message on the console, since it's attempting to access a property of an undefined var).
I have created a PR that should fix it. I'll wait for feedback on it, then (hopefully) this will be fixed.

vidaaudrey:
Thank @jreyesr for reporting and trying to fix the issue. Look forward to reviewing the PR.

PR #561, opened 2020-12-16
Teleop panel

jreyesr:
_(First time contributor, so apologies for any mistakes)_

## Summary

This PR adds a Teleoperation panel which operates in a similar way to the [teleop_twist_keyboard](http://wiki.ros.org/teleop_twist_keyboard) node. It can be controlled by clicking buttons or with the arrow keys.

Teleop panel

The main motivation of this change would be its use in remotely controlled (i.e., non-autonomous) robots. Such robots can't currently be easily controlled via Webviz (at least, I don't know of a simple way), and it would be nice to have the teleoperation interface next to, for example, an Image panel showing the robot cameras, or a Map panel showing the robot location (or both!).

This PR started by copying the Publish panel, since its functionality is similar. The major difference is that the Teleop panel sends specific messages to a fixed (configurable) topic with a fixed (non-configurable) datatype, instead of arbitrary mesages to any topic. Unnecessary settings were deleted, and the UI was edited to only contain a few buttons which trigger a message publication via a Publisher component. A global KeyListener element was also added, to allow keypresses to also trigger the messages.

## Test plan

The new panel was manually tested on a local installation, with ROS running on a Docker container and the Websockets server exposed via port 9090. This change makes no sense when using local or remote bags.
The test files from the Publish panel were also copied, adapted and run. All tests pass successfully.

## Versioning impact

This PR should only require a minor semver update. 

## Addendum: Known limitations & questions
* At present, the panel suffers from issue https://github.com/cruise-automation/webviz/issues/534 which is also present in the Publish panel, since they both use the same Publisher component. However, unlike the Publish panel, this one displays an error message over the panel, with a "Reload Panel" button. Clicking on the button (seems to) fix the error. A fix to the linked issue should (?) remove this bug.
* The panel only allows four directions (up, down, left, right) plus a Stop command. More complex robots can use up to nine keys  for direction, plus a few extra to change rotation and translation speed. Should those additional keys be implemented?
* Currently, the four directions are bound to the arrow keys, and Stop is bound to the Space key. Should the keybindings be configurable? If so, how? Maybe a dropdown for every command, with a list of valid keys?

### DIFF
--- a/packages/webviz-core/src/panels/Teleop/index.js
+++ b/packages/webviz-core/src/panels/Teleop/index.js
@@ -28,6 +28,7 @@ import { string } from "prop-types";
 
 type Config = {|
   buttonColor: string,
+  topicName: string,
 |};
 
 type Props = {
@@ -105,6 +106,7 @@ class Teleop extends React.PureComponent<Props, PanelState> {
   static panelType = "Teleop";
   static defaultConfig = {
     buttonColor: "#00A871",
+    topicName: "/cmd_vel",
   };
 
   _publisher = React.createRef<Publisher>();
@@ -212,6 +214,17 @@ class Teleop extends React.PureComponent<Props, PanelState> {
             placeholder="rgba(1,1,1,1) or #FFFFFF"
           />
         </Item>
+        <Item>
+          <PanelToolbarLabel>Topic name</PanelToolbarLabel>
+          <PanelToolbarInput
+            type="text"
+            value={config.topicName}
+            onChange={(event) => {
+              saveConfig({ topicName: event.target.value });
+            }}
+            placeholder="Choose /cmd_vel if unsure"
+          />
+        </Item>
       </>
     );
   }
@@ -220,7 +233,7 @@ class Teleop extends React.PureComponent<Props, PanelState> {
     const {
       capabilities,
       topics,
-      config: { buttonColor },
+      config: { buttonColor, topicName },
     } = this.props;
 
     const { datatypeNames, parsedObject, error, pressing } = this.state;
@@ -228,7 +241,9 @@ class Teleop extends React.PureComponent<Props, PanelState> {
 
     return (
       <Flex col style={{ height: "100%", padding: "12px" }} onKeyPress={this.handleKey}>
-        <Publisher ref={this._publisher} name="Publish" topic="/cmd_vel" datatype="geometry_msgs/Twist" />
+        {topicName && (
+          <Publisher ref={this._publisher} name="Publish" topic={topicName} datatype="geometry_msgs/Twist" />
+        )}
         
         <KeyListener global keyDownHandlers={this._keyHandlers} keyUpHandlers={this._keyHandlers} />
         <PanelToolbar floating menuContent={this._renderMenuContent()} />
--- a/packages/webviz-core/src/panels/Teleop/index.js
+++ b/packages/webviz-core/src/panels/Teleop/index.js
@@ -1,0 +1,276 @@
+// @flow
+//
+//  Copyright (c) 2019-present, Cruise LLC
+//
+//  This source code is licensed under the Apache License, Version 2.0,
+//  found in the LICENSE file in the root directory of this source tree.
+//  You may not use this file except in compliance with the License.
+import CheckboxBlankOutlineIcon from "@mdi/svg/svg/checkbox-blank-outline.svg";
+import CheckboxMarkedIcon from "@mdi/svg/svg/checkbox-marked.svg";
+import * as React from "react";
+import { hot } from "react-hot-loader/root";
+import styled from "styled-components";
+
+import { PanelToolbarInput, PanelToolbarLabel } from "webviz-core/shared/panelToolbarStyles";
+import Autocomplete from "webviz-core/src/components/Autocomplete";
+import Button from "webviz-core/src/components/Button";
+import Flex from "webviz-core/src/components/Flex";
+import Item from "webviz-core/src/components/Menu/Item";
+import Panel from "webviz-core/src/components/Panel";
+import PanelToolbar from "webviz-core/src/components/PanelToolbar";
+import Publisher from "webviz-core/src/components/Publisher";
+import KeyListener from "webviz-core/src/components/KeyListener";
+import { PlayerCapabilities, type Topic } from "webviz-core/src/players/types";
+import colors from "webviz-core/src/styles/colors.module.scss";
+import { colors as c } from "webviz-core/src/util/sharedStyleConstants";
+import type { RosDatatypes } from "webviz-core/src/types/RosDatatypes";
+import { string } from "prop-types";
+
+type Config = {|
+  buttonColor: string,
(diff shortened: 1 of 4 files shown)


packages/webviz-core/src/panels/Teleop/index.js
```diff
@@ -0,0 +1,291 @@
...
+    const { datatypeNames, parsedObject, error, pressing } = this.state;
+    const canPublish = capabilities.includes(PlayerCapabilities.advertise);
+
+    return (
+      <Flex col style={{ height: "100%", padding: "12px" }} onKeyPress={this.handleKey}>
+        {topicName && (
+          <Publisher ref={this._publisher} name="Publish" topic={topicName} datatype="geometry_msgs/Twist" />
+        )}
+        
+        <KeyListener global keyDownHandlers={this._keyHandlers} keyUpHandlers={this._keyHandlers} />
```
vidaaudrey, line 248: Let's keep the key listerner local? We'd want to preserve the arrow keys for global usage, e.g. currently left/right arrow will seek back/forward 100ms.

packages/webviz-core/src/panels/Teleop/index.js
```diff
@@ -0,0 +1,291 @@
...
+    } else {
+      parsedObject = parsedAny;
+    }
+  } catch (e) {
+    error = value ? e.message : "";
+  }
+  return { error, parsedObject };
+}
+
+class Teleop extends React.PureComponent<Props, PanelState> {
```
vidaaudrey, line 105: Could you convert this component to functional component. We've converted the majority of components but still a few left. Would be nice to use the new syntax.

packages/webviz-core/src/panels/Teleop/index.js
```diff
@@ -0,0 +1,291 @@
...
+
+    return (
+      <>
+        <Item>
+          <PanelToolbarLabel>Button color (rgba or hex)</PanelToolbarLabel>
+          <PanelToolbarInput
+            type="text"
+            value={config.buttonColor}
+            onChange={(event) => {
+              saveConfig({ buttonColor: event.target.value });
```
vidaaudrey, line 212: Curious what's your use case for customizing button color?

packages/webviz-core/src/panels/Teleop/index.js
```diff
@@ -0,0 +1,291 @@
...
+            placeholder="rgba(1,1,1,1) or #FFFFFF"
+          />
+        </Item>
+        <Item>
+          <PanelToolbarLabel>Topic name</PanelToolbarLabel>
+          <PanelToolbarInput
+            type="text"
+            value={config.topicName}
+            onChange={(event) => {
+              saveConfig({ topicName: event.target.value });
```
vidaaudrey, line 223: Would need some data validation to ensure the saved topic name is valid

vidaaudrey (commented):
Thanks for your contribution.  Added a few comments. There are some lint errors in CI. If you are using vscode, you could install `eslint` to auto format the code. Let me know if you need any help.

packages/webviz-core/src/panels/Teleop/index.js
jreyesr: Done. It makes things a bit awkward, though, you now need to click on any button before the key presses are sent to the component. I guess it can't be avoided, you need to give focus to the component if it's not global.

packages/webviz-core/src/panels/Teleop/index.js
jreyesr: Done (I think so, at least, I don't really work on React and so I had to google what a functional component is). Can you confirm that I actually converted the component to functional style, please?

packages/webviz-core/src/panels/Teleop/index.js
jreyesr: Nothing actually, it's a remnant from the Publish panel from which I pulled the original code. I deleted it.

packages/webviz-core/src/panels/Teleop/index.js
jreyesr: Not sure about what would be required to validate the topic. Would it require checking that the topic name exists and it's of the correct datatype? Doing that would limit use a lot, I think, since it may very well happen that the topic doesn't "exist" since no other nodes have ever published to it, and the Teleop panel will be the only one publishing. Any pointers? 
Perhaps a) if it doesn't exist, just accept it, and b) if it already exists, ensure that the datatype is actually "geometry_msgs/Twist"?

jaguardo:
@jreyesr I tried the changes (granted, I might be missing something as well!) but I couldn't get the pop up panel for the add panel or change panel to list the teleop as an option... is there something else that has to changes after making those changes? thanks!  From the pictures, I love the functionality, great!

jreyesr:
@jaguardo Weird. I just ran `build-static-webviz` from scratch and it does show me the new panel (on Chrome, Linux). Also tested it on Edge 87.0.664.60, and it works nicely. No idea where that error came from, I do use a KeyListener, but I use it in the same way that other panels (e.g., the Plot panel) use it.

jreyesr:
@vidaaudrey I added a few commits fixing the linter errors and converting to a functional component. CI still fails, but the error now happens when attempting to upload something (screenshots, I think) to S3; the error says `Missing credentials in config`. 
Still unsure about the data validation (see https://github.com/cruise-automation/webviz/pull/561#discussion_r544742818).

vidaaudrey:
@jreyesr we are aware of the CI issue and we have a task to migrate the screenshot testing to GCP soon. At the same time, I just made a copy of your branch at https://github.com/cruise-automation/webviz/pull/567 so you could check the CI output in 1 hour or so.

packages/webviz-core/src/panels/Teleop/index.js
vidaaudrey: Sorry about the late reply. A simple validation like this should be ok:
- name is not empty 
- name should start with `/`
- name should be at least 2 characters 

You could refer to some of the validators here: https://github.com/cruise-automation/webviz/blob/master/packages/webviz-core/shared/validators.js#L102-L107

jreyesr: I've added some validation on 671f133. It shows an error message on the panel if those conditions are not met. It also disables all the buttons and keypresses do nothing.

amacneil:
For anyone watching this issue - we recently added a teleop panel to [Foxglove Studio](https://github.com/foxglove/studio) (webviz fork) in [#1429](https://github.com/foxglove/studio/pull/1429).

pringithub:
> @jaguardo Weird. I just ran `build-static-webviz` from scratch and it does show me the new panel (on Chrome, Linux). Also tested it on Edge 87.0.664.60, and it works nicely. No idea where that error came from, I do use a KeyListener, but I use it in the same way that other panels (e.g., the Plot panel) use it.

@jreyesr Doesn't work for me either. I did a clean build and it doesn't show up with `serve-static-webviz`. Something I'm doing wrong? It seems

PR #563, opened 2020-12-22
Closes #534

### LINKED ISSUE #534: Error when Publisher component is rendered before rosClient connection hook runs

jreyesr:
## Summary

This PR fixes the bug reported in #534: when using a Rosbridge server, there is a race condition that may cause Publisher components (used, for example, in the Publish panel and the -currently unmerged- Teleop panel) to get an undefined `rosClient`, which causes a panel crash when a message is published. Current workaround is to force a rerender of the Publisher component by changing any of its props (for example, the published topic) as specified in https://github.com/cruise-automation/webviz/issues/534#issuecomment-749531771.

This PR fixes the bug by adding a new array of pending publishers. If the `setPublishers` method of RosbridgePlayer is called before `rosClient` is set, the publishers are instead buffered on the new array. Any `publish()` calls are rejected gracefully. When the `connect` hook is eventually called and `rosClient` gets a value, the most recent set of pending publishers is sent to `setPublishers`.

## Test plan

This change was tested using the Publish and Teleop panels (which instantiate a Publisher component) and a Rosbridge Websocket server. The test case detailed in https://github.com/cruise-automation/webviz/issues/534#issue-728355717 no longer fails.

## Versioning impact

This PR is a patch revision.

### DIFF
--- a/packages/webviz-core/src/players/RosbridgePlayer.js
+++ b/packages/webviz-core/src/players/RosbridgePlayer.js
@@ -62,6 +62,7 @@ export default class RosbridgePlayer implements Player {
   _bobjects: BobjectMessage[] = []; // Queue of bobjects that we'll send in next _emitState() call.
   _messageOrder: TimestampMethod = "receiveTime";
   _requestTopicsTimeout: ?TimeoutID; // setTimeout() handle for _requestTopics().
+  _pendingTopicPublishers: ?(AdvertisePayload[]) = null;
   _topicPublishers: { [topicName: string]: ROSLIB.Topic } = {};
   _parsedMessageDefinitionsByTopic: ParsedMessageDefinitionsByTopic = {};
   _bobjectTopics: Set<string> = new Set();
@@ -101,6 +102,12 @@ export default class RosbridgePlayer implements Player {
       }
       this._rosClient = rosClient;
       this._requestTopics();
+
+      // If required, set publishers to what was requested before connection was established
+      if (this._pendingTopicPublishers) {
+        this.setPublishers(this._pendingTopicPublishers);
+        this._pendingTopicPublishers = null;
+      }
     });
 
     rosClient.on("error", (error) => {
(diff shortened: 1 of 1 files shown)


jasonimercer:
@jreyesr , I see foxglove used this commit. We're looking to cherry-pick this into our internal build of webviz. Since you've put up the PR here over a year ago, have you had and negative experience with the change in your builds?

Just trying to exercise extra due diligence as we're looking to incorporate this change late in our internal release schedule.

jasonimercer:
I should mention this change has addressed the same issue it was meant to correct (in our tests). We just don't know the webviz architecture enough to understand unintended side effects.

PR #565, opened 2020-12-22
Upgrade prettier to 1.16.0

hassoncs:
This includes the `eslint --fix` for all the files to match prettier's new format. The main change is react callbacks now take up less lines than they did before.

image

https://prettier.io/blog/2019/01/20/1.16.0.html

Using "hide whitespace changes" makes the diff much nicer:
image


vidaaudrey (approved):
Nice!

davidswinegar (approved):

PR #575, opened 2021-01-06
Make Worldview worker-friendly

hhsaez:
## Summary

In order to support rendering in workers, Worldview needs to be modified to avoid depending on React and the DOM. This PR, which is the first of a series of improvements that we're planning to do during the next weeks, includes changes to export some of the existing rendering commands like lines, cubes, etc in a way that can be used in workers.

Commands are exported in two different ways:
* If the command is ready to be used as is, we just export it using it's name in lowercase (i.e. `lines`, `cubes`, `spheres`, etc).
* If the command needs extra params in order to be used, we export a `make*Command` (i.e. `makePointsCommand`, `makeGLTextCommand`). I think I might use this approach for all commands in the future even if they take no params in order to have a unique export mechanism.

## Test plan

Covered by existing tests

## Versioning impact

Minor. Changes are backward compatible.


packages/regl-worldview/src/utils/Dimensions.js
```diff
@@ -26,7 +28,8 @@ class ResizeObserverMock {
   }
   unobserve() {}
 }
-const ResizeObserverImpl = process.env.NODE_ENV === "test" ? (ResizeObserverMock: any) : ResizeObserver;
+const ResizeObserverImpl =
+  process.env.NODE_ENV === "test" || inWebWorker() ? (ResizeObserverMock: any) : ResizeObserver;
```
MatthewSteel, line 32: I'm not sure about this. Does it make sense to have a `Dimensions` object/component inside the worker? I suspect we don't need support here. (If we don't, we probably don't need the implementation of `inWebWorker`.)

packages/regl-worldview/src/commands/Points.js
```diff
@@ -104,6 +104,8 @@ const makePointsCommand = (useWorldSpaceSize: boolean) => {
   };
 };
 
+export const points = makePointsCommand(false);
```
MatthewSteel, line 107: Not letting the param be configurable might give us a headache in the future -- webviz uses it both ways. Maybe we should export `makePointsCommand` instead.

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -559,6 +568,18 @@ function makeTextCommand(alphabet?: string[]) {
...
+  const command = makeTextCommand(props.alphabet);
+  // HACK: Worldview doesn't provide an easy way to pass a command-level prop into the regl commands,
+  // so just attach it to the command object for now.
+  command.autoBackgroundColor = props.autoBackgroundColor;
+  command.resolution = Math.max(MIN_RESOLUTION, props.resolution || DEFAULT_RESOLUTION);
+  command.scaleInvariant = props.scaleInvariantFontSize != null;
+  command.scaleInvariantSize = props.scaleInvariantFontSize ?? 0;
+  command.textAtlas = props.textAtlas;
+  return command;
+};
```
MatthewSteel, line 581: Would be nice for the `GLText` export to use this logic too.

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -53,6 +53,15 @@ type TextMarkerProps = TextMarker & {
...
   highlightColor?: ?Color,
 };
+
+type GLTextProps = {|
+  autoBackgroundColor?: boolean,
+  scaleInvariantFontSize?: number,
+  resolution?: number,
+  alphabet?: string[],
+  textAtlas?: GeneratedAtlas,
+|};
```
MatthewSteel, line 63: Can use this inside `Props` below as another `...spread`. No strong feelings either way from me.

MatthewSteel (commented):
Looking good, pretty clean way forward.

packages/regl-worldview/src/commands/GLText.js
hhsaez: I agree, but I had to disable flow checks because of the indexed property in `CommonCommandProps`

packages/regl-worldview/src/utils/Dimensions.js
hhsaez: We don't need `Dimensions` inside workers. But since I haven't make every feature of `Worldview` to be worker-friendly yet, not having this check ends up causing a runtime error because `ResizeObserver` is not defined. I'm going to get back to this in a future PR

packages/regl-worldview/src/commands/Points.js
hhsaez: Agreed. Check my updated description in this PR. I'm using two different approaches based on whether or not the command requires extra params.

packages/regl-worldview/src/utils/Dimensions.js
MatthewSteel: Fair enough. Maybe add a TODO to remove the worker check if we think we'll fix the import order later, but not a big deal.

packages/regl-worldview/src/commands/GLText.js
MatthewSteel: Ah dang, fine either way I think 👍 .

MatthewSteel (approved):

PR #580, opened 2021-01-12
Add webviz-dev workflow

troygibb:
## Summary

Currently we have a pretty non-intuitive way to dev on Webviz, which is via `npm run docs` and navigating to `http://localhost:8080/app` to view Webviz. 

This update adds a quick hack around that, by re-tooling our Webpack config with a `WEBVIZ_DEV` flag that sets up all the relative paths and html-webpack-plugin.

Other things:
- Added a quick update to the docs as well to reflect this new workflow.
- I noticed a few `javascript heap out of memory` errors-- hence `NODE_OPTIONS='--max-old-space-size=4096'`

## Test Plan

Verified things looked good locally.


README.md
```diff
@@ -66,6 +66,24 @@ Similarly to recording videos, we have a script to measure the performance of a
...
+
+### Docs
+
+```npm run docs```
+
+### Storybook
+
+```npm run storybook```
+
+Useful for `regl-worldview` changes.
```
vidaaudrey, line 83: The other way to see worldview changes is to run `npm run watch` and `npm run docs` in two different tabs. Maybe we could combine and create `docs dev`?

README.md
```diff
@@ -66,6 +66,24 @@ Similarly to recording videos, we have a script to measure the performance of a
...
+
+### Full list of commands
```
vidaaudrey, line 85: These are called `scripts` in the npm :) Technically we have other scripts as well, maybe change to `Other scripts`?

package.json
```diff
@@ -133,7 +134,7 @@
     "build-storybook": "build-storybook -c stories",
     "flow": "flow",
     "flow-typed-rebuild": "rm -rf flow-typed/ && flow-typed install && flow-typed install --rootDir packages/@cruise-automation/hooks && flow-typed install --rootDir packages/@cruise-automation/button && flow-typed install --rootDir packages/@cruise-automation/tooltip && flow-typed install --rootDir packages/regl-worldview",
-    "docs": "DEV_SERVER=true webpack-dev-server",
+    "docs": "NODE_OPTIONS='--max-old-space-size=4096' DEV_SERVER=true webpack-dev-server",
```
vidaaudrey, line 137: Curious did you hit some performance issue in the open source version or just thought it's a good idea to add `--max-old-space-size`?

vidaaudrey (approved):
LGTM. A few nit comments.

package.json
troygibb: i mention it in the pr description-- i was hitting errors without this option

README.md
troygibb: docs-dev is good! I'll updated

README.md
troygibb: Will do

PR #584, opened 2021-01-20
Add support for rendering Worldview in an offscreen canvas

hhsaez:
## Summary

In this PR I'm modifying Worldview so it can use an offscreen `<canvas />` that is managed by the client application. If an offscreen canvas is used, clients must use the `<OffscreenWorldview />` component instead of `<Worldview />` one since dimensions are handled externally.

Additional changes are required to forward mouse events to rendering components, since they're now initiated in an external element. 

## Test plan

Handled by existing tests in Webviz. 

## Versioning impact

Minor. These changes are backward compatible.

### DIFF
--- a/packages/regl-worldview/src/Worldview.js
+++ b/packages/regl-worldview/src/Worldview.js
@@ -65,6 +65,7 @@ export type BaseProps = {|
 
   // Context attributes passed into canvas.getContext.
   contextAttributes?: ?{ [string]: any },
+  offscreenCanvas: HTMLCanvasElement,
 |};
 
 type State = {|
@@ -115,7 +116,9 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
       defaultCameraState,
       hitmapOnMouseMove,
       disableHitmapForEvents,
+      offscreenCanvas,
     } = props;
+    this._canvas.current = offscreenCanvas;
     if (onCameraStateChange) {
       if (!cameraState) {
         console.warn(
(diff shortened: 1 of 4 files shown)


packages/regl-worldview/src/Worldview.js
```diff
@@ -332,13 +354,16 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
...
+      offscreenCanvas,
     } = this.props;
     const { worldviewContext } = this.state;
     // If we are supplied controlled camera state and no onCameraStateChange callback
     // then there is a 'fixed' camera from outside of worldview itself.
     const isFixedCamera = cameraState && !onCameraStateChange;
     const canvasScale = resolutionScale || 1;
-    const canvasHtml = (
+    const canvasHtml = offscreenCanvas ? (
+      <React.Fragment />
```
MatthewSteel, line 365: Can probably just be `null`.

packages/regl-worldview/src/Worldview.js
```diff
@@ -197,41 +202,58 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
...
+  ) => {
     const { worldviewContext } = this.state;
     const worldviewHandler = this.props[mouseEventName];
 
-    if (!(e.target instanceof window.HTMLElement) || e.button !== 0) {
+    if (!fromOffscreenTarget && (!(e.target instanceof window.HTMLElement) || e.button !== 0)) {
       return;
     }
 
+    // $FlowFixMe: Because of `fromOffscreenTarget`, target might not be an actual HTMLElement instance but still needs to implement `getBoundingClientRect`
```
MatthewSteel, line 256: What's the error? We're kinda lying about the type -- `e.target` _is_ always an `HTMLElement` as far as Flow is aware, isn't it?

packages/regl-worldview/src/Worldview.js
```diff
@@ -197,41 +202,58 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
...
+  _onMouseInteraction = (
+    e: SyntheticMouseEvent<HTMLCanvasElement>,
+    mouseEventName: MouseEventEnum,
+    fromOffscreenTarget: boolean
+  ) => {
     const { worldviewContext } = this.state;
     const worldviewHandler = this.props[mouseEventName];
 
-    if (!(e.target instanceof window.HTMLElement) || e.button !== 0) {
+    if (!fromOffscreenTarget && (!(e.target instanceof window.HTMLElement) || e.button !== 0)) {
```
MatthewSteel, line 252: Worth a comment saying why these checks aren't needed when the canvas is provided I think (or why they _are_ needed when the canvas isn't provided, if we know.)

packages/regl-worldview/src/Worldview.js
```diff
@@ -197,41 +202,58 @@ export class WorldviewBase extends React.Component<BaseProps, State> {
...
+  _onMouseUp = (e: SyntheticMouseEvent<HTMLCanvasElement>, fromOffscreenTarget: boolean) => {
+    this._onMouseInteraction(e, "onMouseUp", !!fromOffscreenTarget);
     const { _dragStartPos } = this;
     if (_dragStartPos) {
       const deltaX = e.clientX - _dragStartPos.x;
       const deltaY = e.clientY - _dragStartPos.y;
       const distance = Math.sqrt(deltaX * deltaX + deltaY * deltaY);
       if (distance < DEFAULT_MOUSE_CLICK_RADIUS) {
-        this._onMouseInteraction(e, "onClick");
+        this._onMouseInteraction(e, "onClick", !!fromOffscreenTarget);
```
MatthewSteel, line 238: `!!bool` seems redundant, here and elsewhere.

packages/regl-worldview/src/Worldview.js
```diff
@@ -65,6 +65,7 @@ export type BaseProps = {|
 
   // Context attributes passed into canvas.getContext.
   contextAttributes?: ?{ [string]: any },
+  offscreenCanvas?: HTMLCanvasElement,
```
MatthewSteel, line 68: Any thoughts on lining up the name and the type? IMO it'd be nice to either call it `canvas` or have the type be either `OffscreenCanvas` or `HTMLCanvasElement | OffscreenCanvas`. But I don't mind this too much either.

packages/regl-worldview/src/Worldview.js
hhsaez: Because of the check for `fromOffscreenTarget` a couple of lines above, the evaluation of `e.target instanceof window.HTMLElement` type is now optional (see my next comment). Then, `e.target` might be an instance of `HTMLElement`, but Flow can be 100% sure about that. The error is that `getBoundingClientRect` is not defined for `EventTarget`.

packages/regl-worldview/src/Worldview.js
hhsaez: The check is required because `window` is not defined for a web worker and the check throws an error. Checking for `fromOffscreenTarget` first prevents the error and also ignores the actual type of `e.target`, which is not an actual `HTMLElement`.

MatthewSteel: Makes sense. Still, I think it's worth a comment in the code -- maybe just "assume events coming from handleOffscreenMouseEvent have relevant targets".

packages/regl-worldview/src/index.js
```diff
@@ -6,7 +6,7 @@
 //  found in the LICENSE file in the root directory of this source tree.
 //  You may not use this file except in compliance with the License.
 
-import Worldview from "./Worldview";
+import Worldview, { WorldviewBase as OffscreenWorldview } from "./Worldview";
```
MatthewSteel, line 9: The main difference here is that `WorldviewBase` doesn't use a `ContainerDimensions`? I can't think of a better name, but I feel like that's the important difference -- not necessarily how we use it. But this name is ok too.

MatthewSteel (approved):

packages/regl-worldview/src/index.js
hhsaez: That's correct. We need `WorldviewBase` because it's not using `ContainerDimensions`. I can't find a better name either. The alternative was to just export it as `WorldviewBase`

MatthewSteel: Maybe `ControlledSizeWorldview` or `FixedSizeWorldview`?

PR #596, opened 2021-03-10

hassoncs:
Changelog

### Fixes
- Fixed flicker in the playback controls tooltip.
- Added ability to preload audios in the Audio panel.
- Fixed messages with header stamps exactly equal to the playback time not being displayed in "Header stamp" playback mode.
- Large 64-bit integer values are now rendered accurately in the Raw Messages panel. 
- Fixed a bug where Webviz would exit fullscreen mode when switching applications with `command + tab` on a Mac. 
- Fixed crash in the Raw Messages panel when diffing equal arrays.
- Added support to visualize icon markers in the Image and 3D panel.

### Improvements
- Added ability to seek by precise ROS time by using `?seek-to=SEC.NSEC`.
- Added per panel keyboard shortcuts to the global shortcuts modal (open with `cmd + /`).
- Added the ability to resize the bottom section of the Node Playground panel.
- Added automatic support for `/webviz_source_2` topics in Node Playground scripts. 
- Added support to synchronize audio playback cursor with Plot panel and the playback bar. 
- Added more enum support in the Raw Messages panel. 
- Added `Global Variable Dropdown` panel, which takes a topic path to populate a dropdown menu. User can then specify the global variable that should be set to the dropdown menu's selected option.
- Added more relevant info and cleaned up styling in the Connection Picker.
- The full-screen panel state now persists across page refreshes, and is encoded in shared URLs.
- Row/cell expansion in the Table Panel persists through playback and page refreshes.
- Added conditional formatting to the Table panel.


vidaaudrey (approved):
🎉 🎉 🎉

janpaul123:
Nice work!! Did any particular perf work ship? Pointclouds feel faster to me, but maybe it's my imagination :)

PR #601, opened 2021-03-16
Use a default sampler for textures if none is defined in GLTF model

hhsaez:
## Summary

`<GLTFScene/>` command assumes a sampler is always present for every texture, which is not the case for some models. Since the [GLTF 2.0 Specification](https://github.com/KhronosGroup/glTF/blob/master/specification/2.0/README.md#texture) indicates that `sampler` is an optional property, I'm making a small change here to use a default sampler if none is provided by the model.

## Test plan

Manually tested

## Versioning impact

Patch. This is a bug fix and backward compatible.


troygibb (commented):
Nice! Very slick fix. Do we have any screenshot coverage for gltf models? If not, we should add some here. I can try to create a dummy model in Blender to repro the issue (in case you aren't using one already)

hhsaez:
> Nice! Very slick fix. Do we have any screenshot coverage for gltf models? If not, we should add some here. I can try to create a dummy model in Blender to repro the issue (in case you aren't using one already)

I added a new story in my latest commit, but it uses the Duck model. I tried exporting models from Blender to reproduce this problem, but all of them include the default samplers. Maybe I should use an older Blender version? If you do have a file at hand, please send it to me and I'll update the story.

troygibb (approved):
👌

Issue #600, opened 2021-03-16
withTransforms() (created by ForwardRef) issue running `npm run webviz-dev`


cjds:
Getting this when I run `npm run webviz-dev` .  

This version looks broken to me
```
root@FR-1603:~/src/webviz# node --version
v10.22.1
root@FR-1603:~/src/webviz# git log -n1
commit f16b3691951b7a177a0e27dc45b509fa3f884c31 (HEAD, cruise/master)
Author: Chris Hasson <chris@remix.com>
Date:   Thu Mar 11 14:54:15 2021 -0800

    Update packages/webviz-core from internal repo (#596)
    
    Changelog:
    -
    
    Co-authored-by: Chris Hasson <chris.hasson@getcruise.com>
```

![image](https://user-images.githubusercontent.com/4957157/111340310-be5bad80-8635-11eb-8538-4d01e1881991.png)

```
Error stack:
Invariant Violation: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://fb.me/react-invalid-hook-call for tips about how to debug and fix this problem.
    at eval (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:16237:26)
    at Object.throwInvalidHookError (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:16240:5)
    at Object.useRef (webpack-internal:///./node_modules/react/cjs/react.development.js:1645:21)
    at withTransforms() (webpack-internal:///./packages/webviz-core/src/panels/ThreeDimensionalViz/withTransforms.js:30:30)
    at Component.hotComponentRender (webpack-internal:///./packages/webviz-core/node_modules/react-hot-loader/dist/react-hot-loader.development.js:742:16)
    at Component.proxiedRender (webpack-internal:///./packages/webviz-core/node_modules/react-hot-loader/dist/react-hot-loader.development.js:770:36)
    at finishClassComponent (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:18407:31)
    at updateClassComponent (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:18360:24)
    at beginWork$1 (webpack-internal:///./node_modules/react-dom/cjs/react-dom.development.js:20108:16)
[...]

PR #602, opened 2021-03-16
Fix withTransforms() breaks the rules of hooks

hassoncs:
Fix this issue where `withTransforms` seems to violate the rules of hooks: https://github.com/cruise-automation/webviz/issues/600

React is complaining that returning a function component here breaks the "rules of hooks".
I tried setting `Component.prototype = React.Component.Prototype` like the warning suggests, but it didn't seem to help.

I messed with it for a while trying to get it to work, but couldn't figure out why React is fine with this approach in `FrameCompatibilityDEPRECATED`, but doesn't like it in `withTransforms`. It might be due to the double layer of `hoistNonReactStatics`/`React.forwardRef` in both `FrameCompatibilityDEPRECATED` and `withTransforms`? I'm not entirely sure, but switching it back to a class component fixes the problem in the short-term.

### DIFF
--- a/packages/webviz-core/package-lock.json
+++ b/packages/webviz-core/package-lock.json
@@ -32,10 +32,27 @@
         }
       }
     },
+    "@cruise-automation/button": {
+      "version": "0.0.7",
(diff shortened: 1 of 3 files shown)


MatthewSteel (approved):

80avin:
Why these changes are not available in the master branch ?

PR #603, opened 2021-03-18
Update flow to 0.110.0 and update from internal repo

hassoncs:
Changelog:
- Fixed a bug where playback would get stuck in the "paused" state.
- Fixed a bug where Webviz would show no data momentarily when changing global variables quickly.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15887,9 +15887,9 @@
       }
     },
     "flow-bin": {
-      "version": "0.106.2",
-      "resolved": "https://registry.npmjs.org/flow-bin/-/flow-bin-0.106.2.tgz",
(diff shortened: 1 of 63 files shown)


davidswinegar (approved):

jasonimercer:
@hassoncs , it looks like this fix has helped the stuck paused issue when streaming bag data from a remote source but I can still trigger it when dragging a bag into the browser.

To reproduce for the bag in browser case, I had to let the full bag get processed so the light grey fills the entire bar. Up until that point, pause/play would work. After that point a single pause would stick.

I'm not sure if the remote source fix is just because it takes longer to fully cache? 

Darn, if a remote source is allowed to fully cache, the stuck pause resurfaces.

PR #614, opened 2021-04-06
Worldview: Paint again if onDirty was called during paint

jtbandes:
## Summary

In some cases, painting may trigger a React state update (since components can register arbitrary functions to be called during `paint()`) which causes us to call `onDirty` again. But since a paint is currently in progress, we just ignore the `onDirty` call, which means that the new state never gets drawn until another paint is triggered in the future.

Fix this by keeping track of whether onDirty is called during paint, and if so just queue up another frame.

## Test plan

Tested locally by toggling perspective mode, and verifying that the change happens immediately instead of on the next render.

I'm not sure why this doesn't cause an issue on webviz.io, but I was testing in Foxglove Studio which is on React 17, which may make some difference in how/when updates are dispatched. There was a state update happening during paint in `Overlay`, and I think the World component's state update might have been batched with this: https://github.com/cruise-automation/webviz/blob/d82468c05fb2d3aeca670ff37999faeafa4d7ffc/packages/regl-worldview/src/commands/Overlay.js#L61

## Versioning impact

Should be a patch. I guess there is some risk of an infinite render loop, but I think someone would have to be doing some pretty weird things in their component for this to be a problem.

### DIFF
--- a/packages/regl-worldview/src/WorldviewContext.js
+++ b/packages/regl-worldview/src/WorldviewContext.js
@@ -84,6 +84,7 @@ export class WorldviewContext {
   _compiled: Map<Function, CompiledReglCommand<any>> = new Map();
   _drawCalls: Map<React.Component<any>, DrawInput> = new Map();
   _frame: ?AnimationFrameID;
+  _needsPaint = false;
   _paintCalls: Map<PaintFn, PaintFn> = new Map();
   _hitmapObjectIdManager: HitmapObjectIdManager = new HitmapObjectIdManager();
   _cachedReadHitmapCall: ?{
@@ -233,6 +234,7 @@ export class WorldviewContext {
   }
 
   _paint() {
+    this._needsPaint = false;
     const start = Date.now();
     this.reglCommandObjects.forEach((cmd) => (cmd.stats.count = 0));
     if (!this.initializedData) {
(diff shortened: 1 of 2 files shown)


packages/regl-worldview/src/WorldviewContext.js
```diff
@@ -233,6 +234,7 @@ export class WorldviewContext {
   }
 
   _paint() {
+    this._needsPaint = false;
```
vidaaudrey, line 237: Should we set `this._needsPaint=false` at the end of this function? I'm a little confused.

jtbandes: The goal is to see if it was set to true before the paint finished

vidaaudrey: ah, that makes sense. I didn't read your PR description carefully.

vidaaudrey (approved):
LGTM. We haven't fixed the CI problem so we'd need to commit the change internally and push to master.  Will do it later and publish a new version

vidaaudrey (commented):
Maybe update the package version as well?

vidaaudrey:
I opened #616 which is a copy of this one and small fix on package dependency. We had to go through the flow to by-pass the CI auth issue now. Hopefully we can get it fixed soon.

jtbandes:
Merge as #616

PR #615, opened 2021-04-06

hassoncs:
### Changelog

- Made Node Playground scripts with large outputs faster. There may be changes in behavior. If you see node playground outputs behaving strangely (most likely: markers in the 3D panel remaining in the scene too long), this new behavior can be deactivated. In the future the fallback option will be removed though, so contact us for guidance around updating your nodes/layouts.
- Added a GUI filter to the Table panel for each column.
- Added enum support in the Table panel.
- Added support for column filtering in the Table panel.
- Added a dropdown for expanded columns in the Table panel.
- Added support to visualize `nav_msgs/Path` in the 3D panel.
- Fixed the playback bar getting truncated in header-stamp mode.


vidaaudrey (approved):
🎉

PR #616, opened 2021-04-07
Worldview: Paint again if onDirty was called during paint (copy of #614)

vidaaudrey:
Copy of #614 with minor fix on build error (missing package).


jtbandes (approved):
Thanks!

PR #620, opened 2021-04-09
Fix rollup config on Windows

amacneil:
## Summary

This is the final issue preventing us from building worldview from source on Windows (allowing us to import worldview from source into Foxglove where we require Windows compatibility).

When running rollup on windows, the dist bundle was nearly empty (<1kb). After much debugging, I found the cause to be this `isExternal` function, which was treating paths like `C:\foo\bar.js` as external (and thus not bundling them). The solution was pretty easy after many hours of debugging.

## Test plan

Tested rollup on macos, linux, and windows, and they now produce identical output.

## Versioning impact

n/a

### DIFF
--- a/common/rollup.config.js
+++ b/common/rollup.config.js
@@ -26,7 +26,16 @@ const input = "src/index.js";
 const libraryName = "ReglWorldview";
 
 const globals = { react: "React", "react-dom": "ReactDOM" };
-const isExternal = (id) => !id.startsWith(".") && !id.startsWith("/");
+const isExternal = (id) => {
+  if (id.startsWith(".")) {
+    return false;
+  }
+  if (id.startsWith(__dirname)) {
+    return false;
+  }
+
+  return true;
+};
 
 export default function(pkg) {
   return [


common/rollup.config.js
```diff
@@ -26,7 +26,16 @@ const input = "src/index.js";
 const libraryName = "ReglWorldview";
 
 const globals = { react: "React", "react-dom": "ReactDOM" };
-const isExternal = (id) => !id.startsWith(".") && !id.startsWith("/");
+const isExternal = (id) => {
+  if (id.startsWith(".")) {
+    return false;
+  }
+  if (id.startsWith(__dirname)) {
```
vidaaudrey, line 33: Nice find! Maybe add a comment inline?

vidaaudrey (approved):
LGTM. Thanks for the contribution. Our CI is failing for external contributors and the team hasn't had time to migrate the screenshots from AWS to GCS yet. Once you are ready to merge, I could create a PR to combine your recent PRs so we can get it merged.

amacneil:
Closing because Foxglove Studio has moved away from Worldview. Feel free to reuse PR if needed.

PR #622, opened 2021-04-16
Make GLText work in workers

MatthewSteel:
This is half-reasonable, half-hack:
* Preferring `OffscreenCanvas`, and detecting it at runtime is
  reasonable.
* Hacking `document` for TinySDF is a bit crazy. I've raised an issue
  upstream, hopefully this won't live long.

Upstream issue: https://github.com/mapbox/tiny-sdf/issues/35

Test plan: Linked against a local webviz and opened a bag containing a
strange character with the worker flag on.

### DIFF
--- a/packages/regl-worldview/package.json
+++ b/packages/regl-worldview/package.json
@@ -1,6 +1,6 @@
 {
   "name": "regl-worldview",
-  "version": "0.16.1",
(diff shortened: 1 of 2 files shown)


hhsaez (approved):
Nice hack!

PR #625, opened 2021-04-30
GLTFScene: basic support for KHR_materials_pbrSpecularGlossiness extension

jtbandes:
Handle `material.extensions.KHR_materials_pbrSpecularGlossiness` by using the `diffuseTexture` in place of the `pbrMetallicRoughness.baseColorTexture`. This is not a perfect/complete implementation, however, it's better than having no support at all for pbrSpecularGlossiness!

Other small fixes:
- Set `wrapS` and `wrapT` defaults to REPEAT as described in https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#sampler
- Set `byteOffset` default to 0 as described in https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#accessor

Example model from: [SpecGlossVsMetalRough](https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/SpecGlossVsMetalRough)

The two bottles are supposed to look identical, so obviously this is not a complete rendering implementation.

image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+common/fixtures/SpecGlossVsMetalRough.glb filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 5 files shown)


hhsaez (approved):
These changes look good to me. Although, like you mentioned in the description, the end result might not look correct, since we don't really have support for any lighting model. 

Thanks for adding the extra fixes as well!

I'm not sure how to fix the CI errors, though. I'll ask around.

PR #626, opened 2021-04-30
GLTFScene: basic support for KHR_materials_pbrSpecularGlossiness (copy #625)

vidaaudrey:
Copy of https://github.com/cruise-automation/webviz/pull/625 for viewing CI results. (We are still reviewing the policy for sharing S3 secrets with 3rd party orgs)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+common/fixtures/SpecGlossVsMetalRough.glb filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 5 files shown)


jtbandes:
I guess the Git LFS file didn't work 😞   I'll let you fix that however you like - maybe you can tweak the CI configuration to check out with LFS or run lfs pull after checkout?

https://github.com/cruise-automation/webviz/blob/7407ef1687e19615a43194c003aec6608c4f7c51/.circleci/config.yml#L13

PR #630, opened 2021-05-05
Hernan/check regl instance

hhsaez:
## Summary

There have been some reports of errors during Worldview initialization:

`cannot read property 'context' of null` in `WorldviewContext.initialize`

This PR does not solve that problem, but in an attempt to better understand it, I'm adding extra checks when creating the `regl` instance and each of the rendering commands. 

## Versioning impact

Patch

### DIFF
--- a/packages/regl-worldview/package-lock.json
+++ b/packages/regl-worldview/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "regl-worldview",
-	"version": "0.16.0",
+	"version": "0.16.3",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
--- a/packages/regl-worldview/package.json
+++ b/packages/regl-worldview/package.json
@@ -1,6 +1,6 @@
 {
   "name": "regl-worldview",
-  "version": "0.16.2",
+  "version": "0.16.3",
   "description": "A reusable component for rendering 2D and 3D views using regl",
   "license": "Apache-2.0",
   "repository": "cruise-automation/webviz/tree/master/packages/regl-worldview",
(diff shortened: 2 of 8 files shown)


webpack.config.js
```diff
@@ -82,6 +82,11 @@ module.exports = {
     // https://webpack.js.org/configuration/resolve/#resolve-symlinks
     // and https://github.com/webpack/webpack/issues/1866
     symlinks: false,
+    alias: {
+      // The Buffer bundled by webpack copies data when doing Buffer.from(sharedArrayBuffer).
+      // Force use of the version in webviz-core/node_modules.
+      buffer$: path.resolve(`${__dirname}/packages/webviz-core/node_modules/buffer`),
+    },
```
troygibb, line 89: Was this change intended?

troygibb (commented):
Is the assumption here that we may be losing our regl reference in one of the lower commands? The current stack trace shows that this is an initialize issue: 

```

AppError: TypeError: Cannot read property 'context' of null
    at new <anonymous> (https://webviz.robot.car/dist/vendors~hooks_bundle.3f3672d1fe4d484fd9f6.js:2:2273756)
    at WorldviewContext.initialize (https://webviz.robot.car/dist/vendors~hooks_bundle.3f3...
```

webpack.config.js
hhsaez: Weird. That change is already in master branch. It shouldn't appear here. I rebased again and now is no longer showing as part of this PR. Sorry!

hhsaez:
> Is the assumption here that we may be losing our regl reference in one of the lower commands? The current stack trace shows that this is an initialize issue:

Yes. My assumption is that the `regl` instance is not created correctly during initialization. Since there are no checks at the moment, we proceed to compiling commands (which are the ones that end up using `regl.context`).

troygibb (approved):

PR #640, opened 2021-05-13
Add support for Draco-compressed GLTF files

hhsaez:
## Summary

GLTF supports binary data compression by using the [Draco](https://github.com/google/draco) extension. 

If a mesh contains compressed data, we need to decode each of its buffers using the tools provided by the `draco` library. Once decoded, the buffers will be stored inside the primitive (instead of using the global accessors). 

As a side effect of these changes, I had to upgrade Rollup to the latest version in order to support wasm modules. 

## Test plan

Added a story in which I'm loading a model that uses Draco compression.

## Versioning impact

This is a minor change.


packages/regl-worldview/src/utils/draco.js
```diff
@@ -0,0 +1,119 @@
+import draco3d from "draco3d";
+import draco3dWasm from "draco3d/draco_decoder.wasm";
```
troygibb, line 2: 👌  great to see this package rolled out in wasm!

troygibb (approved):
Looks great! TIL about draco, nice find :)

jtbandes:
Cool! Would you mind also reviewing #625?

Issue #610, opened 2021-03-28
Regl crash and break with custom sensor_msgs/LaserScan configuration


cjds:
Hello, 

I'm getting a traceback from regl that looks like this 

```
Error: (regl) invalid buffer data
    at Function.raise (regl.js:47)
    at initBufferFromData (regl.js:1564)
    at Object.createStream (regl.js:1506)
    at batch (eval at compile (regl.js:5549), <anonymous>:1740:9)
    at regl.js:9432
    at WorldviewContext.js:417
    at Array.forEach (<anonymous>)
    at WorldviewContext._drawInput (WorldviewContext.js:398)
    at WorldviewContext.js:246
    at Object.scope (eval at compile (regl.js:5549), <anonymous>:398:1
```

This is the bag file

https://drive.google.com/file/d/1KXH2e79CKBese_gjsAJ-XuXPSZaGYQSw/view?usp=sharing

This is the configuration

```
{
  "layout": {
    "direction": "row",
    "first": "TwoDimensionalPlot!1y50vbd",
    "second": "3D Panel!3x9jdqm",
    "splitPercentage": 30.9375
  },
  "savedProps": {
    "3D Panel!3x9jdqm": {
      "checkedKeys": [
        "name:Topics",
        "t:/atlas/localization",
        "t:/jiffy/debug/remaining_path_downsampled",
        "t:/controller_path",
        "t:/base_scan_raw"
      ],
      "expandedKeys": [
        "name:Topics"
      ],
      "followTf": "odom",
      "cameraState": {
        "targetOffset": [
          -36.91382422969203,
          -15.715546465510126,
          0
        ],
        "thetaOffset": 3.1419939577039293,
        "distance": 53.69242995869661,
        "perspective": false,
        "phi": 0.7853981633974483,
        "fovy": 0.7853981633974483,
        "near": 0.01,
        "far": 5000
      },
      "modifiedNamespaceTopics": [],
      "pinTopics": false,
      "settingsByKey": {
        "t:/base_scan_raw": {
          "overrideColor": {
            "r": 1,
            "g": 0.5058823529411764,
            "b": 0.5058823529411764,
            "a": 1
          },
          "pointSize": 5,
          "pointShape": "circle"
        }
      },
      "autoSyncCameraState": false,
      "autoTextBackgroundColor": true,
      "followOrientation": false
    }
  },
  "globalVariables": {},
  "userNodes": {},
  "linkedGlobalVariables": [],
  "playbackConfig": {
[...]

PR #592, opened 2021-03-04
Support nav_msgs/Path in Webviz

cjds:
## Summary

Add support for the message nav_msgs/Path to webviz
## Test plan

I built locally and tested with a bag with a Path in it. Here's a gif of the Path  in Webviz

## Versioning impact

Might recommend a patch version at most

PR #642, opened 2021-05-21
Don't apply override color to all nonMarkerMessages.

iwanders:
## Summary

This is a potential fix for https://github.com/cruise-automation/webviz/issues/610. The fix is based on [this comment](https://github.com/foxglove/studio/pull/374#issuecomment-809786945) in https://github.com/foxglove/studio/pull/374. The fix is not immediately apparent from the commit there, as the fix is rolled in with the changes from https://github.com/cruise-automation/webviz/pull/592. I got to this fix by comparing the current `master` branch of Webviz with the state at the end of https://github.com/foxglove/studio/pull/374.

I don't fully understand why this fixes the crash, before we were applying the `color` attribute in the `_consumeNonMarkerMessage` function, which is also used by the LaserScan message, now we apply the `color` attribute for the path message before we go into the `_consumeNonMarkerMessage` call. Maybe the `color` attribute was set twice for the LaserScan messages? It's a key though, so I would expect the last value to overwrite the first one. If anyone with more insight can chime in that'd be nice, to confirm this is an isolated fix that doesn't cause other issues.

## Test plan

I have tested this change against internal data which is known to reproduce the crash as seen in https://github.com/cruise-automation/webviz/issues/610. I have also used the bag provided by @cjds to confirm this still allows the path message to be colored. Coloring of laser scans also still works as expected.

## Versioning impact

Bug fix / patch - this bug has prevented us from updating since it started occuring on `master`.


cjds (approved):
@vidaaudrey can you please review this?

vidaaudrey:
@cjds  thanks for contributing. I'll have to pull the changes internally and see if there is any issue. Will let you know once I'm done testing. Probably later today or tomorrow.

vidaaudrey:
Hi @cjds, there main issue is with laserScan and I just sent a fix in #644. There are also some other build issues. With #643 and #644, you should be able to view the bag. I'll try to get it merged soon.

iwanders:
Thanks for taking a look at this PR & issue @vidaaudrey, I pulled 286198ec3300df3ff4791ff8f4e5b2ce3ab2436b to our internal test server and confirmed this does indeed also fix the regl buffer error.

What's the difference between the fix I proposed here and the one in #644 ? Main change in #644 seems to be the `toRGBA` to ensure the type is correct, whereas the 'fix' here changes at what place the color override is applied for the laser scan only, perhaps also affecting the type of the color override when it's finally used?

Either way, I'm happier to see your fix :+1: shall I just close this PR in favour of #644?

vidaaudrey:
My main rational for fixing the laser scan instead of the override color for non-markers is so that we could keep consistency upstream: all non-markers could use the override color in topic settings. In general ROS markers use `{r, g, b, a}` as color format anyhow. Yeah, feel free to close this PR. @hassoncs  is helping fixing the build errors in #643 now. Should be able to get the changes in soon.

iwanders:
I see, thanks for the explanation!

I'll close this PR and track #644 :+1:

PR #658, opened 2021-07-13
[2021-07-13] Update packages/webviz-core from internal repo 

hassoncs:
### Major Improvements
- 3D panel rendering is now done in a web worker.
- Added new Audio panel

### Improvements
- Very large and very small y-scale values in plots are shown in scientific notation.
- Added an `.@length` topic path suffix to the Plot Panel that gets the length of the array queried.
- Added new color utility functions to Node Playground: `rotateHue`, `rgbToHsl`, and `hslToRgb`.
- Added new lodash-like utility functions to Node Playground: `keyBy`, `groupBy`, and `mapValues`.
- Improved the measure tool's usability at small distances by adding camera-based scaling and transparency to the markers. 
- Added the ability to screenshot a panel directly to the clipboard. It can be found in the Panel settings menu under Capture as image > Copy to clipboard.
- Added a warning when Node Playground nodes have names that include more than two forward slashes.
- Added a keyboard shortcut, `l`, to open the layout menu.
- Added a "Find in Topic Tree" button to the the 3D panel's "Selected object" menu. 
- Added support to show icon groups.
- Updated the chart component

### Fixes
- Fixed an issue where the topic tree would overflow the 3D panel.
- Fixed the measuring tool so it's always drawn at a constant pixel size.
- Fixed "Set as default" layout button in the Layouts menu.
- Fixed icons persisting in the 3D panel after their topics are disabled.
- Fixed messages getting dropped when the player is paused.
- Fixed copying arrays of messages in the Raw Messages panel.
- `NaN` values do not generate diffs in the Raw Messages panel's diff mode.


davidswinegar (approved):

vidaaudrey (approved):
wow, lots of changes! 🎉  🎉  🎉  
In PR description, maybe 
- Remove the jira tickets
- Mention that we added Audio panel
- Mention we can now preload transforms and we've updated the chart component

PR #661, opened 2021-07-23
Allow absolute import paths everywhere

hassoncs:
We already enabled this in the closed-source repo, but let's add it here as well.
This allows all code to import using paths like `import X from "webviz-core/any/path"`.

This allows script files like `record-local-bag-video.js` to use absolute import paths as well. I realized we needed this when the close-source repo started erroring when running the `test_with_open_source` due to this error: `Error: Cannot find module 'webviz-core/src/util/globalConstants'`. This was because this repo has relied on some import paths being relative, but this should resolve it going forward.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "babel-eslint": "^10.0.1",
     "babel-jest": "^23.4.2",
     "babel-loader": "^8.0.4",
+    "babel-plugin-module-resolver": "^4.1.0",
     "case-sensitive-paths-webpack-plugin": "^2.1.2",
     "css-loader": "^1.0.1",
     "enzyme": "^3.9.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -27,6 +27,7 @@
     "babel-eslint": "^10.0.1",
     "babel-jest": "^23.4.2",
     "babel-loader": "^8.0.4",
+    "babel-plugin-module-resolver": "^4.1.0",
```
troygibb, line 30: #nit: prefer non-semantic versions

troygibb (approved):
👌  thx for following up on this

package.json
hassoncs: Good call @troygibb! I'm going to merge this PR to unblock my other PR, but I'm going to switch all the packages over this sprint. Tracking here: https://jira.robot.car/browse/AVA-13524

PR #667, opened 2021-08-16

davidswinegar:
* Log errors instead of crashing the page when WebGL2 isn't available.
* Allowed consuming static URDF transforms.
* Playhead hover marker now constrains to valid times when hovering invalid times on plots with a time axis.
* Made the URL properly sync to the playback time in "header stamp" mode.
* Increased max zoom percent from 150% to 1,000% in the Image panel.
* Speed unit conversion operators were added to the plot panel: @mph2kph, @mph2mps (miles per hour to metres per second) etc.
* Fixed typos in the "Seek" section of the Keyboard Shortcuts list.
* Fixed OOM error when connecting to a live ROS instance.


packages/webviz-core/script/stats/master_0d5abd39_2021-03-15T19:25:27.400Z.json
```diff
@@ -0,0 +1,199 @@
+{
```
MatthewSteel, line 1: Is this dir meant to be here?

MatthewSteel (approved):
Looks good except the benchmark data I think.

packages/webviz-core/script/stats/master_0d5abd39_2021-03-15T19:25:27.400Z.json
davidswinegar: Looks wrong

PR #673, opened 2021-09-23

hassoncs:
**Changelog**
- Added a 'clear' button to the 3D panel's polygon drawing tool menu.
- Outputting nested arrays in Node Playground now throws a type error.
- Fixed a bug in the layout migration system causing the 3D panel to crash.
- Fixed a bug in the Performance Trace panel where filtering by node in `Ticks` mode would make all traces disappear. 
- State Transitions panel will no longer slide data off the screen.
- Added global variable string interpolation to the 3D panel search using `${variable_name}` syntax.
- The 3D panel's search text is now stored in the layout.
- Holding `⌘` while seeking using the cursor keys causes a seek of 1ms.


vidaaudrey (approved):
🎉

PR #687, opened 2021-12-02
Update GLText to support padding, rounded corners, and unicode strings

hassoncs:
This updates the GLText component in exciting ways:
* Arbitrary padding for text backgrounds
* % based per-line rounded corners
* Unicode character support (great for icon rendering)

Previously we relied on the natural placement of characters during rendering to place characters in the atlas. This required every character to use the full line-height in the atlas, taking up unnecessary space (even for small characters like '.'). In order to support better packing of SDF texture atlases, this PR adds two new per-character properties, `height` and `yOffset`.
These properties will allow us to pack characters tighter in the atlas without worrying about overlapping characters at runtime. Since this changes the way worldview consumes existing atlases, we should probably cut a new major version.

- Updates to the latest version of TinySDF
- Adds an SDFGenerator component allowing users to easily create their own precomputed textAtlases

Test plan: Updated and added new screenshot tests

Rounded corners
image
image
image

**SDF Generator Component**
image


packages/regl-worldview/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "regl-worldview",
-  "version": "0.17.0",
+  "version": "0.18.0-prerelease.3",
```
hassoncs, line 3: Change this to an actual release version

jest/jest.config.js
```diff
@@ -34,4 +34,5 @@ module.exports = {
     "react-monaco-editor": "<rootDir>/packages/webviz-core/src/test/stubs/MonacoEditor.js",
     "\\.css$": "<rootDir>/jest/styleMock.js",
   },
+  transformIgnorePatterns: ["node_modules/(?!(@mapbox\\/tiny-sdf)/)"],
```
hassoncs, line 37: Seems like this is needed for Jest to run with tiny-sdf version 2 since the source isn't compiled

packages/regl-worldview/src/commands/GLText.js
```diff
@@ -272,63 +299,100 @@ const vert = `
...
+  float rectangleMask(vec2 uv, vec2 position, vec2 size) {
+    float t = 0.;
+    if ((uv.x > position.x - size.x / 2.) && (uv.x < position.x + size.x / 2.)
+      && (uv.y > position.y - size.y / 2.) && (uv.y < position.y + size.y / 2.)) {
+      t = 1.;
+    }
+    return t;
+  }
+
+  float roundedRectangleMask(vec2 position, vec2 size, float radius) {
```
hassoncs, line 361: These could probably be the same function since they do the same thing, but one has rounded corners

packages/regl-worldview/src/stories/SDFGenerator.stories.js
```diff
@@ -0,0 +1,42 @@
...
+import { storiesOf } from "@storybook/react";
+import React from "react";
+
+import SDFGenerator from "../utils/SDFGenerator";
+
+const DEFAULT_ATLAS_CONFIG = {
+  fontSize: 14,
+  fontFamily: "sans-serif",
+  charSet: new Set(
+    " !\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~".split("")
```
hhsaez, line 16: Not sure if it's worth it, but I would add a couple of icons here too.

packages/regl-worldview/src/commands/GLText.js
hhsaez: I guess `rectangleMask` can call `roundedRectangleMask` with no radius, maybe?

hhsaez (approved):
This is amazing!

Some screenshots are showing differences and text is disappearing ([here](https://open-source-webviz-ui.s3.amazonaws.com/b39aeb24e53e6cd8a7258cda1ae8b2aa8afc4c4d/index.html#Interaction/%20open-close%20behavior/does%20not%20open%20after%20selectedObject%20is%20set%20if%20disableAutoOpenClickedObject%20enabled.png)). Are those timing issues when capturing the screenshot?

packages/regl-worldview/src/stories/SDFGenerator.stories.js
hassoncs: Good call, I added a few

PR #705, opened 2022-02-07
Allow importing JSON into Sass using node-sass-json-importer

hassoncs:
This will allow us to consolidate all of our colors in one file and import that file into SASS.
This isn't used at the moment, but will unblock the closed-source PR.

Test plan: No behavior changes, build should pass


troygibb (approved):
👌

PR #715, opened 2022-02-17
Update CI to node v16

pankdm:
## Summary

The previous PR only bumped the version, but didn't change the CI config. This is now getting fixed along with the errors discovered in CI:
 - Starting npm cli v7, the user for npm run scripts changed (now this is owner of current directory: https://docs.npmjs.com/cli/v8/using-npm/scripts#user) so we need to update CircleCI script accordingly
 -  `unhandledRejection` was upgraded from Warning to Error, so we need to disable this check
 - Error messages for undefined attributes got changed
 - Couple of libraries updates

## Test plan

- Green CI. 
- Screenshots test shows  very minor differences (pixel level) so should be fine.


packages/webviz-core/package.json
```diff
@@ -19,6 +19,7 @@
     "classnames": "2.2.6",
     "compressjs": "1.0.3",
     "connected-react-router": "6.5.2",
+    "dd-trace": "^2.2.0",
```
hassoncs, line 22:
```suggestion
    "dd-trace": "2.2.0",
```

packages/webviz-core/src/util/debouncePromise.test.js
```diff
@@ -9,6 +9,11 @@
 import debouncePromise from "./debouncePromise";
 import signal from "webviz-core/shared/signal";
 
+process.on("unhandledRejection", () => {
+  // Have a dummy handler for all unhandled rejections
+  // This is to prevent UnhandledPromiseRejection on promises that we reject in tests here
+});
+
```
hassoncs, line 16: I wonder if we should put this somewhere more generic like `webviz-core/src/test/setup.js`?

hassoncs (approved):
Awesome! Thanks for writing out those helpful comments. 
Just a few minor #nit comments, but overall looks good!

packages/webviz-core/src/util/debouncePromise.test.js
pankdm: I think it's fine to leave as is - we only have 2 places where we use it and I would rather have less indirection for 2 lines of code (if 5 or more then it would make more sense to extract in my opinion)

PR #723, opened 2022-02-22
Install dirmngr when installing ffmpeg

pankdm:
Otherwise adding the keys sometimes fails with the error:
```
gpg: keyserver receive failed: No dirmngr
```

## Test plan

* Green CI should be sufficient


hassoncs (approved):
Interesting... I'm not terribly familiar with this, but if it works, it works?

pankdm:
Yeah, it's not super clear why this is only sporadically happens, but I guess installing this is fine to make sure.

pankdm:
Apparently not needed - on the latest docker image this works fine.

PR #724, opened 2022-02-23
Install rust build tools in docker

pankdm:
## Test plan

- Green CI

## Versioning impact

N/A

### DIFF
--- a/Dockerfile-webviz-ci
+++ b/Dockerfile-webviz-ci
@@ -21,6 +21,23 @@ RUN wget --no-check-certificate -q -O - https://dl-ssl.google.com/linux/linux_si
     && apt-get install -y google-chrome-stable fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst fonts-freefont-ttf --no-install-recommends
 ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
 
+# Install rust-specific libraries
+RUN apt update
+RUN apt -y install sudo
+RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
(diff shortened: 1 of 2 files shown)


Dockerfile-webviz-ci
```diff
@@ -21,6 +21,23 @@ RUN wget --no-check-certificate -q -O - https://dl-ssl.google.com/linux/linux_si
     && apt-get install -y google-chrome-stable fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst fonts-freefont-ttf --no-install-recommends
 ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
 
+# Install rust-specific libraries
+RUN apt update
+RUN apt -y install sudo
+RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
+RUN git clone https://github.com/plaurent/gnustep-build && gnustep-build/debian-10-clang-8.0/GNUstep-buildon-debian10.sh
+RUN apt install -y libatk1.0-0 libatk-bridge2.0-0 libxkbcommon0
```
hassoncs, line 29: I mentioned this in the other review, but if we merge some of these lines, we'll theoretically get a faster build (if there's a perviously cached layer available):
https://github.robot.car/cruise/webviz/pull/5796#discussion_r1039452

hassoncs (approved):

Dockerfile-webviz-ci
pankdm: Here I think this actually doesn't matter since we are not running this in CI (but rather pushing the image manually which happens fairly infrequently). I'll keep as is since to not re-trigger and wait for CI.

PR #732, opened 2022-03-10
Add vertex offset before applying geometry's pose

hhsaez:
## Summary

Vertex offset is currently applied after the pose, but offset is in local space instead of world space.
This lead to a bug where marker orientations end up being ignored in some cases.

### Before
Screen Shot 2022-03-10 at 11 09 54 AM

Notice how spheres are not in the correct final position for the object to the right of the screen. Orientation is not applied in this case (spheres have the same orientation as in the object to the left).

### After
Screen Shot 2022-03-10 at 11 05 14 AM

Now spheres are in the correct position and orientation.

## Test plan

All existing tests must pass

## Versioning impact

This is a bug fix, so it should be a patch.

### DIFF
--- a/packages/regl-worldview/package-lock.json
+++ b/packages/regl-worldview/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "regl-worldview",
-	"version": "0.18.0",
+	"version": "0.18.1",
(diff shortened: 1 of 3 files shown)


hassoncs (approved):
Nice

